PR #32, opened 2015-01-02
Update dependencies: Current GoDep doesn't capture CoreOs build.

rjnagal:
Godep also upgraded yaml to v2.


vishh:
Can you try building with the new deps @rjnagal ?

rjnagal:
godep go build ./... was failing before the update. Works fine after adding CoreOs deps.

vishh:
Ok. LGTM. Thanks for the fix @rjnagal :+1:

PR #36, opened 2015-01-02
Add vet checks for all subdirectories.

rjnagal:
Just realized that cAdvisor wasn't checking vet for all dirs. Same applies here too.


vishh:
Good catch. LGTM

PR #47, opened 2015-01-12
Fix integration test.

vishh:


integration/framework.go
```diff
@@ -39,6 +39,9 @@ type kubeFramework interface {
 
 	// Run kubectl.sh command
 	RunKubectlCmd(cmd ...string) (string, error)
+
+	// Cleanup cluster
+	Cleanup()
```
vmarmol: DestroyCluster?

vmarmol:
Can we change the commit message to provide more detail about the change? Seems like it is cluster-destruction related.

rjnagal:
btw, it requires credentials to view jenkins failure.

vishh:
I haven't figured out the security model for jenkins yet. That is why its
restricted. I will create a user for you @rjnagal and @vmarmol with
temporary credentials.

integration/framework.go
vishh: Done.

rjnagal:
Merging, @vishh is still adding more fixes to the test.

vishh:
@vmarmol: Addressed your comments.

PR #49, opened 2015-01-13
Convert json config to yaml format. Updated integration tests.

vishh:


vmarmol:
LGTM, although aren't there some scripts that automatically add the user/password? Those may need changing.

vishh:
In the kube world, passwords are updated. Once this change goes in, I will
get rid of auth for grafana even in the kube world.

vishh:
This is ready to be merged now.

PR #50, opened 2015-01-14
Add support to use custom docker images for integration testing.

vishh:


vishh:
This is the final piece of integration testing. Yayyy!

vmarmol:
WDYT of building the images as files and shipping those manually? Instead of depending on the registry and a specific account?

vishh:
Unless we have a separate registry, how can we ship docker images from the test machine over to the nodes?

vmarmol:
I've been working on something similar for cAdvisor (target is Jenkins), but I haven't quite finished. Take a look here:

https://github.com/vmarmol/cadvisor/commit/20046d6d629052d809c88b234dfbcf2d063c2265

Specifically, `docker save` and `docker load`. That code saves a specific container into a tar file, ships it over to another machine and loads it there.

vishh:
Awesome. Yeah this sounds like a good idea. How about making this an
incremental change on top of this PR?

vmarmol:
LGTM

PR #55, opened 2015-01-16
Add some debug information to heapster startup.

rjnagal:
Ideally, we want to have a validate endpoint for heapster. But we need to expose a port for that.
Not sure if its worth just for debugging info and if there's a lighter solution.


vmarmol:
LGTM

vmarmol:
I also think its fine to add an HTTP server and listen on it. It should be light enough that we won't notice.

PR #57, opened 2015-01-16
Update deps

vishh:


vishh:
Verified that e2e testing works.

PR #61, opened 2015-01-19
Updating influxdb version to 0.8.8. 

vishh:
Bumped up influx container version number.

This update paves way for adding collectd support in influxdb.


rjnagal:
LGTM

PR #64, opened 2015-01-21
Add internal state of heapster to validate endpoint.

rjnagal:
Sample output:
Heapster Version: 0.5

Source type: Kube
	Using kubelet port "10250"
	Supported kubelet versions [v0.3]
	Healthy Nodes:
		kubernetes-minion-1.c.vishnuk-cloud.internal
		kubernetes-minion-2.c.vishnuk-cloud.internal
	No node errors
	No pod errors

Sink type: Influxdb
	client: Host "10.0.113.151:8085", Database "k8s"
	Data buffering duration: 10s
	Number of write failures: 4

### DIFF
--- a/sinks/influxdb.go
+++ b/sinks/influxdb.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/heapster/sources"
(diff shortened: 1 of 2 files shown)


sources/kube.go
```diff
@@ -36,14 +37,66 @@ var kubeVersions = []string{"v0.3"}
...
+
+	self.nodeErrors[name] = self.nodeErrors[name] + 1
+}
+
+func (self *KubeSource) recordPodError(podName string, hostIp string) {
+	self.stateLock.Lock()
+	defer self.stateLock.Unlock()
+
+	podInstance := PodInstance{Pod: podName, HostIp: hostIp}
+	self.podErrors[podInstance] = self.podErrors[podInstance] + 1
```
vmarmol: nit: self.podErrors[...]++

here and above

vmarmol:
Just nit, otherwise LGTM

sinks/influxdb.go
```diff
@@ -40,6 +41,23 @@ type InfluxdbSink struct {
...
+	stateLock      sync.RWMutex
+	// TODO(rjnagal): switch to atomic if writeFailures is the only protected data.
+	writeFailures int // guarded by stateLock
+}
+
+func (self *InfluxdbSink) recordWriteFailure() {
+	self.stateLock.Lock()
+	defer self.stateLock.Unlock()
+
+	self.writeFailures += 1
```
vishh: nit: self.writeFailures++ instead?

sources/kube.go
```diff
@@ -36,14 +37,66 @@ var kubeVersions = []string{"v0.3"}
 // Cadvisor port in kubernetes.
 const cadvisorPort = 4194
 
+type PodInstance struct {
+	Pod    string
+	HostIp string
```
vishh: Pod UID might be a better to differentiate between different instances of the same Pod.

sources/kube.go
rjnagal: done.

sources/kube.go
rjnagal: The errors are probably due to problem from host. Do you want to add UID in addition to pod and host name?

vishh: Ahh I see the purpose. Lets add the UID also just to be sure.

vishh:
This will be very useful @rjnagal. Awesome. 
Ignore the integration tests until I fix them.

rjnagal:
Updated with added pod ID and other fixes.

vishh:
LGTM

PR #65, opened 2015-01-21
Fix messages to show master ip rather than a pointer address.

rjnagal:


sources/kube.go
```diff
@@ -247,7 +247,7 @@ func newKubeSource() (*KubeSource, error) {
 		Insecure: true,
 	})
 
-	glog.Infof("Using Kubernetes client %+v\n", kubeClient)
+	glog.Infof("Using Kubernetes client with master %q and version 'v1beta1'\n", *argMaster)
```
vmarmol: nit: Can we constify "v1beta1" here, below, and in the struct above so they all come from one variable?

vmarmol:
LGTM

sources/kube.go
rjnagal: done.

vmarmol:
LGTM

vishh:
I have fixed the integration tests and started one for this build. LGTM.

vmarmol:
:D

PR #66, opened 2015-01-21
Make heapster discover InfluxDB service using DNS.

vishh:
Environment variables based service discovery is still the first choice.
Updated service config to use port 80 since kube dns seems to require that.
Fixed initialization errors in kube source.

### DIFF
--- a/deploy/grafana-service.yaml
+++ b/deploy/grafana-service.yaml
@@ -1,7 +1,7 @@
 apiVersion: "v1beta1"
 kind: "Service"
 id: "monitoring-grafana"
-port: 8087
+port: 80
 containerPort: 80
 selector: 
   name: "influxGrafana"
--- a/deploy/run.sh
+++ b/deploy/run.sh
@@ -1,20 +1,24 @@
 #!/bin/bash
 
-set -e
-
-KUBE_ARGS=""
-
-if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
-    echo "Detected Kube specific args. Starting in Kube mode."
-    KUBE_ARGS="--kubernetes_master $KUBERNETES_RO_SERVICE_HOST:$KUBERNETES_RO_SERVICE_PORT"
-fi
+set -ex
 
 # Check if InfluxDB service is running
-if [ ! -z $MONITORING_INFLUXDB_SERVICE_PORT ]; then
-# TODO(vishh): add support for passing in user name and password.    
-    /usr/bin/heapster $KUBE_ARGS --sink influxdb --sink_influxdb_host "${MONITORING_INFLUXDB_SERVICE_HOST}:${MONITORING_INFLUXDB_SERVICE_PORT}"
+if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
+  # TODO(vishh): add support for passing in user name and password.    
+  INFLUXDB_ADDRESS=""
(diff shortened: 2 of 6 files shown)


vishh:
cc @rjnagal

vishh:
Integration tests are now working :)

deploy/grafana-service.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "v1beta1"
 kind: "Service"
 id: "monitoring-grafana"
-port: 8087
+port: 80
```
rjnagal: Why change all ports to 80?

sources/kube.go
```diff
@@ -255,6 +255,9 @@ func newKubeSource() (*KubeSource, error) {
 		client:      kubeClient,
 		lastQuery:   time.Now(),
 		kubeletPort: *argKubeletPort,
+		goodNodes:   make([]string, 0),
```
rjnagal: slices don't need initialization

RELEASES.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## 0.6 (1-21-2014)
+- New /validate REST endpoint to probe heapster.
+- Heapster uses InfluxDB service DNS name while running in kube mode.
```
rjnagal: Do you want to separate out release PR

deploy/grafana-service.yaml
vishh: Kube DNS service assumes port '80' to be the default port for all services. It is not identifying the port specified for the service in the Spec. This might be a bug that has already been fixed. Since every service has its own IP, the port here doesn't really matter much.

rjnagal:
@vishh can you rebase?

RELEASES.md
vishh: I will be making a separate release via github. This is just updating the relevant files.

sources/kube.go
vishh: Done.

RELEASES.md
vishh: Done.

vishh:
@rjnagal: PTAL

deploy/run.sh
```diff
@@ -1,20 +1,24 @@
...
 
+# Check if InfluxDB service is running
 if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
-    echo "Detected Kube specific args. Starting in Kube mode."
-    KUBE_ARGS="--kubernetes_master $KUBERNETES_RO_SERVICE_HOST:$KUBERNETES_RO_SERVICE_PORT"
-fi
+  # TODO(vishh): add support for passing in user name and password.    
+  INFLUXDB_ADDRESS=""
+  if [ ! -z $MONITORING_INFLUXDB_SERVICE_HOST ]; then
+    INFLUXDB_ADDRESS="${MONITORING_INFLUXDB_SERVICE_HOST}:${MONITORING_INFLUXDB_SERVICE_PORT}"
```
rjnagal: Do we need to check if a port was passed in (MONITORING_INFLUXDB_SERVICE_PORT)?

vishh: These variables are populated by the kubelet as part of static service
discovery. We can add more additional checks for sure, but it might not be
necessary.

rjnagal:
LGTM

PR #69, opened 2015-01-21
Do not update errors for pod instances that are not yet up on the node.

rjnagal:


vmarmol:
LGTM

sources/kube.go
```diff
@@ -259,6 +272,7 @@ func newKubeSource() (*KubeSource, error) {
 		client:      kubeClient,
 		lastQuery:   time.Now(),
 		kubeletPort: *argKubeletPort,
+		activePods:  make(map[PodInstance]bool),
```
vishh: This might result in linear memory growth since we don't prune the activePods map at all.

rjnagal: yes, this is a problem because we don't keep a list of pods being tracked. Would it be good enough to regenerate the list in getPods() - purge all old entries that are not present in the new listing from getPods()?

vishh: Yeah. we can do that.

rjnagal:
Updated the change to purge old pods from activePods and podError maps.

vmarmol:
LGTM

sources/kube.go
```diff
@@ -66,11 +68,39 @@ func (self *KubeSource) recordNodeError(name string) {
...
+	self.podErrors = newPodErrors
+}
 
+func (self *KubeSource) recordPodError(podName, podId, hostIp string) {
+	self.stateLock.Lock()
+	defer self.stateLock.Unlock()
 	podInstance := PodInstance{Pod: podName, PodId: podId, HostIp: hostIp}
+
+	// Heapster knows about pods before they are up and running on a node.
+	// Ignore errors till we have seen an instance up at least once.
```
vishh: It just occurred to me that another approach to handle this is to ignore Pods that are not in "Running" state.

PR #71, opened 2015-01-22
Support kube namespaces 

### LINKED ISSUE #31: Namespaced containers stats aren't reported
Not sure if it's a problem with cAdvisor, Heapster or Kubernetes but using [kubernetes setup for monitoring](https://github.com/GoogleCloudPlatform/kubernetes/tree/master/examples/monitoring) I don't see in grafana pods started in namespaces other than the default one.

vishh:
Fixes #31.

This PR breaks backwards compatibility since the new kubelet API is available only from versions 0.9.0. Users who run older clusters can use an older version of heapster though.

### DIFF
--- a/integration/kube_test.go
+++ b/integration/kube_test.go
@@ -40,7 +40,7 @@ var (
 	heapsterFirewallRule        = "heapster-e2e"
 	influxdbLabels              = map[string]string{"name": "influxGrafana"}
 	heapsterLabels              = map[string]string{"name": "heapster"}
-	kubeVersions                = flag.String("kube_versions", "0.7.2,0.8.1", "Comma separated list of kube versions to test against")
+	kubeVersions                = flag.String("kube_versions", "", "Comma separated list of kube versions to test against")
 	heapsterManifestFile        = flag.String("heapster_controller", "../deploy/heapster-controller.yaml", "Path to heapster replication controller file.")
 	influxdbGrafanaManifestFile = flag.String("influxdb_grafana_controller", "../deploy/influxdb-grafana-controller.yaml", "Path to Influxdb-Grafana replication controller file.")
 	influxdbServiceFile         = flag.String("influxdb_service", "../deploy/influxdb-service.yaml", "Path to Inlufxdb service file.")
--- a/sources/kube.go
+++ b/sources/kube.go
@@ -161,6 +161,7 @@ func (self *KubeSource) listMinions() (*nodeList, error) {
 func (self *KubeSource) parsePod(pod *kube_api.Pod) *Pod {
 	localPod := Pod{
 		Name:       pod.Name,
+		Namespace:  pod.Namespace,
 		ID:         string(pod.UID),
 		Hostname:   pod.Status.Host,
 		Status:     string(pod.Status.Phase),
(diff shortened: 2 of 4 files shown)


integration/kube_test.go
```diff
@@ -40,7 +40,7 @@ var (
 	heapsterFirewallRule        = "heapster-e2e"
 	influxdbLabels              = map[string]string{"name": "influxGrafana"}
 	heapsterLabels              = map[string]string{"name": "heapster"}
-	kubeVersions                = flag.String("kube_versions", "0.7.2,0.8.1", "Comma separated list of kube versions to test against")
+	kubeVersions                = flag.String("kube_versions", "", "Comma separated list of kube versions to test against")
```
vmarmol: Should we get at least 1? Or waiting for 0.9.0?

vmarmol:
LGTM, although needs rebase.

integration/kube_test.go
vishh: I removed the flag defaults here and added the flag to integration/.jenkins.sh

vishh:
@rjnagal: I simplified the pod errors code a bit. 
@vmarmol: Rebased

integration/kube_test.go
vmarmol: Ah I see

vmarmol:
Can we split that into a separate PR? Its's not even mentioned in the commits.

sources/kube.go
```diff
@@ -68,39 +66,17 @@ func (self *KubeSource) recordNodeError(name string) {
...
-			newActivePods[podInstance] = true
-			if self.podErrors[podInstance] != 0 {
-				newPodErrors[podInstance] = self.podErrors[podInstance]
-			}
-		}
+func (self *KubeSource) recordPodError(pod Pod) {
+	// Heapster knows about pods before they are up and running on a node.
+	// Ignore errors for Pods that are not Running.
+	if pod.Status != "Running" {
+		return
```
rjnagal: kubelet will set pod to running when at least one container instance is running. Will that return it back to the earlier state?

vishh: Kubelet is not expected to set the Pod state to running unless all the containers are in running state. If you notice a different behavior its probably a bug.

vishh:
@vmarmol: I have rearranged the commits a bit to provide better context. PTAL

sources/kube.go
rjnagal: SG. I think the comment in pod_cache differentiates between started and running, but if kubelet does it, we can trust that.

vishh: YUP. AFAIK kubelet waits until all containers are running. On the other hand what is the intended purpose of this counter? Is it to capture infrastructure (kubelet/cadvisor) errors while trying to retrieve stats information for a "running" pod?

vishh:
Ping.

sources/kube.go
rjnagal: The purpose is to record any errors retrieving pod data once the pod is up and running. I think this should do. I need to look at pod restart behavior, but it doesn't need to be done here.

vmarmol:
LGTM

rjnagal:
LGTM

PR #75, opened 2015-01-28
Export fs stats at the node level.

vishh:


vishh:
Ping @rjnagal

rjnagal:
LGTM.

Can we auto-generate the js files? 80% of the file is boilerplate.

PR #80, opened 2015-02-04
Refactored node discovery part into a separate package

vishh:
This code is not yet integrated into heapster core


rjnagal:
LGTM

PR #81, opened 2015-02-05
Allow https kube master

jimmidyson:
Currently this is hard coded as http. This PR lets user specify URL including protocol to use, defaulting to http so as not to affect existing users.


rjnagal:
Thanks for the fix.

LGTM

vishh:
LGTM

PR #85, opened 2015-02-07
Add creation and pushing of the first few metrics.

vmarmol:


vmarmol:
Here are the first couple of metrics. More will come, but I wanted to get the flow right first.

sinks/gcm/sink.go
```diff
@@ -31,24 +37,233 @@ func NewSink() (*GcmSink, error) {
...
+	var buffer bytes.Buffer
+	first := true
+	for key, value := range labels {
+		if !first {
+			buffer.WriteString(",")
+		}
+		buffer.WriteString(key)
+		buffer.WriteString("=")
+		buffer.WriteString(value)
+		first = false
```
rjnagal: nit: move to !first check.

sinks/gcm/sink.go
```diff
@@ -31,24 +37,233 @@ func NewSink() (*GcmSink, error) {
...
+func (self *GcmSink) containerToMetrics(container *sources.Container, labels map[string]string) []Metric {
+	labels[lblContainerName] = container.Name
+
+	// One metric value per data point.
+	metrics := make([]Metric, 0, len(container.Stats))
+	for _, stat := range container.Stats {
+		// Add all supported metrics that have values.
+		for _, supported := range self.supportedMetrics {
+			if supported.HasValue(&container.Spec) {
+				// Cumulative stats have as a start time the container creation time.
```
rjnagal: nit: reword - 'Cumulative stats have container creation time as start time'

sinks/gcm/sink.go
```diff
@@ -31,24 +37,233 @@ func NewSink() (*GcmSink, error) {
...
+	for _, label := range allLabels {
+		desc += fmt.Sprintf("\t\t%s: %s", label.Key, label.Description)
+	}
+
 	desc += "\n"
 	return desc
 }
+
+const (
+	lblPodId         = "pod_id"
```
rjnagal: nit: maybe just spell lbl out, eg labelPodId?

rjnagal:
LGTM

sinks/gcm/sink.go
vmarmol: Do you mean to put it inside the if? It won't trigger the first time. Moved it closer to the if in case that is what you meant :)

sinks/gcm/sink.go
vmarmol: Done

sinks/gcm/sink.go
vmarmol: Done

sinks/gcm/sink.go
```diff
@@ -31,24 +37,233 @@ func NewSink() (*GcmSink, error) {
...
+		Key:         labelContainerName,
+		Description: "User-provided name of the container or full container name for system containers",
+	},
+	LabelDescriptor{
+		Key:         labelLabels,
+		Description: "Comma-separated list of user-provided labels",
+	},
+	LabelDescriptor{
+		Key:         labelHostname,
+		Description: "Hostname where the container ran",
```
vishh: To be unambiguous this would be the "hostname of the node where the container ran"

sinks/gcm/sink.go
```diff
@@ -31,24 +37,233 @@ func NewSink() (*GcmSink, error) {
...
+
+	// Returns whether this metric is present.
+	HasValue func(*cadvisor.ContainerSpec) bool
+
+	// Returns the desired data point for this metric from the stats.
+	GetValue func(*cadvisor.ContainerStats) interface{}
+}
+
+// TODO(vmarmol): Add the rest of the metrics.
+var allMetrics = []supportedMetric{
```
vishh: How about moving this to a separate file since it will grow quite a bit?

vishh:
LGTM. Merging this since this should not affect existing integration tests.

Issue #95, opened 2015-02-10
export diskio stats


vishh:
cAdvisor supports diskio stats. It needs to be exported to the storage backends.

piosz:
closing in favor of #690

PR #96, opened 2015-02-10
Remove diskio stats from being pushed to InfluxDB. 

vishh:
The stats are handled incorrectly.
#95


sinks/gcm/metrics.go
```diff
@@ -30,7 +30,7 @@ type supportedMetric struct {
 
 // TODO(vmarmol): Add the rest of the metrics.
 var allMetrics = []supportedMetric{
-	supportedMetric{
+	{
```
vmarmol: nit: these unrelated changes should be in another commit

vmarmol:
LGTM, merging now since it's only a nit.

PR #105, opened 2015-02-16
Fix Pod with Namespace, fix Minion with IP as Name

cedbossneo:
- Fix Pod with Namespace
- Fix Minion with IP as Name instead of Hostname

### DIFF
--- a/sources/nodes/kube.go
+++ b/sources/nodes/kube.go
@@ -78,8 +78,15 @@ func (self *kubeNodes) List() (*NodeList, error) {
 			nodeList.Items[Host(node.Name)] = Info{PublicIP: node.Status.HostIP, InternalIP: addrs[0].String()}
 			goodNodes = append(goodNodes, node.Name)
 		} else {
-			glog.Errorf("Skipping host %s since looking up its IP failed - %s", node.Name, err)
-			self.recordNodeError(node.Name)
+			ip := net.ParseIP(node.Name)
+			if ip != nil {
+				//If the node name can't be resolved, and we can parse nodeName as IP, we use IP directly, this solve CoreOS problems
+				nodeList.Items[Host(node.Name)] = Info{PublicIP: node.Status.HostIP, InternalIP: node.Name}
+				goodNodes = append(goodNodes, node.Name)
+			} else {
+				glog.Errorf("Skipping host %s since looking up its IP failed and %s is not an IP address - %s", node.Name, err)
+				self.recordNodeError(node.Name)
+			}
 		}
 	}
 	self.recordGoodNodes(goodNodes)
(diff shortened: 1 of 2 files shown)


vmarmol:
OK to test

vmarmol:
ok to test

sources/nodes/kube.go
```diff
@@ -78,8 +78,15 @@ func (self *kubeNodes) List() (*NodeList, error) {
 			nodeList.Items[Host(node.Name)] = Info{PublicIP: node.Status.HostIP, InternalIP: addrs[0].String()}
 			goodNodes = append(goodNodes, node.Name)
 		} else {
-			glog.Errorf("Skipping host %s since looking up its IP failed - %s", node.Name, err)
-			self.recordNodeError(node.Name)
+			ip := net.ParseIP(node.Name)
```
vmarmol: Looks like the only difference is getting the InternalIP, can we move that code above and not duplicate the append and adding to nodeList?

Sorry to partially be a pain :)

vishh: Is this a custom cluster setup where the node name is its IP? Why so? Why not give the node an unique hostname that can be resolves to an IP?

cedbossneo: We can, some systems use IP instead to avoid hostname resolution issues.

Le Mon Feb 16 2015 at 19:03:03, Vish Kannan <notifications@github.com> a
écrit :

vishh: Just so that I understand clearly, the hostname of the machines in the cluster matches the IP? Does the rest of kubernetes work given the resolution issues?
Sorry for being a pain, I am just trying to understand the underlying problem!

cedbossneo: Yes, the node names(not hostname) match the IP of the node because on RackSpace you can't resolve the name of other machines, if you want to do this, you must declare each machines in your hosts file or add a local dns mechanism.

cedbossneo: I think we can split the pull request because the first changeset (	resource := filepath.Join("stats", pod.Namespace, pod.Name, pod.ID, containerName)) is required if you have a kubernetes cluster with multiple namespaces. Maybe i can find a better solution for the hostname issue.

vishh: Ahh. Thanks for explaining. +1 to splitting the PRs and thanks for identifying and fixing the namespace handling bug :+1: 
Let us know if you are not able to find a way to get around the hostname resolution issue and I can explore to see if the proxy server on the kube-master can proxy traffic to kubelets instead of having to talk to the kubelets directly.

What does `no.Status.HostIP` point to in the case of Rackspace? Does it point to an internally accessible IP with the kubelet port exposed?

PR #112, opened 2015-02-18
Adding GCM as an available Heapster sink.

vmarmol:


sinks/types.go
```diff
@@ -19,7 +19,7 @@ import (
 	"fmt"
 )
 
-var argSink = flag.String("sink", "memory", "Backend storage. Options are [memory | influxdb | bigquery]")
+var argSink = flag.String("sink", "memory", "Backend storage. Options are [memory | influxdb | bigquery | gcm ]")
```
vishh: Do we need any other auth information for gcm? What will the user have to do to get GCM to work?

vishh:
LGTM.

sinks/types.go
vmarmol: Today we only support running in GCE (which we check and error out on in New) and we get the auth information from the machine's service accounts (which we also check and error out on).

vishh: Ok. We can clarify that in the Readme as well.

PR #116, opened 2015-02-19
Native support for CoreOS.

### LINKED ISSUE #99: Suggestion: get rid of the CoreOS / "buddy"
While flexible, the buddy / `hosts` file things complicates the deployment.
I ended up moving Fleet logic into Heapster and getting rid of noticeable chunk of the code. As there are no currently other types of the buddies (Mesos?), I suggest doing something like the following:
https://github.com/arkadijs/heapster/commit/7c4fc6ca628a2e9548fd26241993f5d30dba301e

1. If there is Kubernetes - use listMinions
2. Otherwise try Fleet.

The code is little bit messy - coupling KubeSource to ExternalSource, but my baseline is CoreOS with Deis, with optional Kubernetes, thus such practical solution.

vishh:
Fixes #99

### DIFF
--- a/sources/nodes/coreos.go
+++ b/sources/nodes/coreos.go
@@ -1,0 +1,106 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nodes
+
+import (
+	"flag"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	fleetClient "github.com/coreos/fleet/client"
+	"github.com/coreos/fleet/etcd"
+	fleetPkg "github.com/coreos/fleet/pkg"
+	"github.com/coreos/fleet/registry"
+	"github.com/golang/glog"
+)
+
+var argFleetEndpoints = flag.String("fleet_endpoints", "http://127.0.0.1:4001", "Comma separated list of fleet server endpoints")
+
+const etcdRegistry = "/_coreos.com/fleet/"
+
+type fleetNodes struct {
+	client         fleetClient.API
+	nodes          *NodeList
+	apiErrors      int
+	recentApiError error
+}
+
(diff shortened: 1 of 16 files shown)


vmarmol:
```
$ godep go test -a -test.short ./...
# github.com/GoogleCloudPlatform/heapster/sources/nodes
sources/nodes/external_test.go:53: too few values in struct initializer
?   	github.com/GoogleCloudPlatform/heapster	[no test files]
?   	github.com/GoogleCloudPlatform/heapster/clusters/coreos	[no test files]
ok  	github.com/GoogleCloudPlatform/heapster/integration	0.016s
?   	github.com/GoogleCloudPlatform/heapster/sinks	[no test files]
?   	github.com/GoogleCloudPlatform/heapster/sinks/gcm	[no test files]
ok  	github.com/GoogleCloudPlatform/heapster/sources	0.018s
?   	github.com/GoogleCloudPlatform/heapster/sources/api	[no test files]
ok  	github.com/GoogleCloudPlatform/heapster/sources/datasource	0.019s
FAIL	github.com/GoogleCloudPlatform/heapster/sources/nodes [build failed]
?   	github.com/GoogleCloudPlatform/heapster/validate	[no test files]
?   	github.com/GoogleCloudPlatform/heapster/version	[no test files]
godep: go exit status 2
The command "godep go test -a -test.short ./..." exited with 1.
```

vishh:
@vmarmol @rjnagal: I want to include this in the v0.7 release. I tested manually and it works. I have updated the readme and release notes.

sources/nodes/coreos.go
```diff
@@ -0,0 +1,106 @@
...
+	if self.nodes != nil {
+		output = fmt.Sprintf("%s\nCadvisor Nodes: %v", output, self.nodes.Items)
+	}
+	return output
+}
+
+func getFleetRegistryClient(fleetEndpoints []string) (fleetClient.API, error) {
+	var dial func(string, string) (net.Conn, error)
+
+	tlsConfig, err := fleetPkg.ReadTLSConfigFiles("", "", "")
```
vmarmol: This won't let us do HTTPS no? What is the alternative to this? Do you actually have to give it the file paths?

vmarmol:
LGTM, just a question we can continue in another PR

sources/nodes/coreos.go
vishh: That is true. If users set up https, then we will have to pass auth info. I don't want to do it unless some one needs it.

vmarmol: I *think* we can just get this from the filesystem.

vishh: Depends on if we have access to the 'filesystem'.

vmarmol: We do from the container :)

I have to send a PR to add a hostVolume for Heapster so we can use HTTPS. We need it for GCM and I forgot to add it :P

vishh: Great!

PR #121, opened 2015-02-20
Add host volume with TLS certs.

### LINKED ISSUE #120: Add files needed for TLS to Heapster container
Add the host volumes needed to send TLS requests in Heapster. This is needed by some sinks like GCM.

vmarmol:
Fixes #120.

### DIFF
--- a/deploy/heapster-controller.yaml
+++ b/deploy/heapster-controller.yaml
@@ -18,6 +18,15 @@ desiredState:
               - 
                 name: "INFLUXDB_HOST"
                 value: "monitoring-influxdb"
+            volumeMounts:
+              - name: ssl-certs
+                mountPath: /etc/ssl/certs
+                readOnly: true
+        volumes:
+          - name: ssl-certs
+            source:
+              hostDir:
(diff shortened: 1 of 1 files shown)


deploy/heapster-controller.yaml
```diff
@@ -18,6 +18,15 @@ desiredState:
               - 
                 name: "INFLUXDB_HOST"
                 value: "monitoring-influxdb"
+            volumeMounts:
+              - name: ssl-certs
+                mountPath: /etc/ssl/certs
```
vishh: Noob question: Is this the default location for storing certs for most applications?

vishh:
LGTM.

deploy/heapster-controller.yaml
vmarmol: From documentation I've found and references by @kelseyhightower, that is indeed the case.

vishh: Ok.

vishh:
Thanks for the PR @vmarmol. Merging.

PR #122, opened 2015-02-20
Let Heapster monitor localhost with an empty hosts file.

vmarmol:


rjnagal:
Would it be a problem if we accidentally forgot to add the flag and heapster just continued assuming localhost?

vmarmol:
The default of the flag is a specific file, so you'd have to specifically set it to the empty string.

rjnagal:
I just realized that after sending the comment.

LGTM

vmarmol:
:)

vishh:
LGTM

PR #124, opened 2015-02-23
Remove HostIP check while filtering pods. 

vishh:
This is required to handle situations when HostIP is not populated by the master on Pod objects.


vmarmol:
LGTM

PR #128, opened 2015-02-23
Add uptime metric to GCM sink.

vmarmol:
This metric is computed versus the container start time.


vishh:
Can you describe how you think this metric can be used? When a container dies, heapster will stop exporting all metrics including 'uptime'.

vmarmol:
The absence of this metric is useful to track, but we can also track container restarts with this. If one of the containers in your pods crashes and gets brought back up, the uptime will go down, although the pod will not change name and the rest of the containers will stay up.

vishh:
LGTM. This will be useful until we have events collection and processing implemented.

PR #131, opened 2015-02-25
Fix grafana container port config and update the readme.

vishh:


vmarmol:
LGTM. Looks like the CI is unhappy with the gcloud invocations.

vishh:
@vmarmol: Tests passed.

PR #139, opened 2015-02-26
Fix cadvisor unit test flake.

vishh:


vmarmol:
LGTM

rjnagal:
LGTM. Thanks for the quick fix :)

vishh:
This is a test only change. I think we can ignore the integration tests.

PR #141, opened 2015-02-27
Fix bugs in heapster core.

vishh:


sources/kube.go
```diff
@@ -147,11 +147,11 @@ func (self *kubeSource) getPodInfo(nodeList *nodes.NodeList) ([]api.Pod, error)
 	var (
 		wg sync.WaitGroup
 	)
-	for _, pod := range pods {
+	for index := range pods {
```
vmarmol: index is unused

vmarmol: Ah I see now, this GitHub diff is terrible :P

sources/kube.go
```diff
@@ -147,11 +147,11 @@ func (self *kubeSource) getPodInfo(nodeList *nodes.NodeList) ([]api.Pod, error)
...
 		wg.Add(1)
 		go func(pod *api.Pod) {
 			defer wg.Done()
-			for _, container := range pod.Containers {
+			for index, container := range pod.Containers {
```
vmarmol: Lets not use container and only index?

sources/kube.go
```diff
@@ -147,11 +147,11 @@ func (self *kubeSource) getPodInfo(nodeList *nodes.NodeList) ([]api.Pod, error)
...
 				rawContainer, err := self.getStatsFromKubelet(pod, container.Name)
 				if err != nil {
 					// Containers could be in the process of being setup or restarting while the pod is alive.
```
vmarmol: Is the log line below right? It will display "<container name>/<namespace> <pod>"

sources/pods.go
```diff
@@ -81,9 +81,9 @@ func (self *realPodsApi) parsePod(podNodePair *podNodePair) *api.Pod {
 
 func (self *realPodsApi) parseAllPods(podNodePairs []podNodePair) []api.Pod {
 	out := make([]api.Pod, 0)
-	for _, podNodePair := range podNodePairs {
+	for i, podNodePair := range podNodePairs {
```
vmarmol: Should we use both for this file too? It seems confusing and error prone.

vmarmol:
Can we add more description as to the bugs in the commit message? :)

sources/kube.go
vishh: Having container results in lesser code inside the loop. :)

sources/kube.go
vishh: Which log line? The two ones below seem to be fine.

sources/pods.go
vishh: Agreed! Done.

sources/kube.go
vishh: yeah. It is indeed confusing.

vmarmol:
LGTM

PR #149, opened 2015-03-02
Use GCM sink if SINK=gcm is specified.

### LINKED ISSUE #117: Add deployment support for the new GCM driver.
`deploy/run.sh` needs to be updated to support the new gcm backend driver.

vmarmol:
Default is influxdb which mirrors today's behavior.

Fixes #117.

### DIFF
--- a/deploy/run.sh
+++ b/deploy/run.sh
@@ -12,22 +12,42 @@ if [ ! -x $CADVISOR_PORT ]; then
   EXTRA_ARGS="$EXTRA_ARGS --cadvisor_port=$CADVISOR_PORT"
 fi
 
-HEAPSTER="/usr/bin/heapster $EXTRA_ARGS "
-# Check if InfluxDB service is running
+# If in Kubernetes, target the master.
 if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
-  # TODO(vishh): add support for passing in user name and password.    
-  INFLUXDB_ADDRESS=""
-  if [ ! -z $MONITORING_INFLUXDB_SERVICE_HOST ]; then
-    INFLUXDB_ADDRESS="${MONITORING_INFLUXDB_SERVICE_HOST}:${MONITORING_INFLUXDB_SERVICE_PORT}"
+  EXTRA_ARGS="$EXTRA_ARGS --kubernetes_master ${KUBERNETES_RO_SERVICE_HOST}:${KUBERNETES_RO_SERVICE_PORT}"
+fi
+
+# Select what source sink to use. Options: "influxdb,gcm"
+# Default is InfluxDB.
+if [ -x $SINK ]; then
+  SINK="influxdb"
+fi
+
+HEAPSTER="/usr/bin/heapster $EXTRA_ARGS "
+
+if [ "$SINK" == "influxdb" ]; then
+  HEAPSTER="$HEAPSTER --sink influxdb"
+
+  # Check if in Kubernetes.
+  if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
+    # TODO(vishh): add support for passing in user name and password.
+    INFLUXDB_ADDRESS=""
+    if [ ! -z $MONITORING_INFLUXDB_SERVICE_HOST ]; then
+      INFLUXDB_ADDRESS="${MONITORING_INFLUXDB_SERVICE_HOST}:${MONITORING_INFLUXDB_SERVICE_PORT}"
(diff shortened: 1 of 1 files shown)


vmarmol:
Tested on a new cluster and it comes up as expected.

deploy/run.sh
```diff
@@ -12,22 +12,37 @@ if [ ! -x $CADVISOR_PORT ]; then
...
 HEAPSTER="/usr/bin/heapster $EXTRA_ARGS "
-# Check if InfluxDB service is running
-if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
-  # TODO(vishh): add support for passing in user name and password.    
-  INFLUXDB_ADDRESS=""
-  if [ ! -z $MONITORING_INFLUXDB_SERVICE_HOST ]; then
-    INFLUXDB_ADDRESS="${MONITORING_INFLUXDB_SERVICE_HOST}:${MONITORING_INFLUXDB_SERVICE_PORT}"
+
+if [ "$SINK" == "influxdb" ]; then
+  # Check if InfluxDB service is running
```
vishh: nit: The comment here should be "check if we are running in kubernetes". i see that it was incorrect to begin with.

deploy/run.sh
```diff
@@ -12,22 +12,37 @@ if [ ! -x $CADVISOR_PORT ]; then
...
-  else 
-    echo "InfluxDB service address not found. Exiting."
-    exit 1
+    $HEAPSTER --sink influxdb --sink_influxdb_host ${INFLUXDB_HOST}
+  else
+    $HEAPSTER
+  fi
+elif [ "$SINK" == "gcm" ]; then
+  if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
+    $HEAPSTER --kubernetes_master "${KUBERNETES_RO_SERVICE_HOST}:${KUBERNETES_RO_SERVICE_PORT}" --sink gcm
```
vishh: Should we do the master service discovery ahead of time and prepare heapster args before checking for storage drivers?

vishh:
I think its time to rewrite `run.sh` in golang. May be after this PR.

deploy/run.sh
vmarmol: Done, looks cleaner.

deploy/run.sh
vmarmol: Done.

vmarmol:
+1, this logic can probably live in the Heapster binary itself. We don't *need* a wrapper per say.

deploy/run.sh
```diff
@@ -12,22 +12,42 @@ if [ ! -x $CADVISOR_PORT ]; then
...
+  SINK="influxdb"
+fi
+
+HEAPSTER="/usr/bin/heapster $EXTRA_ARGS "
+
+if [ "$SINK" == "influxdb" ]; then
+  HEAPSTER="$HEAPSTER --sink influxdb"
+
+  # Check if in Kubernetes.
+  if [ ! -z $KUBERNETES_RO_SERVICE_HOST ]; then
```
vishh: I think we can get rid of this check.

deploy/run.sh
```diff
@@ -12,22 +12,42 @@ if [ ! -x $CADVISOR_PORT ]; then
...
+    # TODO(vishh): add support for passing in user name and password.
+    INFLUXDB_ADDRESS=""
+    if [ ! -z $MONITORING_INFLUXDB_SERVICE_HOST ]; then
+      INFLUXDB_ADDRESS="${MONITORING_INFLUXDB_SERVICE_HOST}:${MONITORING_INFLUXDB_SERVICE_PORT}"
+    elif [ ! -z $INFLUXDB_HOST ]; then
+      INFLUXDB_ADDRESS=${INFLUXDB_HOST}
+    else
+      echo "InfluxDB service address not found. Exiting."
```
vishh: nit: InfluxDB service address not specified.

deploy/run.sh
vishh: Ahh ignore my comment. I see below why its necessary.

vishh:
LGTM except for the one nit. If we are moving to a dynamic configuration model, we can get rid of this logic completely.

vishh:
@vmarmol: Feel free to self merge after fixing the nit.

deploy/run.sh
vmarmol: Done

vmarmol:
Thanks @vishh! Will merge on green.

PR #150, opened 2015-03-02
Adding network metrics to GCM sink.

vmarmol:
Added TODO for the remaining two types: diskIO and FS. DiskIO we may
not need yet and FS requires more infrastructure.

### DIFF
--- a/sinks/gcm/metrics.go
+++ b/sinks/gcm/metrics.go
@@ -34,7 +34,6 @@ func GetSupportedMetrics() []SupportedMetric {
 	return allMetrics
 }
 
-// TODO(vmarmol): Add the rest of the metrics.
 var allMetrics = []SupportedMetric{
 	{
 		MetricDescriptor: MetricDescriptor{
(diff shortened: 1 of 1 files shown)


vmarmol:
Looks like the minions have a full disk :P free'd up some space and kicked off the build again. Still investigating...

sinks/gcm/metrics.go
```diff
@@ -92,4 +91,76 @@ var allMetrics = []SupportedMetric{
...
 	},
+	{
+		MetricDescriptor: MetricDescriptor{
+			Name:        "memory/page_faults",
+			Description: "Number of major page faults",
+			Type:        MetricGauge,
+			ValueType:   ValueInt64,
+		},
+		HasValue: func(spec *cadvisor.ContainerSpec) bool {
+			return spec.HasMemory
```
vishh: Not related to this PR: Can we consider stashing the Has* info into cadvisor Stats instead of requiring the Spec in the latest API?

vishh:
LGTM! Let me know if you need help fixing the e2e test machine.

sinks/gcm/metrics.go
vmarmol: Yes, I think that makes sense. Similar to has_<field> in proto. Will file an issue :)

PR #151, opened 2015-03-02
Support standalone mode throught the --standalone flag.

vmarmol:
Replaces the old method of specifying an empty external hosts file to an
explicit --standalone flag.

### DIFF
--- a/sources/nodes/external.go
+++ b/sources/nodes/external.go
@@ -27,16 +27,22 @@ import (
 
 // Provides a list of external cadvisor nodes to monitor.
 type externalCadvisorNodes struct {
+	// File containing list of hosts to monitor.
 	hostsFile string
-	nodes     *NodeList
+
+	// Whether to run in "standalone" mode and only consider localhost.
+	standalone bool
+
+	nodes *NodeList
 }
 
 // While updating this, also update heapster/deploy/Dockerfile.
 var hostsFile = flag.String("external_hosts_file", "/var/run/heapster/hosts", "A file that heapster refers to get a list of nodes to monitor.")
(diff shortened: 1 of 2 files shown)


sources/nodes/external.go
```diff
@@ -27,16 +27,22 @@ import (
...
+
+	// Whether to run in "standalone" mode and only consider localhost.
+	standalone bool
+
+	nodes *NodeList
 }
 
 // While updating this, also update heapster/deploy/Dockerfile.
 var hostsFile = flag.String("external_hosts_file", "/var/run/heapster/hosts", "A file that heapster refers to get a list of nodes to monitor.")
+var standaloneMode = flag.Bool("standalone", false, "Whether to run Heapster in \"standalone\" mode where it only targets the current node.")
```
vishh: This could apply to local cadvisor or kubelet. How do we differentiate that?

vishh:
Do we have any docs for running in standalone mode?

sources/nodes/external.go
vmarmol: Today we can't target the local Kubelet since we get the pods from the apiserver no? When we change that we can move the flag to heapster.go and take it as a param here I think.

vmarmol:
I wanted to get this in and then I'll write the doc :)

vmarmol:
Failing due to disk space, cleaning up and kicking again.

vmarmol:
For some reason the web trigger is not firing, but this passes on Jenkins:

http://104.154.52.74/job/heapster-e2e-gce/547/

vishh:
LGTM

sources/nodes/external.go
vishh: Ack.

PR #152, opened 2015-03-02
Remove temporary files created by the integration test 

vishh:
Integration test created temporary files on cluster nodes. Remove them.
Cleanup unused containers and images from kubernetes nodes.
Added a doc for integration testing against local cluster.

### DIFF
--- a/docs/integration.md
+++ b/docs/integration.md
@@ -1,0 +1,24 @@
+## Heapster integration testing
+
+Heapster can be tested against an existing Kubernetes cluster running on GCE. Tests under [integration](../integration) can be used for integration testing.
+
+Figure out the server version of your current Kubernetes cluster.
+```shell
+$ kubectl.sh version
+```
+Use the full version specified there for integration testing. For example the full version can be `0.11.0` or `0.9.1`.
+
+Run the integration test as follows:
+```shell
+$ cd integration
+$ MY_VERSION=0.11.0
+$ godep go test -a -v --vmodule=*=1 --timeout=30m --kube_versions=$MY_VERSION
+```
+
+Note that the test expects `~/.kubernetes_auth` file to exist. If it does not exist, which will be the case with recent kubernetes versions,
+you can pass in the path to the auth file via the flag `--auth_config` or create a symlink from the new auth file to `~/.kubernetes_auth`.
+```shell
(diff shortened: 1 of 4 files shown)


vishh:
@vmarmol: This PR should get rid of the disk full issue.

integration/utils.go
```diff
@@ -68,3 +75,16 @@ func removeDockerImage(imageName string) error {
 	}
 	return nil
 }
+
+func cleanupRemoteHost(hostname string) error {
+	out, err := exec.Command("gcutil", "ssh", hostname, "sudo", "docker", "rm", "`docker ps -a -q`").CombinedOutput()
```
vmarmol: this will try to remove the existing kubernetes containers and fail (which is fine). Maybe just ignore the error?

integration/utils.go
```diff
@@ -68,3 +75,16 @@ func removeDockerImage(imageName string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to remove stopped containers on host %q (%q) - %q", hostname, err, out)
+	}
+	out, err = exec.Command("gcutil", "ssh", hostname, "sudo", "docker", "rmi", "`docker images -a -q`").CombinedOutput()
```
vmarmol: Same thing here with images

docs/integration.md
```diff
@@ -0,0 +1,24 @@
...
+Use the full version specified there for integration testing. For example the full version can be `0.11.0` or `0.9.1`.
+
+Run the integration test as follows:
+```shell
+$ cd integration
+$ MY_VERSION=0.11.0
+$ godep go test -a -v --vmodule=*=1 --timeout=30m --kube_versions=$MY_VERSION ./...
+```
+
+Note that the test expects `~/.kubernetes_auth` file to exist. If it does not exist, which will be the case with recent kubernetes versions,
```
vmarmol: Maybe just mention the kubernetes auth? Now-a-days they're in `~/.kube/<cluster>/.kubernetes_auth` but that's too much detail for here :)

integration/utils.go
vishh: Done

integration/utils.go
vishh: Done

docs/integration.md
vishh: is it `cluster` or `context`?

vmarmol:
LGTM

docs/integration.md
vishh: I updated it to `cluster-name`.

PR #154, opened 2015-03-03
New external sink api

vishh:
The existing GCm sink api is generic enough to support other types of backends. Hence an external Sink API has been created based on the existing GCM driver API.
Added an external sink manager to provide support for running multiple external sinks at the same time - 
one each for monitoring, alerting and storing events.
This is a WIP. I am working on moving the existing influxdb implementation to use the new Sink API. This should make the Influxdb storage model more scalable.
DO NOT MERGE until I complete and test this PR thoroughly.
I am happy to break down this PR if required.


vishh:
@vmarmol @rjnagal. Take a look at this PR when you get a chance. This should make adding new sinks very easy! I can split up the PR into three parts - API, GCM transition, Influxdb + tests if required.

vmarmol:
+1 to splitting up :) the split sounds good and smaller in scope

PR #155, opened 2015-03-04
Update deps and add gofuzz dependency.

vishh:


vishh:
Ping @vmarmol @rjnagal

vmarmol:
LGTM, but can we add a note as to why? :)

vishh:
Required for #154.

PR #156, opened 2015-03-04
Created a unified external sinks api based off the existing gcm infrastructure.

vishh:
Forked from #154.

### DIFF
--- a/sinks/api/decoder.go
+++ b/sinks/api/decoder.go
@@ -1,0 +1,148 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	source_api "github.com/GoogleCloudPlatform/heapster/sources/api"
+)
+
+type defaultDecoder struct {
+	supportedStatMetrics []SupportedStatMetric
+
+	// TODO: Garbage collect data.
+	// TODO: Deprecate this once we the core is fixed to never export duplicate stats.
+	lastExported map[timeseriesKey]time.Time
+}
+
+type timeseriesKey struct {
+	// Name of the metric.
+	Name string
+
+	// Mangled labels on the metric.
+	Labels string
+}
+
+func (self *defaultDecoder) Timeseries(input source_api.AggregateData) ([]Timeseries, error) {
+	var result []Timeseries
+	// Format metrics and push them.
+	for _, pod := range input.Pods {
+		result = append(result, self.getPodMetrics(&pod)...)
+	}
+	result = append(result, self.getContainerSliceMetrics(input.Containers)...)
+	result = append(result, self.getContainerSliceMetrics(input.Machine)...)
+
+	return result, nil
+}
+
+func (self *defaultDecoder) getPodMetrics(pod *source_api.Pod) []Timeseries {
+	// Generate the labels.
+	labels := make(map[string]string)
+	labels[labelPodId] = pod.ID
+	labels[labelLabels] = self.labelsToString(pod.Labels)
+	labels[labelHostname] = pod.Hostname
+
+	// Break the individual metrics from the container statistics.
+	var result []Timeseries
+	for index := range pod.Containers {
+		result = append(result, self.getContainerMetrics(&pod.Containers[index], labels)...)
+	}
+
+	return result
+}
+
+func (self *defaultDecoder) getContainerSliceMetrics(containers []source_api.Container) []Timeseries {
+	labels := make(map[string]string)
+	var result []Timeseries
+	for _, container := range containers {
+		labels[labelHostname] = container.Hostname
+		result = append(result, self.getContainerMetrics(&container, labels)...)
+	}
+
+	return result
+}
+
+// Concatenates a map of labels into a comma-separated key=value pairs.
+func (self *defaultDecoder) labelsToString(labels map[string]string) string {
+	output := make([]string, 0, len(labels))
+	for key, value := range labels {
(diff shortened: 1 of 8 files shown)


sinks/api/types.go
```diff
@@ -0,0 +1,167 @@
...
+		return "double"
+	}
+	return ""
+}
+
+type MetricUnitsType int
+
+const (
+	// A metric with no units.
+	UnitsNone MetricUnitsType = iota
```
vmarmol: Is there such a thing? Everything we measure has to have a unit no?

sinks/api/types.go
```diff
@@ -0,0 +1,167 @@
...
+	// Descriptor of the labels used by this metric.
+	Labels []LabelDescriptor
+
+	// Type and value of metric data.
+	Type      MetricType
+	ValueType MetricValueType
+	Units     MetricUnitsType
+}
+
+type Metric struct {
```
vmarmol: I wanted to rename this since it was getting confusing. This is arguably a Timeseries or a point in time or something along those lines. A lot of the monitoring systems call it "timeseries" so that makes sense. WDYT?

sinks/api/types.go
```diff
@@ -0,0 +1,167 @@
...
+type CompositeData struct {
+	Metrics []Timeseries
+	// TODO: Add map[string]string to support storing raw data like node or pod status, spec, etc.
+	// TODO: Add events.
+}
+
+// Sink is the interface that needs to be implemented by all storage backends.
+type Sink interface {
+	// Stores input CompositeData into the backend.
+	Store(*CompositeData) error
```
vmarmol: What about having Store*() for each type we plan instead of Store of a CompositeData type? Probably doesn't matter either way, but worth considering.

sinks/api/types.go
vmarmol: Ah, I think you want a Count type as well. That takes care of the None usecase above.

sinks/api/decoder.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Mangled labels on the metric.
+	Labels string
+}
+
+// TODO(vmarmol): Paralellize this.
+func (self *defaultDecoder) Metrics(input interface{}) ([]Timeseries, error) {
+	data, ok := input.(source_api.AggregateData)
+	if !ok {
+		return nil, fmt.Errorf("requesting unrecognized type to be stored in GCM")
```
vmarmol: remove mention to GCM

sinks/api/types.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Returns debuf information specific to the sink.
+	DebugData() string
+}
+
+// Codec represents an engine that translated from sources.api to sink.api objects.
+type Decoder interface {
+	// Metrics returns the metrics found in input as a timeseries slice.
+	Metrics(input interface{}) ([]Timeseries, error)
```
vmarmol: Can we try to make this of a stronger type than interface{}? I think we have a pretty good idea of what those could be today.

sinks/api/decoder.go
```diff
@@ -0,0 +1,152 @@
...
+
+type timeseriesKey struct {
+	// Name of the metric.
+	Name string
+
+	// Mangled labels on the metric.
+	Labels string
+}
+
+// TODO(vmarmol): Paralellize this.
```
vmarmol: We can remove this TODO, the pushing is what we should do in parallel :)

sinks/api/decoder.go
```diff
@@ -0,0 +1,152 @@
...
+		if stat == nil {
+			continue
+		}
+		// Add all supported metrics that have values.
+		for index, supported := range self.supportedStatMetrics {
+			key := timeseriesKey{
+				Name:   supported.Name,
+				Labels: labelsAsString,
+			}
+			if data, ok := self.lastExported[key]; ok && data.After(stat.Timestamp) {
```
vmarmol: Add a TODO to remove this when we fix the issue.

sinks/api/types.go
vmarmol: Or MetricValue?

sinks/api/types.go
vmarmol: We can have MetricsFrom<type> functions here too. It'll be easier to extend :)

sinks/api/types.go
vishh: Calling it a "Point". A timeseries here is a wrapper around a point and descriptor.

sinks/api/types.go
vishh: I want it to be an interface{} because not all backends can handle all input types. For example events cannot be handled by GCM.

sinks/api/types.go
vishh: As of today we understand stats, but events and config data are not yet defined. That is the reason for generic interface. Is there a reason why you are suggesting a stronger type? Testing?

sinks/api/types.go
vmarmol: MetricPoint?

sinks/api/types.go
vmarmol: But the backend should drop those requests silently instead of having an overly generic interface. It's fine that some backends don't handle it.

sinks/api/types.go
vmarmol: Stronger interfaces make it easier for the implementer and for the person writing the code. With an interface I have no idea what types of inputs I get and what I need to support. A strong type tells me exactly what that is and I can choose what to do if I can't handle it.

sinks/api/decoder.go
vishh: Done

sinks/api/decoder.go
vishh: Done

sinks/api/decoder.go
vishh: Done

sinks/api/types.go
vishh: `Count` make sense

vishh:
PTAL @vmarmol

vmarmol:
Only remaining comments are around the interfaces that use interface{}

sinks/api/types.go
vishh: Ack.

sinks/api/types.go
vishh: Ack.

vishh:
Removed the interfaces.

vmarmol:
LGTM, will merge on green.

vmarmol:
Looks like we're missing the gofuzz dep?

vishh:
Not sure why thats happening. I added gofuzz in a previous PR and it is
there on HEAD.

PR #162, opened 2015-03-13
Fix heapster startup script.

vishh:
It does not work in certain modes without this PR


rjnagal:
LGTM, although I don't understand half of the mysterious ways of these scripts :)

PR #163, opened 2015-03-13
Adding Influxdb sequence number generator

vishh:
This is required for proper ordering of cpu cumulative stats. Without this computing instantaneous cpu usage seems to be error prone.

### DIFF
--- a/sinks/influxdb/sequence_number.go
+++ b/sinks/influxdb/sequence_number.go
@@ -1,0 +1,40 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package influxdb
+
+import (
+	"sync"
+)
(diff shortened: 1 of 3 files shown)


vmarmol:
nit: Can we add a "why" to the commit message? :)

vishh:
The actual patch has the reason.

sinks/influxdb/sequence_number.go
```diff
@@ -0,0 +1,43 @@
...
+	"sync"
+)
+
+type metricSequenceNum interface {
+	Get(metric string) uint64
+}
+
+type realMetricSequenceNum struct {
+	store map[string]uint64
+	sync.Mutex
```
vmarmol: nit: don't embed else the Lock() and Unlock() are public members of the struct.

vishh:
and updated the PR description @vmarmol

sinks/influxdb/sequence_number.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (self *realMetricSequenceNum) Get(metric string) uint64 {
+	self.Lock()
+	defer self.Unlock()
+	if _, ok := self.store[metric]; !ok {
```
vmarmol: You shouldn't need this since the empty value is 0.

vmarmol:
Minor nits, LGTM otherwise. Thanks for the update! I missed it :)

vmarmol:
LGTM, will merge on green. Thanks @vishh!

sinks/influxdb/sequence_number.go
vishh: Ahh din't realize that. Done

sinks/influxdb/sequence_number.go
vishh: Done

vmarmol:
Merging.

PR #165, opened 2015-03-17
Update InfluxDB doc, cAdvisor is no longer a pod.

vmarmol:


rjnagal:
LGTM, ignoring build failure for doc change (cluster is already setup when integration test starts).

PR #168, opened 2015-03-19
deploy: update INFLUXDB_HOST including the port

simon3z:
This patch adds the port to the influxdb host and removes the url path (/api/v1beta1/proxy/...) that is already included inside the run.sh script of heapster_grafana:v0.5.


vishh:
ok to test

vishh:
Thanks for the fix @simon3z!

PR #169, opened 2015-03-20
Fix grafana service config. It is currently using an invalid selector.

vishh:
Updated the v1beta1 configs to be in sync with v1beta3. May be its time to have a script dynamically generate the configs instead of maintaining one.


vishh:
cc @vmarmol @rjnagal

vmarmol:
LGTM.

PR #171, opened 2015-03-20
Making heapster resilient to kubelet failures.

### LINKED ISSUE #170: Make heapster resilient to kubelet failures
As of now heapster fails stats collection completely if kubelet fails to return stats for a container. This is undesirable. heapster should log error and move on instead of failing completely.

vishh:
Improved error handling of kubelet http requests.

Fixes #170


vmarmol:
LGTM

Issue #175, opened 2015-03-20
Add an integration test for GCM sink


vmarmol:
Assigning to me :)

vishh:
cc @piosz.

piosz:
I can take this (with medium priority).

piosz:
closing as obsolete

PR #176, opened 2015-03-20
Apply GCM request policies.

vmarmol:
- Round metric timestamps to 1s.
- Only send strictly greater than metrics.
- Don't send more than 200 per request.
- Only one metric of type per request.

Filed #175 for an integration test.

### DIFF
--- a/sinks/gcm/driver.go
+++ b/sinks/gcm/driver.go
@@ -142,12 +142,6 @@ const maxTimeseriesPerRequest = 200
 
 // Pushes the specified metric values in input. The metrics must already exist.
 func (self *gcmSink) StoreTimeseries(input []sink_api.Timeseries) error {
-	// Check we're not being asked to write more timeseries than we can.
-	// TODO: Why not split the input into multiple batches.
-	if len(input) > maxTimeseriesPerRequest {
-		return fmt.Errorf("unable to write more than %d metrics at once and %d were provided", maxTimeseriesPerRequest, len(input))
-	}
-
 	// Ensure the metrics exist.
 	for _, entry := range input {
 		metric := entry.Point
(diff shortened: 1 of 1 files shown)


sinks/gcm/driver.go
```diff
@@ -42,6 +42,15 @@ type gcmSink struct {
 	// TODO(vmarmol): Also store labels?
 	// Map of metrics we currently export.
 	exportedMetrics map[string]sink_api.MetricDescriptor
+
+	// TODO(vmarmol): Garbage collect these.
+	// Time we last exported a particular timeseries.
+	lastExportedTime map[lastExportedKey]time.Time
```
vishh: This might be worth doing in the [central decoder](https://github.com/GoogleCloudPlatform/heapster/blob/master/sinks/api/decoder.go#L102)

vmarmol: We already do, but this is used here (again) because GCM rounds to seconds. We can enforce that on all backends or re-do this here where it is enforced.

sinks/gcm/driver.go
```diff
@@ -185,6 +203,39 @@ func (self *gcmSink) StoreTimeseries(input []sink_api.Timeseries) error {
...
+
+	return lastErr
+}
+
+func (self *gcmSink) pushMetrics(request *metricWriteRequest) error {
+	if len(request.Timeseries) == 0 {
+		return nil
+	}
+	if len(request.Timeseries) > maxTimeseriesPerRequest {
+		return fmt.Errorf("unable to write more than %d metrics at once and %d were provided", maxTimeseriesPerRequest, len(request.Timeseries))
```
vishh: Should we add a TODO to batch writes?

sinks/gcm/driver.go
vishh: I think we can round to seconds everywhere.

sinks/gcm/driver.go
vmarmol: This is just a sanity check, by now we should have done this.

vishh: True. Once we start adding custom metrics, wouldn't we hit this condition?

vishh:
Ping @vmarmol.

vishh:
Ping @vmarmol

vmarmol:
Pushed a new version PTAL @vishh

vishh:
LGTM.

PR #180, opened 2015-03-22
Allow flags with spaces

agios:
So that multiple options can be passed in


vishh:
ok to test

vishh:
@agios Thanks for the PR! LGTM. Will merge once the tests pass.

PR #142, opened 2015-02-27
InfluxDB is now proxied by Nginx for Grafana seamless integration.

pires:
Fixes #135 

Grafana section on ```influx-grafana-controller.yaml``` should be like this now
```
yaml
- name: grafana
    image: kubernetes/heapster_grafana:v0.4
        env: 
            - name: "HTTP_USER"
              value: "admin"
            - name: "HTTP_PASS"
              value: "admin"
```
or something like (these are the default values, by the way)
```yaml
- name: grafana
    image: kubernetes/heapster_grafana:v0.4
        env: 
            - name: "HTTP_USER"
              value: "admin"
            - name: "HTTP_PASS"
              value: "admin"
            - name: "INFLUXDB_PROTO"
              value: "http"
            - name: "INFLUXDB_HOST"
              value: "localhost"
            - name: "INFLUXDB_PORT"
              value: "8086"
```

PR #182, opened 2015-03-23
grafana: simplify docker image using database urls

simon3z:


rjnagal:
ok to test

deploy/kube-config/influxdb-grafana-controller.yaml
```diff
@@ -29,17 +29,9 @@ desiredState:
...
-                value: "https"
-              - name: "INFLUXDB_HOST"
-                value: "localhost"
-              - name: "INFLUXDB_PORT"
-                value: ""
-              - name: "KUBERNETES_API_PORT"
-                value: ""
+              - name: "INFLUXDB_METRICS_URL"
+                value: "/api/v1beta3/proxy/namespaces/default/services/monitoring-influxdb/db/k8s"
+              - name: "INFLUXDB_GRAGANA_URL"
```
vishh: nit: s/GRAGANA/GRAFANA

deploy/kube-config/v1beta3/influxdb-grafana-controller.yaml
```diff
@@ -1,40 +1,36 @@
...
-              value: '"+window.location.protocol.slice(0,-1)+"'
-            - name: "INFLUXDB_HOST"
-              value: '"+window.location.host+"'
-            - name: "INFLUXDB_PORT"
-              value: ""
-            - name: "KUBERNETES_API_PORT"
-              value: ""
+            - name: "INFLUXDB_METRICS_URL"
+              value: "/api/v1beta3/proxy/namespaces/default/services/monitoring-influxdb/db/k8s"
+            - name: "INFLUXDB_GRAGANA_URL"
```
vishh: nit: s/GRAGANA/GRAFANA

simon3z:
@vishh note that you need a new kubernetes/heapster_grafana image: v0.6

I have some other changes in mind, for example, what do you think about the possibility to revamp the shell script (in possibly just a single one)?

If you're ok with that you can hold this PR and I'll go ahead with larger changes tomorrow (so I'll spare you to build the docker image twice).

simon3z:
Oops previous question was for @rjnagal  (but for @vishh as well eventually)

vishh:
@simon3z: Thanks for this PR. The shell scripts are painful to maintain. If I get some time I would to rewrite all of them in go. That said, any changes to make the scripts easier to manage are welcome :+1:

rjnagal:
:+1: holding PR SG.

jimmidyson:
I use a grafana docker image at https://github.com/jimmidyson/docker-grafana which uses go http server to serve files, but currently uses direct browser connection to influxdb service. I need to add in a reverse proxy to influxdb.

Using the built in kubernetes master proxy is not great for environments that are secured & require credentials to use the kubernetes master proxy. I can try to add in the reverse proxy tomorrow if you're interested?

vishh:
@jimmidyson: The reason for switching to the proxy is to provide an off-the-box experience with Kubernetes. As you stated, it is clearly not a production solution. #142 was an attempt to address that. It would be great if you can get the reverse proxy setup to work along with the apiserver proxy access pattern.
Monitoring is being considered as a core Kubernetes service and that is the reason for re-using master auth. But since the current state of auth in the master is very primitive, it makes sense to not rely on that.
Your go based setup looks much simpler that the nginx setup. I have no issues dropping the nginx setup in favor of your go based implementation.

simon3z:
@vishh @rjnagal I am little confused by the fact that these scripts are available also in kubernetes here:

https://github.com/GoogleCloudPlatform/kubernetes/tree/master/cluster/addons/cluster-monitoring

Maintaining them in two places is painful, do you think we can centralized them here?

jimmidyson:
@vissh why do you need to use the apiserver proxy rather than just a reverse proxy to the kube service directly?

vishh:
@simon3z: The reason for having a separate copy in kubernetes was for bundling reasons. A specific version of kubernetes will always have a monitoring setup that works for that version and is tested continuously using e2e tests.
I can add some documentation in kube repo to request users to post their updates here and then periodically sync those changes to the kube repo. WDYT?

vishh:
@jimmidyson: The apiserver proxy is being used mainly for external access to Grafana. By reverse proxy are you referring to the intra-cluster link between influxdb and grafana?

jimmidyson:
@vishh - yeah. So I've updated https://github.com/jimmidyson/docker-grafana to create a RP, similar to the nginx config one of the other PRs does, to the env var specified influxdb service so there's no need to go through the kube master proxy. I'll send in a PR for updating the Grafana image so you can take a look.

vishh:
@simon3z: Should we close this PR in favor of #187 ?

pires:
@jimmidyson please see #142 where I also made `nginx` - which already is used to serve `grafana` - to reverse-proxy `influxdb`.

I find it weird that @vishh is open to accept #183 when he says that #142 is a basic PR and gives all sort of (also weird) justifications not to accept it.

vishh:
@pires: I think there has been some miscommunication. My only issue with #142 was that, it made it not possible to access Grafana from outside the cluster using the apiserver proxy. The RP changes are needed for sure. I believe I had the very same discussion with @jimmidyson and we agreed to not break that.

simon3z:
Closing this because I moved to https://github.com/GoogleCloudPlatform/heapster/pull/187

Issue #135, opened 2015-02-26
Proxy calls to InfluxDB with nginx


pires:
So, having something like
```
yaml
- name: "INFLUXDB_HOST"
  value: '"+window.location.hostname+"/api/v1beta1/proxy/services/monitoring-influxdb'
```
is very limitative when it comes to running ```influxdb``` + ```grafana```, especially inside Kubernetes.

See [my suggested solution](https://github.com/GoogleCloudPlatform/kubernetes/issues/4841).

vishh:
+1 to making this change. As long as it works on most kubernetes cluster
setups, I am fine with that change.

pires:
In progress..

Issue #132, opened 2015-02-25
Support persistent storage for InfluxDB backend.


vishh:
As of now InfluxDB pod uses local storage.

pires:
+1

vishh:
Excepting local disk none of the existing Storage solutions in kubernetes works for all the users. I am thinking the best way to address this issue is to provide a cookbook of storage recipes in the repo and make the users choose one based on their deployment. 
We can simplify this by having a deployment binary with simple flags.
Any thoughts @pires ?

jimmidyson:
I think by default we should be using an empty-dir volume, but as we're seeing with lots of things this is deployment/implementation specific. A deployment binary might be a good idea, but we really need this to be in the influxdb manifest at the very least.

pires:
I agree with @jimmidyson. I think it's up to the user to adopt whatever storage strategy he/she is more comfortable with. In the end, I'd really like to see something like [`flocker`](https://github.com/ClusterHQ/flocker) coming into Kubernetes, but right now it's too early to decide on an approach other than empty-dir.

piosz:
closing in favor of #1197

PR #183, opened 2015-03-24
Switch to golang http server with reverse proxy to influxdb service

jimmidyson:
Following discussion in #182, #132, #135.


rjnagal:
ok to test

grafana/Dockerfile.dev
```diff
@@ -0,0 +1,26 @@
+FROM gliderlabs/alpine:3.1
```
vishh: Why add this file?

grafana/Makefile
```diff
@@ -0,0 +1,24 @@
+NAME=heapster_grafana
+VERSION=$(shell cat VERSION)
+
+dev:
+	@docker build -f Dockerfile.dev -t $(NAME):dev .
+	@docker run --rm \
+		-v $(PWD):/go/src/github.com/GoogleCloudPlatform/heapster/grafana \
+		-e INFLUXDB_SERVICE_NAME=INFLUXDB \
+		-e INFLUXDB_SERVICE_HOST=172.30.17.250 \
```
vishh: For testing purposes, can we can spawn an influxdb container and use that instead?

grafana/Dockerfile.dev
vishh: nvm. I see a testing use case below. Should we move to this under a '/test' directory?

jimmidyson: It's more of a convenience for doing builds for those with docker but
without full go dev environment set up locally.

grafana/README.md
```diff
@@ -1,87 +1,43 @@
...
+```yaml
+containers:
+  - name: grafana-image
+    image: kubernetes/heapster_grafana:latest
+    ports:
+      - name: grafana-port
+        containerPort: 3000
+    env:
+      - name: INFLUXDB_URL
+        value: http://${INFLUX_MASTER_SERVICE_HOST}:${INFLUX_MASTER_SERVICE_PORT}
```
vishh: Runtime expansion sounds awesome. But DNS should take care of this. So in kubernetes, this could just be the Influxdb service name.

jimmidyson:
If you wouldn't mind holding off on this, I'm looking at putting together a generic go utility that serves static content, processes config template to file & proxies to kubernetes service. I see this as pretty generic & useful.

grafana/main.go
```diff
@@ -0,0 +1,79 @@
...
+	return env
+}
+
+func main() {
+	influxDbUrl := os.Getenv("INFLUXDB_URL")
+	if len(influxDbUrl) == 0 && len(os.Getenv("INFLUXDB_SERVICE_NAME")) > 0 {
+		os.Setenv("INFLUXDB_URL", "")
+		influxDbServiceUrl := "${INFLUXDB_PROTO}://${" +
+			os.Getenv("INFLUXDB_SERVICE_NAME") + "_SERVICE_HOST}:${" +
+			os.Getenv("INFLUXDB_SERVICE_NAME") + "_SERVICE_PORT}/"
```
vishh: We can take in as input the influxdb service name and rely on DNS instead. That approach avoids requiring an influxdb service to running prior to deploying Grafana (This is very useful when run `kubectl.sh create -f configs/`)

grafana/main.go
```diff
@@ -0,0 +1,79 @@
...
+	env := make(map[string]string)
+	for _, i := range os.Environ() {
+		sep := strings.Index(i, "=")
+		env[i[0:sep]] = i[sep+1:]
+	}
+	return env
+}
+
+func main() {
+	influxDbUrl := os.Getenv("INFLUXDB_URL")
```
vishh: "INFLUXDB_URL" is not specified in the dockerfile. Is this intentional? Can this be defaulted to the influxdb service name?

jimmidyson:
Yes we can default & use dns too. Dns wasn't available as core service when I first put this together but makes sense to use it now.

vishh:
@jimmidyson: Overall it looks good. I will not merge this PR until you are done with your re-factoring.
Thanks for the PR!

jimmidyson:
@vishh @pires @simon3z - I have updated this image to use `kuisp` (https://github.com/jimmidyson/kuisp) which I've built for this purpose: serve static content & proxy to kubernetes services. It greatly simplifies the scripts while allowing the same functionality. Please review kuisp if you feel you need to.

This gives heapster grafana a more easily configurable image than nginx/apache, as well as being very lightweight & small image.

pires:
Thanks @jimmidyson

grafana/build.sh
```diff
@@ -1,5 +0,0 @@
-#! /bin/bash
-
-pushd $( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
vishh: Any reason to delete this file?

vishh:
@jimmidyson: I tried testing your PR and it does not work for me. My request is to have the default config work via the apiserver proxy, that way it works for everyone. Users can then make some changes to either use a public IP for the grafana service or create an ELB in place of the proxy. Specifically, if the default setup can be accessible via `<masterip>/api/v1beta1/proxy/services/monitoring-grafana/`, that will be awesome. If that is not possible, I am open to discussing how to provide an alternative off-the-box experience.

Thanks for working on this PR!

jimmidyson:
@vishh what do you receive in your browser when you try it?

Accessing HTML pages via the api server proxy is not really a good idea due to the amount of rewriting the HTML that is required to sort stuff like absolute links, hard coded base hrefs, etc.

grafana/config.js.tmpl
```diff
@@ -0,0 +1,27 @@
+define(['settings'],
+function (Settings) {
+  return new Settings({
+    datasources: {
+      'metrics': {
+        type: 'influxdb',
+        url: "/db/{{ .Env.INFLUXDB_NAME }}",
```
vishh: The 'url' needs to be that of where Grafana is accessible (ex. :`https://<master>/api/v1beta1/proxy/services/monitoring-grafana`). 
The intention is to make Grafana client side re-use its URL for accessing influxdb.

grafana/Dockerfile
```diff
@@ -1,29 +1,33 @@
...
-ENV HTTP_PASS **Random**
+ENV KUISP_VERSION 0.3
 
+ENV GRAFANA_VERSION 1.9.1
+ENV INFLUXDB_NAME k8s
+ENV GRAFANA_DB_NAME grafana
+ENV GRAFANA_DEFAULT_DASHBOARD /dashboard/file/kubernetes.json
 ENV INFLUXDB_PROTO http
-ENV INFLUXDB_HOST localhost
+ENV INFLUXDB_HOST influxdb
```
vishh: So I see two separate config requirements here.
1. How to access Influxdb internally from Grafana container.
2. How to access Influxdb externally from outside the cluster (which is that of reaching back to the proxy endpoint in the Grafana container).
IIUC, both of those configs are conflated now.

vishh:
@jimmidyson: I tried testing this PR by adding a `hostPort` to Grafana container. I get a connection reset error `curl: (56) Recv failure: Connection reset by peer`. 
I agree with you on the fact that the proxy is not an ideal solution. I am open to suggestions on making Grafana accessible from outside the cluster without using the apiserver proxy.

pires:
What about using `publicIPs` flag from Services? This is what I've been advocating all the time.

jimmidyson:
Accessing services remotely in a consistent way across environments/clouds
is pretty impossible IMO. Public ips is probably the easiest way to achieve
that but as the ip will vary across envs it makes deployment a bit harder
than simply deploying a few RCs & services - needs user input to deploy.

Having said that, it's the only way to get a nice cross-env UX where users
can hit a simple url, without remembering /api/v1beta1/.... Users are gonna
want to set up a dns entry to view grafana directly.

I vote for publicips option but pointing to documentation on different
cloud set ups for accessing kubernetes services, which grafana will just be
another one.

jimmidyson:
@vishh - if you've set the host port properly, the pod is actually running & you can route to the node then I don't understand how you get connection refused. The single container process is the web server - of the pod is running then it's serving on port 8080...

pires:
@jimmidyson you can also go with `createrExternalLoadBalancer` but that requires a cloud-provider.

jimmidyson:
Yeah so that's the other option: document how to deploy on different clouds/envs but that's not the consistent ux that @vishh is after I don't think.

vishh:
@jimmidyson: I din't notice that the server is bound to '8080' instead of port '80'. Trying again.
@pires: Agreed. PublicIP is definitely the best option. But it requires manual configuration. The only reason for switching over the proxy was to give a simple ux to users. Power users like yourself, know to switch over to Public IPs or ELBs. 
@jimmidyson: My hope was for the default setup to work via the proxy and have alternative (better) deployment configs, with documentation, that will use ELBs or Public IPs. Is that too much to ask for or is it difficult to achieve with your new model?

jimmidyson:
@vishh - reason for 8080 is to run as non-root user (nobody) in container for extra security.

Iirc how the proxy works it should still be accessible through the api server proxy, but I'm not sure how the grafana ui will look with the html rewrites that the api server proxy still probably need to do.

vishh:
@jimmidyson: Ack. It works perfectly without the proxy. I am trying to debug why access via the proxy fails.

vishh:
I figured out the issue with the proxy. I am merging this PR as-is. I will fix the proxy stuff in a subsequent PR.

Thanks for the contributions @jimmidyson and thanks for your input @pires.

PR #184, opened 2015-03-24
Kubernetes yaml file updates

simon3z:


simon3z:
@vishh @rjnagal this is urgent as it's related to https://github.com/GoogleCloudPlatform/kubernetes/issues/5850

vmarmol:
ok to test

vmarmol:
This looks good, although I have a feeling that it will fail the integration test since we have an older release on Kubernetes up.

PR #187, opened 2015-03-24
Grafana services and docker image update

simon3z:


simon3z:
/cc @vishh @jimmidyson 

If we have an agreement on using apache + its reverse proxy I can go ahead and complete that as well.

vishh:
ok to test

vishh:
LGTM. Thanks for the cleanup @simon3z. Will merge once the tests pass.
@simon3z: Lets discuss about apache vs native go server in #183.

vishh:
@simon3z: #183 acheives the same goals as this PR. It includes the reverse proxy changes as well. Can you take a look at that and see if that works for you?

vishh:
@simon3z: I merged #183 yesterday which adds a RP to access Influxdb. Can you see if that works for you?

Issue #118, opened 2015-02-20
Collect and export kubernetes events and docker events 


vishh:
InfluxDB supports events natively. Other backends might not. We can place structure around some specific events like container starts, deletions, restarts, etc and export them as metrics, but that might be sub-optimal.
Need to explore how events is currently handled by various monitoring/alerting systems.

vishh:
@saad-ali is working on this.

saad-ali:
#252

PR #192, opened 2015-03-25
Add events source to heapster

saad-ali:
This implements the first part of #118 Follow up PR will handle events from the sink side.

Related Kubernetes Issues:

### DIFF
--- a/sources/events.go
+++ b/sources/events.go
@@ -1,0 +1,177 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sources
+
+import (
+	"errors"
+	"fmt"
+
+	kubeapi "github.com/GoogleCloudPlatform/kubernetes/pkg/api"
+	kubeclient "github.com/GoogleCloudPlatform/kubernetes/pkg/client"
+	kubefields "github.com/GoogleCloudPlatform/kubernetes/pkg/fields"
+	kubelabels "github.com/GoogleCloudPlatform/kubernetes/pkg/labels"
+	kubewatch "github.com/GoogleCloudPlatform/kubernetes/pkg/watch"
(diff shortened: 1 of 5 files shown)


vmarmol:
ok to test

sources/events.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+			if eventsUpdate.events.Items == nil {
+				return nil, errors.New("Error: received an event list with nil Items.")
+			}
+			for _, event := range eventsUpdate.events.Items {
+				glog.V(3).Infof("Received new event: %#v\r\n", event)
+				events = append(events, event)
+			}
+		default:
+			break UpdateLoop
```
vishh: nit: Why not return here?

sources/events.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return events, nil
+}
+
+// watchLoop loops forever looking for new events.  If an error occurs it will close the channel and return.
+func watchLoop(eventClient kubeclient.EventInterface, eventsChan chan<- eventsUpdate) {
+	defer close(eventsChan)
+	events, err := eventClient.List(kubelabels.Everything(), kubelabels.Everything())
+	if err != nil {
+		glog.Errorf("Failed to load events: %v", err)
```
vishh: In case of an error, we should bubble up the error and try again.

sources/events.go
saad-ali: Instead of having two identical return statements, I prefer having one return statement for the method. But, no big deal either way. If you feel strongly, I can change it.

sources/kube.go
```diff
@@ -185,9 +186,14 @@ func (self *kubeSource) GetInfo() (api.AggregateData, error) {
 		return api.AggregateData{}, err
 	}
 	glog.V(2).Info("Fetched list of nodes from the master")
+	events, err := self.eventsApi.GetEvents()
+	if err != nil {
+		return api.AggregateData{}, err
```
vishh: Should we continue and return stats even if we cannot retrieve events?
The existing code also needs to change as well. So its fine to deal with this in a separate PR.

sources/events.go
vishh: Ack. I don't mind leaving it as-is.

vishh:
Build is failing on Travis.

sources/events.go
saad-ali: Implemented logic to restart watch loop if it fails. However this is not very robust. As discussed in person we will revisit this, as it's a bigger issue.

sources/kube.go
saad-ali: We can address this as a separate issue.

vishh:
LGTM. Adding a unit/e2e test will be helpful in the long run. Again not required for this PR.
I will merge as soon as the unit tests are fixed.

saad-ali:
PTAL

PR #195, opened 2015-03-27
Add kubernetes master authentication args & config to kube source

jimmidyson:


vishh:
ok to test.

vishh:
@jimmidyson: Will auth be necessary even if heapster were to use the readonly master service `kubernetes-ro`?

jimmidyson:
@vishh - it depends on the implementation. I know at least one that requires authentication for both `kubernetes` & `kubernetes-ro`. Even the RO service can expose info... We at least need the capability to auth, regardless.

jimmidyson:
@vishh - btw this failed earlier because of the github outage :( it's not building again for some reason.

vishh:
@jimmidyson: Suggestion. Can we use the [kube clientauth](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/clientauth/clientauth.go) library instead? @erictune is working on introducing secrets into containers and using the clientauth library will be necessary soon.

jimmidyson:
@vishh - yes this makes a lot more sense. I was unaware of this coming but I like it :) Are there any timescales for when this will be supported in a kubernetes release?

vishh:
@etune mentioned that he was going to start working on it today. So I
expect to see something in a week on Kubernetes HEAD.

jimmidyson:
@vishh - ok i'll update this PR to use `clientauth` & we can wait to merge it until it's usable in Kubernetes?

vishh:
SGTM

jimmidyson:
@vishh - i guess we could merge this now as it defaults to disabled & only use when introducing secrets to containers is sorted in kubernetes?

vishh:
LGTM. 
@jimmidyson: Agreed. Merging this PR now. Thanks for the PR!

PR #201, opened 2015-03-28
Add back in command line insecure flag client config

jimmidyson:
Missed when I added the kube clientauth piece - useful to be able to override insecure flag with command flag.


rjnagal:
ok to test

rjnagal:
LGTM.

Will merge after test comes back.

PR #202, opened 2015-03-29
Use node legacy host ip on older kubernetes versions

jimmidyson:


vishh:
Just to clarify, the current assumption is that older kubernetes versions should not be using the latest heapster. We bundle a specific version of heapster with a kubernetes release. We can re-think that strategy in the future once the kube API and some of the internal code stabilizes.

vishh:
ok to test

vishh:
LGTM. Will merge once the tests pass.

vishh:

PR #205, opened 2015-03-31
Push events to influxdb heapster sink

saad-ali:
This implements the second part of #118 (follow up to PR #192).

Related Kubernetes Issues:

Screenshot showing events in InfluxDB (pulled by heapster from Kubernetes API server and pushed by heapster to influxdb):
influxdbevents

### DIFF
--- a/sinks/api/eventdecoder.go
+++ b/sinks/api/eventdecoder.go
@@ -1,0 +1,85 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	"encoding/json"
+
+	source_api "github.com/GoogleCloudPlatform/heapster/sources/api"
+	kube_api "github.com/GoogleCloudPlatform/kubernetes/pkg/api"
+)
+
+const (
+	EventsSeriesName = "log/events"
+)
+
+type eventDecoder struct{}
+
+func NewEventDecoder() Decoder {
+	return &eventDecoder{}
+}
+
+func (self *eventDecoder) Timeseries(data source_api.AggregateData) ([]Timeseries, error) {
+	result := []Timeseries{}
+	if data.Events != nil && len(data.Events) > 0 {
+		points := []Point{}
+		for _, event := range data.Events {
+			value, err := getEventValue(&event)
+			if err != nil {
+				return result, err
+			}
+			points = append(points, Point{
+				Labels: getEventLabels(&event),
+				Start:  event.LastTimestamp.Time,
+				End:    event.LastTimestamp.Time,
+				Value:  value,
+			})
+		}
+		result = append(result, Timeseries{
+			SeriesName:    EventsSeriesName,
+			TimePrecision: Millisecond,
+			LabelKeys:     SupportedLabelKeys(),
+			Points:        points,
+		})
+	}
+
+	return result, nil
+
+}
+
+// Generate the labels.
+func getEventLabels(event *kube_api.Event) map[string]string {
+	labels := make(map[string]string)
+	if event == nil {
+		return nil
+	}
+
+	labels[labelHostname] = event.Source.Host
+	if event.InvolvedObject.Kind == "Pod" {
+		labels[labelPodName] = event.InvolvedObject.Name
+		labels[labelPodId] = string(event.InvolvedObject.UID)
+	}
+
+	return labels
+}
+
+// Generate point value for event
+func getEventValue(event *kube_api.Event) (string, error) {
+	bytes, err := json.MarshalIndent(event, "", " ")
+	if err != nil {
+		return "", err
+	}
+	return string(bytes), nil
+}
--- a/sinks/api/metricdecoder.go
+++ b/sinks/api/metricdecoder.go
@@ -15,12 +15,13 @@
 package api
 
 import (
+	"fmt"
 	"time"
 
 	source_api "github.com/GoogleCloudPlatform/heapster/sources/api"
 )
 
-type defaultDecoder struct {
+type metricDecoder struct {
 	supportedStatMetrics []SupportedStatMetric
 
 	// TODO: Garbage collect data.
@@ -36,7 +37,7 @@ type timeseriesKey struct {
 	Labels string
 }
 
-func (self *defaultDecoder) Timeseries(input source_api.AggregateData) ([]Timeseries, error) {
+func (self *metricDecoder) Timeseries(input source_api.AggregateData) ([]Timeseries, error) {
 	var result []Timeseries
 	// Format metrics and push them.
 	for index := range input.Pods {
@@ -49,16 +50,17 @@ func (self *defaultDecoder) Timeseries(input source_api.AggregateData) ([]Timese
 }
 
 // Generate the labels.
-func (self *defaultDecoder) getPodLabels(pod *source_api.Pod) map[string]string {
+func (self *metricDecoder) getPodLabels(pod *source_api.Pod) map[string]string {
 	labels := make(map[string]string)
 	labels[labelPodId] = pod.ID
+	labels[labelPodName] = pod.Name
 	labels[labelLabels] = LabelsToString(pod.Labels, ",")
 	labels[labelHostname] = pod.Hostname
 
 	return labels
 }
 
-func (self *defaultDecoder) getPodMetrics(pod *source_api.Pod) []Timeseries {
+func (self *metricDecoder) getPodMetrics(pod *source_api.Pod) []Timeseries {
 	// Break the individual metrics from the container statistics.
 	result := []Timeseries{}
 	for index := range pod.Containers {
@@ -69,7 +71,7 @@ func (self *defaultDecoder) getPodMetrics(pod *source_api.Pod) []Timeseries {
 	return result
 }
 
-func (self *defaultDecoder) getContainerSliceMetrics(containers []source_api.Container) []Timeseries {
+func (self *metricDecoder) getContainerSliceMetrics(containers []source_api.Container) []Timeseries {
 	labels := make(map[string]string)
 	var result []Timeseries
 	for index := range containers {
@@ -88,7 +90,7 @@ func copyLabels(labels map[string]string) map[string]string {
 	return c
 }
 
-func (self *defaultDecoder) getContainerMetrics(container *source_api.Container, labels map[string]string) []Timeseries {
+func (self *metricDecoder) getContainerMetrics(container *source_api.Container, labels map[string]string) []Timeseries {
 	if container == nil {
 		return nil
 	}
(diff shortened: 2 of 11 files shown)


rjnagal:
ok to test

vishh:
Discussed with @saad-ali offline about the schema. He is working on updating the PR to use a more user friendly schema.

saad-ali:
PTAL. I refactored the decoder such that there are separate decoders for events and metrics--they create common `timeseries` objects that the drivers (gcm and influxdb) can consume.
And based on your feedback, events and metrics now share the common schema (I added "pod_name").

New screenshot showing events in InfluxDB (pulled by heapster from Kubernetes API server and pushed by heapster to influxdb):
influxdb_events02

jimmidyson:
Can't we just store the marshalled event object as JSON rather than a string which is much harder to parse for display, etc?

vishh:
@saad-ali: I was thinking of just saving the event message as value. But it looks like you are storing a lot more data that that. If we think we need to store the entire event object, then I agree with @jimmidyson on marshalling the entire object into json. Sorry about the confusion.

saad-ali:
I think there's other information in there that is useful. I'll switch it to JSON.

jimmidyson:
Or we could create columns from the event data instead?

vishh:
@jimmidyson: Having columns derived from Event object is confusing. I feel that the event object wasn't conceived to be human parsable.

saad-ali:
@jimmidyson The original iteration of this PR was deriving columns from event objects. But after syncing with @vishh we agreed that this was contradictory to heapster's goals of having a common schema for all metrics/events. Hence this iteration.

jimmidyson:
Understand now about the common schema. Is this schema documented anywhere? Surely the storage schema might differ between sinks?

Storing marshalled as JSON should cover what anyone needs to do with it.

saad-ali:
@jimmidyson The idea was to try and maintain a common schema to the extent possible, and throw any additional information in to the value. Sink will always have the option to drop data that is irrelevant to them. The schema is documented here: [docs/storage-schema.md](https://github.com/GoogleCloudPlatform/heapster/blob/master/docs/storage-schema.md)

saad-ali:
PTAL.

New screenshot:
influxdbevents05

sinks/api/types.go
```diff
@@ -115,17 +123,14 @@ type MetricDescriptor struct {
...
 type Point struct {
-	// The name of the metric. Must match an existing descriptor.
-	Name string
-
-	// The labels and values for the metric. The keys must match those in the descriptor.
+	// The label keys and values for the metric.
 	Labels map[string]string
 
-	// The start and end time for which this data is representative.
+	// The start and end time for which this data is representative and the precision of the timestamps.
```
vishh: Precision is not included here.

sinks/gcm/driver.go
```diff
@@ -141,43 +151,52 @@ type metricWriteRequest struct {
...
-func (self *gcmSink) StoreTimeseries(input []sink_api.Timeseries) error {
+func (self *gcmSink) StoreTimeseries(inputTimeseriesLists []*[]sink_api.Timeseries) error {
 	// Ensure the metrics exist.
-	for _, entry := range input {
-		metric := entry.Point
-		// TODO: Remove this check if possible.
-		if _, ok := self.exportedMetrics[metric.Name]; !ok {
-			return fmt.Errorf("unable to push unknown metric %q", metric.Name)
+	for _, inputTimeseriesList := range inputTimeseriesLists {
+		if inputTimeseriesList != nil {
```
vishh: nit: How about reversing the condition and continuing if its nil?

sinks/gcm/driver.go
```diff
@@ -141,43 +151,52 @@ type metricWriteRequest struct {
...
-		for key, value := range metric.Labels {
-			labels[fullLabelName(key)] = value
-		}
-
-		// TODO(vmarmol): Validation and cleanup of data.
-		// TODO(vmarmol): Handle non-int64 data types. There is an issue with using omitempty since 0 is a valid value for us.
-		if _, ok := metric.Value.(int64); !ok {
-			return fmt.Errorf("non-int64 data not implemented. Seen for metric %q", metric.Name)
+	for _, inputTimeseriesList := range inputTimeseriesLists {
+		if inputTimeseriesList != nil {
```
vishh: Ditto as above.

sinks/influxdb/driver.go
```diff
@@ -37,72 +37,130 @@ type influxdbSink struct {
...
 
-func (self *influxdbSink) StoreTimeseries(timeseries []sink_api.Timeseries) error {
+func (sink *influxdbSink) storeTimeseriesWithPrecision(timeseriesLists []*[]sink_api.Timeseries, precision sink_api.TimePrecision) error {
 	dataPoints := []*influxdb.Series{}
-	for index := range timeseries {
-		dataPoints = append(dataPoints, self.metricToSeries(&timeseries[index]))
+	for _, timeseriesList := range timeseriesLists {
+		if timeseriesList != nil {
+			for _, timeseries := range *timeseriesList {
+				if timeseries.TimePrecision == precision {
```
vishh: Why do we need this check? Why can't we use the precision that is passed via `timeseries.TimePrecision`?

sinks/influxdb/driver.go
```diff
@@ -37,72 +37,130 @@ type influxdbSink struct {
...
+		err := sink.client.WriteSeriesWithTimePrecision(dataPoints, getInfluxdbPrecision(precision))
+		if err != nil {
+			glog.Errorf("failed to write time series to influxDB - %s", err)
+			for _, dataPoint := range dataPoints {
+				glog.Errorf("  detail: failed to write time series (%s) to influxDB with columns (%s)", dataPoint.Name, dataPoint.Columns)
+			}
+			sink.recordWriteFailure()
+			return err
+		}
+		glog.V(1).Info("Succesfully wrote %q timeseries to influxDB", precision)
```
vishh: This log file is misleading. Why are we logging precision in place of number of timeseries?

sinks/api/types.go
```diff
@@ -148,24 +153,35 @@ type SupportedStatMetric struct {
...
-	Point            *Point
-	MetricDescriptor *MetricDescriptor
+	// The name of the series.
+	SeriesName string
+
+	// Precision of points timestamps.
+	TimePrecision TimePrecision
+
+	// The labels used by this series.
+	LabelKeys []string
```
vishh: Is this for optimization? Why can't we use all the labels defined as part of the point?

sinks/influxdb/driver.go
```diff
@@ -37,72 +37,130 @@ type influxdbSink struct {
...
+	return influxdb.Second
+}
+
+func (sink *influxdbSink) timeSeriesToInfluxdbSeries(timeseries *sink_api.Timeseries) *influxdb.Series {
+	seriesName := timeseries.SeriesName
+
+	columns := []string{}
+	columns = append(columns, "time")            // Column 0 header
+	columns = append(columns, "value")           // Column 1 header
+	columns = append(columns, "sequence_number") // Column 2 header
```
vishh: How about sharing setting up of these standard columns between the two methods here?

sinks/api/eventdecoder.go
```diff
@@ -0,0 +1,85 @@
...
+	if data.Events != nil && len(data.Events) > 0 {
+		points := []Point{}
+		for _, event := range data.Events {
+			value, err := getEventValue(&event)
+			if err != nil {
+				return result, err
+			}
+			points = append(points, Point{
+				Labels: getEventLabels(&event),
+				Start:  event.LastTimestamp.Time,
```
vishh: Should we round the timestamp here to a millisecond?

sinks/influxdb/driver.go
saad-ali: It is required for filtering out other precisions. Each Timeseries can specify it's own precision and the influxDB client expects all points in a WriteSeriesWithTimePrecision call to have the same precision. So we filter the timeseries, and issue one WriteSeriesWithTimePrecision call per precision type (if that type is present).

sinks/api/types.go
saad-ali: Fixed.

sinks/gcm/driver.go
saad-ali: Changed.

sinks/gcm/driver.go
saad-ali: Changed.

sinks/influxdb/driver.go
saad-ali: See above comment about precision, essentially `storeTimeseriesWithPrecision` is called for each possible precision type (second, milli, micro). It filters any timeseries that match that precision, and issues a `WriteSeriesWithTimePrecision` for those points with that precision. So this log message is essentially saying that a time series with a given precision was written.

sinks/api/metricdecoder.go
```diff
@@ -121,24 +123,36 @@ func (self *defaultDecoder) getContainerMetrics(container *source_api.Container,
 				} else {
 					startTime = stat.Timestamp
 				}
-				points := supported.GetValue(&container.Spec, stat)
-				for _, point := range points {
+				seriesName := supported.Name
+				if supported.Units.String() != "" {
+					seriesName = fmt.Sprintf("%s_%s", seriesName, supported.Units.String())
```
vishh: I don't think GCM will like diverging from the registered series names. What ever series name we used while registering the events is what we should use here.
If Influxdb decides to mangle the series name with additional metadata, then it must be local to Influxdb.

vishh:
@saad-ali: The PR looks good. Just a few comments. Some of the comments are related to sink functionalities and unfortunately they are not documented properly :(

sinks/api/types.go
saad-ali: Yes, it is an optimization. Each `timeseries` can have multiple points, and any given point may specify some subset of labels for the timeseries. So, instead of iterating through all points and their lables to figure out the complete set of label keys, we have `LabelKeys` defined at the top defining all the labels that points in this `timeseries` may have (if any individual point defines a label that is not part of `labelKeys`, it is ignored, if any individual point does not provide a value for a label that is part of `labelKeys`, it is left empty).

sinks/influxdb/driver.go
vishh: Hmm. Maybe rephrase to `Successfully wrote timeseries to InfluxDB with precision: %q`?

sinks/influxdb/driver.go
saad-ali: Done.

sinks/influxdb/driver.go
vishh: Ah! Why do we care about precision at an API level? Why not pass timestamps with max precision and round to a sane precision in the sinks?

sinks/gcm/driver.go
vishh: I found the same pattern in a few other files. Can you make the same change in other files as well?

sinks/api/eventdecoder.go
saad-ali: Done.

sinks/api/metricdecoder.go
```diff
@@ -121,24 +123,36 @@ func (self *defaultDecoder) getContainerMetrics(container *source_api.Container,
...
+				if supported.Type.String() != "" {
+					seriesName = fmt.Sprintf("%s_%s", seriesName, supported.Type.String())
+				}
+				internalPoints := supported.GetValue(&container.Spec, stat)
+				points := []Point{}
+				for _, internalPoint := range internalPoints {
 					labels := copyLabels(labels)
-					for name, value := range point.labels {
+					for name, value := range internalPoint.labels {
 						labels[name] = value
```
vishh: These labels might not show up as part of LabelKeys. Why not include all metric labels by default as part of labelkeys?

saad-ali:
Closing this. #234 replaces this.

PR #209, opened 2015-03-31
Update kube version to test against to 0.14.0

vishh:
Also updated the integration test to validate that all pods and nodes in the cluster be present in Influxdb


vmarmol:
LGTM, wait on green.

vishh:
@vmarmol: This is safe to be merged now.

vmarmol:
:D

PR #210, opened 2015-03-31
Revert "Use node legacy host ip on older kubernetes versions"

vishh:
This commit makes heapster use the external ip in GCE.

This reverts commit aa206c499f187984db9761080df7aa01cb0b5627.

### DIFF
--- a/sources/nodes/kube.go
+++ b/sources/nodes/kube.go
@@ -80,8 +80,6 @@ func (self *kubeNodes) List() (*NodeList, error) {
 				nodeInfo.PublicIP = addr.Address
 			case api.NodeInternalIP:
 				nodeInfo.InternalIP = addr.Address
-			case api.NodeLegacyHostIP:
-				nodeInfo.InternalIP = addr.Address
 			case api.NodeHostName:
 				hostname = addr.Address
 			}


vishh:
Ping @vmarmol 

cc @jimmidyson

jimmidyson:
@vishh - so do newer versions of Kubernetes put the external IP in the `NodeLegacyHostIP` address? Otherwise I don't see how that could happen? Perhaps it's because heapster's uing v1beta1 API & the conversion between v1beta3 & v1beta1 in Kubernetes has an issue with that? Switching heapster to version v1beta3 is probably the better solution, but of course more work.

vishh:
ooo@jimmidyson: From the API it is not clear if `NodeLegacyHostIP` refers to the internal or the external IP. We have been punting on using versioned API inside heapster for quite a while. If this is critical, then it might be worth converting internal API objects from API server over to v1beta3 before consuming inside heapster. On the flipside v1beta3 is still not complete AFAIK.

jimmidyson:
@vishh - agreed v1beta3 is still in flux, as is v1beta2 although less so. I guess I an live without it, but I would like to see heapster supporting multiple versions of kubernetes once it's a bit more stable.

vmarmol:
LGTM, is the Jenkins failure expected?

vishh:
I don't know why jenkins is not updating this PR, but, it did pass on jenkins ([successful run](http://104.154.52.74/job/heapster-e2e-gce/))

vishh:
@vmarmol: This PR is safe to be merged now.

vmarmol:
LGTM

PR #213, opened 2015-04-01
Disable events until a sink is added. 

vishh:
As of now it will burn compute resources.

cc @saad-ali


vishh:
This is useful for 0.10.0 release.

saad-ali:
LGTM. Will merge on green.

vishh:
Ping @saad-ali

PR #216, opened 2015-04-02
Update integration tests to use kube version v0.14.1 [Attempt 2].

vishh:


vmarmol:
LGTM

vmarmol:
@vishh is that build live? It's having trouble pulling it. Maybe lets use 0.14.0 for now?

vishh:
0.14.0 has a bug. Wasted a lot of time on that. 0.14.1 shoudl be fine. Zach
is fixing the release so that a tarball will be available.

vmarmol:
Green :D

PR #217, opened 2015-04-03
Fix gcm heapster config.

vishh:
Updated resolution to 1m and poll duration to 2m for GCM to avoid hitting the defautl quota limits.


vishh:
Ping @rjnagal

rjnagal:
LGTM

vishh:
Thanks for the super quick review @rjnagal!!!

PR #219, opened 2015-04-05
Use https for kubelet requests.

rjnagal:
We have k8 master calls defaulting to http too. But those can be changed in the flag specifying master ip.

### DIFF
--- a/sources/datasource/kubelet.go
+++ b/sources/datasource/kubelet.go
@@ -90,7 +90,7 @@ func (self *kubeletSource) getContainer(url string, start, end time.Time, resolu
 }
 
 func (self *kubeletSource) GetContainer(host Host, start, end time.Time, resolution time.Duration) (container *api.Container, err error) {
-	url := fmt.Sprintf("http://%s:%s/%s", host.IP, host.Port, host.Resource)
+	url := fmt.Sprintf("https://%s:%s/%s", host.IP, host.Port, host.Resource)
 	glog.V(3).Infof("about to query kubelet using url: %q", url)
 
 	return self.getContainer(url, start, end, resolution)


vishh:
@rjnagal: Distributing node certs to heapster will be difficult. So for the short term the idea is to use the read-only 'http' endpoint exposed by the kubelet. We will have to update the default kubelet port.

jimmidyson:
Surely we don't need to distribute the individual nice certificates but just the root CA certificate to verify the nodes address? This should be pretty easy through secrets right?

jimmidyson:
Let's make the kubelet scheme configurable at least?

vishh:
cc @roberthbailey who is working on the new kube security model.

rjnagal:
Do we have any certs available outside secrets today?

sources/datasource/kubelet.go
```diff
@@ -90,7 +90,7 @@ func (self *kubeletSource) getContainer(url string, start, end time.Time, resolu
 }
 
 func (self *kubeletSource) GetContainer(host Host, start, end time.Time, resolution time.Duration) (container *api.Container, err error) {
-	url := fmt.Sprintf("http://%s:%s/%s", host.IP, host.Port, host.Resource)
+	url := fmt.Sprintf("https://%s:%s/%s", host.IP, host.Port, host.Resource)
```
roberthbailey: I don't think this will just work. You need to configure an http.Client that will connect to the kubelet without verifying the server's SSL cert (by setting InsecureSkipVerify to true in the TLSClientConfig).

roberthbailey:
@jimmidyson For Kubernetes, we are working on securing the master <-> node communication (see [#3168](https://github.com/GoogleCloudPlatform/kubernetes/issues/3168)) so we want the kubelet listening on an HTTPS port. We would eventually like to be able to run Kubernetes on a less trusted (or untrusted) network where even exposing a read-only interface in the system infrastructure is considered a bad idea. 

@rjnagal I don't believe that we distribute any certs today (using secrets or otherwise).

rjnagal:
Discussed with @vishh offline.

PR #225, opened 2015-04-08
Add code coverage

jimmidyson:
This PR adds code coverage script that will be run on travis when the build runs. Good to know details like this going forward. We can also use https://coveralls.io by specifying a coveralls token & coveralls will feed back to PRs if configured to.

Also updated gofmt script line to fail build if any files needed go fmt'ing - before this would have just been info but passed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 heapster
+.cover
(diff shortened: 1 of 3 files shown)


vishh:

vishh:
Thanks for the PR! I can set up coveralls for this project!

vishh:
Merging this PR since it does not touch the main code.

PR #228, opened 2015-04-10
Make cumulative metrics gauge metrics in GCM

vmarmol:
Will followup this PR with one that makes the metrics a little more human friendly. No need to make CPU be in billionths of a core for example :)

### DIFF
--- a/sinks/gcm/driver.go
+++ b/sinks/gcm/driver.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/gcloud-golang/compute/metadata"
 	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api"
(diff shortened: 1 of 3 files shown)


rjnagal:
LGTM

sinks/gcm/driver.go
```diff
@@ -78,6 +82,51 @@ type metricDescriptor struct {
...
+			name:        "network/tx",
+			description: "Rate of bytes transmitted over the network in bytes per second",
+		},
+		{
+			name:        "network/tx_errors",
+			description: "Rate of errors transmitting over the network in errors per second",
+		},
+	}
+
+	// Replac the description if we have an alternate one.
```
vishh: s/Replac/Replace

PR #230, opened 2015-04-11
Fix metrics decoder uptime test

saad-ali:
A couple of fixes:
* Expected and actual were flipped.
* Uptime test was calling `time.Since` and comparing to a `time.Since` value generated in the code without fixing the creation or current times. It works most of the time because the delta is usually 0, but is  non-deterministic.


vmarmol:
ok to test

vmarmol:
LGTM

PR #232, opened 2015-04-13
deploy: use secured kubernetes service

simon3z:


vishh:
ok to test

vishh:
I have restarted the integration tests. Not sure why it failed. @simon3z this PR needs a rebase. Thank you!

vishh:
@simon3z: This PR does not seem to work with kube version v0.14.1. When was the secure service added to Kube? 
P.S.: You can run the integration test manually by following [this guide](https://github.com/GoogleCloudPlatform/heapster/tree/master/docs/integration.md)

simon3z:
@vishh idea idea why it fails?

vishh:
@simon3z: I haven't tried running it manually yet. From the logs, it looks like heapster is not able to fetch any data, which happens when there are issues reaching out to the apiserver.

jimmidyson:
AFAIK the default read only api service in vanilla kube doesn't use https by default (although it is in other impls like openshift). I suggest making the scheme configurable via an env var, defaulting to http.

vishh:
This might become unnecessary once heapster starts using secrets to talk to the apiserver securely. FYI!

jimmidyson:
One option I thought of is we could sniff whether kubernetes-ro service is https or not by just trying to connect on https, falling back to http if it fails. I've used this on other projects successfully before.

vishh:
@jimmidyson :+1:

vishh:
#279 should have achieved the same goal. Kindly re-open if #279 doesn't satisfy your use case @simon3z

PR #234, opened 2015-04-14
Heapster sink events to InfluxDB

saad-ali:
This implements the second part of #118 (follow up to PR #192).

Related Kubernetes Issues:

Screenshot showing events in InfluxDB (pulled by heapster from Kubernetes API server and pushed by heapster to influxdb):
influxdbevents07

Branched from #205

CC @vishh

### DIFF
--- a/integration/kube_test.go
+++ b/integration/kube_test.go
@@ -44,6 +44,7 @@ const (
 	makefile             = "../Makefile"
 	podlistQuery         = "select distinct(pod_id) from /cpu.*/"
 	nodelistQuery        = "select distinct(hostname) from /cpu.*/"
+	eventsQuery          = "select distinct(value) from \"log/events\""
 )
 
 var (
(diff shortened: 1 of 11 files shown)


vishh:

integration/kube_test.go
```diff
@@ -328,6 +329,20 @@ func expectedItemsExist(expectedItems []string, actualItems map[string]bool) boo
 	return true
 }
 
+func validateEvents(influxdbClient *influxdb.Client) bool {
+	events, err := getInfluxdbData(influxdbClient, eventsQuery)
```
vishh: Should we check for atleast one event from all pods?

saad-ali: I don't think that is necessary. We just need to verify that events are making it through the pipeline.

saad-ali:
Integration tests pass when run locally:
```
--- PASS: TestHeapsterInfluxDBWorks (508.96s)
PASS
ok  	github.com/GoogleCloudPlatform/heapster/integration	508.996s
```

Looks like all recent Travis runs are failing with
```
The command "go get golang.org/x/tools/cmd/vet" failed and exited with 1 during .
```

saad-ali:
Travis is now green

sinks/influxdb/driver_test.go
```diff
@@ -0,0 +1,189 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
vishh: Nice job with these tests! :100:

saad-ali: :)

vishh:
LGTM.

PR #235, opened 2015-04-14
Update v1beta3 configs.

vishh:
Once a new kube release is out, I will update the e2e to use the v1beta3 configs by default. For now, the reviewer will have to trust that I tested this config manually!


vmarmol:
LGTM

vishh:
Self merging since tests don't use v1beta3 confis yet.

PR #236, opened 2015-04-14
Update grafana dashboard grouping time to match the default resolution of 5s in heapster

vishh:


grafana/RELEASES.md
```diff
@@ -1,5 +1,8 @@
 # Release Notes
 
+## 0.7 (Unreleased)
```
vmarmol: Was this meant to be here? Maybe do in another PR or commit? We should probably mention some of the GCM changes as well.

vishh: Note that it says unreleased. I am just tracking all the changes as we make
them under a new release label.

vishh:
Self merging since tests do not capture grafana dashboards.

PR #237, opened 2015-04-15
Update kubelet port to 10255

vishh:
updated integration tests to use 0.15.0 release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,14 @@ sanitize:
 	hooks/check_gofmt.sh
 	hooks/run_vet.sh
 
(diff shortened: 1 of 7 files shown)


rjnagal:
It's failing with the same error we saw before:
kube_test.go:350] failed to query list of pods from influxdb. Query: "select distinct(pod_id) from /cpu.*/", Err: expected only one series from Influxdb for query "select distinct(pod_id) from /cpu.*/". Got []

vishh:
Ping @rjnagal. This PR fixes heapster head

vmarmol:
LGTM

PR #239, opened 2015-04-16
GCM: Export metrics as floats and change units.

vmarmol:
Makes units more user friendly for some metrics:
- uptime: seconds/second
- CPU usage: millicores/second

### DIFF
--- a/sinks/gcm/driver.go
+++ b/sinks/gcm/driver.go
@@ -94,11 +94,11 @@ func getDescription(metric sink_api.MetricDescriptor) string {
 	}{
 		{
 			name:        "uptime",
-			description: "Rate of change of time since start in milliseconds per second",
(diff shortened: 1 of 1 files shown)


sinks/gcm/driver.go
```diff
@@ -94,11 +94,11 @@ func getDescription(metric sink_api.MetricDescriptor) string {
 	}{
 		{
 			name:        "uptime",
-			description: "Rate of change of time since start in milliseconds per second",
```
rjnagal: curious: why is uptime a rate?

rjnagal:
LGTM

sinks/gcm/driver.go
vmarmol: We just turned all cumulative into rates. I don't think it's easy to use either way... Maybe keeping cumulative is less worse

PR #242, opened 2015-04-17
Updating kube deps to match v0.15 release

vishh:
Had to modify the source for the rebase to work. Converted all yaml configs to json. 
Next step is to move the integration tests to use v1beta3 configs and then get rid of v1beta1 configs. Uff!


vishh:
Ping @rjnagal

rjnagal:
LGTM

PR #247, opened 2015-04-18
Add instantaneous metrics in addition to cumulative metrics in GCM driver

vishh:
@vmarmol: If this change works, then I will update the docs before this PR gets merged.

### DIFF
--- a/docs/storage-schema.md
+++ b/docs/storage-schema.md
@@ -59,6 +59,14 @@ Heapster adds timestamp and sequence number to every metric.
 
 ### Google Cloud Monitoring
 
-Metrics mentioned above are stored along with corresponding labels as custom metrics in Google Cloud Monitoring.
-Each metric is pushed with an additonal namespace prefix `kubernetes.io`.
-TODO: Add more information about GCM support.
+Metrics mentioned above are stored along with corresponding labels as [custom metrics](https://cloud.google.com/monitoring/custom-metrics/) in Google Cloud Monitoring.
+1. Metrics are collected every 2 minutes by default and pushed with a 1 minute precision.
+2. Each metric has a custom metric prefix - `custom.cloudmonitoring.googleapis.com`
+3. Each metric is pushed with an additonal namespace prefix - `kubernetes.io`.
+4. GCM does not support visualizing cumulative metrics yet. To work around that, heapster exports an equivalent gauge metric for all cumulative metrics mentioned above.
+   The gauge metrics use their parent cumulative metric name as the prefix, followed by a "_rate" suffix. 
+   E.x.: "cpu/usage", which is cumulative, will have a corresponding gauge metric "cpu/usage_rate"
+   NOTE: The gauge metrics will be deprecated as soon as GCM supports visualizing cumulative metrics.
+
+TODO: Add a snapshot of all the metrics stored in GCM.
+
--- a/sinks/gcm/driver.go
+++ b/sinks/gcm/driver.go
@@ -18,8 +18,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -43,7 +45,7 @@ type gcmSink struct {
 
 	// TODO(vmarmol): Also store labels?
 	// Map of metrics we currently export.
-	exportedMetrics map[string]sink_api.MetricDescriptor
+	exportedMetrics map[string]metricDescriptor
 
 	// The last value we have pushed for every cumulative metric.
 	lastValue gcstore.GCStore
@@ -81,8 +83,6 @@ type metricDescriptor struct {
 	TypeDescriptor typeDescriptor             `json:"typeDescriptor,omitempty"`
 }
 
-const maxNumLabels = 10
-
 // Map of metric name to translation function.
 var translationFuncs = map[string]func(float64) float64{
 	"uptime_rate": func(value float64) float64 {
@@ -95,16 +95,67 @@ var translationFuncs = map[string]func(float64) float64{
 	},
 }
 
-func (self *gcmSink) addMetric(metric sink_api.MetricDescriptor, request metricDescriptor) error {
-	err := sendRequest(fmt.Sprintf("https://www.googleapis.com/cloudmonitoring/v2beta2/projects/%s/metricDescriptors", self.project), self.token, request)
-	glog.Infof("[GCM] Adding metric %q: %v", metric.Name, err)
+type listMetricsResponse struct {
+	Metrics []metricDescriptor `json:"metrics,omitempty"`
+}
+
+func (self *gcmSink) defaultUrlPath() string {
+	return fmt.Sprintf("https://www.googleapis.com/cloudmonitoring/v2beta2/projects/%s", self.project)
+}
+
+func (self *gcmSink) listMetrics() error {
+	response := listMetricsResponse{}
+	url, err := url.Parse(self.defaultUrlPath() + "/metricDescriptors")
 	if err != nil {
 		return err
 	}
+	err = sendRequest("GET", self.token, url, nil, &response)
+	if err != nil {
+		glog.Errorf("[GCM] list metrics failed %v", err)
+		return err
+	}
+	apiPrefix := fmt.Sprintf("%s/%s", customApiPrefix, metricDomain)
+	for idx, m := range response.Metrics {
+		if strings.HasPrefix(m.Name, apiPrefix) {
+			self.exportedMetrics[m.Name] = response.Metrics[idx]
+		}
+	}
 
-	// Add metric to exportedMetrics.
-	self.exportedMetrics[metric.Name] = metric
 	return nil
+}
+
+func (self *gcmSink) deleteMetric(metricName string) error {
+	url := &url.URL{
+		Scheme: "https",
+		Host:   "www.googleapis.com",
+		Opaque: fmt.Sprintf("%s/metricDescriptors/%s", self.defaultUrlPath(), url.QueryEscape(metricName)),
+	}
+
+	err := sendRequest("DELETE", self.token, url, nil, nil)
+	if err != nil {
+		glog.V(2).Infof("[GCM] Deleting metric %q failed: %v", metricName, err)
+	}
+	return err
+}
+
+func (self *gcmSink) addMetric(request metricDescriptor) error {
+	if existingMetric, found := self.exportedMetrics[request.Name]; found {
+		if existingMetric.TypeDescriptor != request.TypeDescriptor {
+			if err := self.deleteMetric(request.Name); err != nil {
+				return err
+			}
+		}
+	}
+	url, err := url.Parse(self.defaultUrlPath() + "/metricDescriptors")
+	if err != nil {
+		return err
+	}
+	if err = sendRequest("POST", self.token, url, request, nil); err == nil {
+		glog.V(3).Infof("[GCM] Added metric %q", request.Name)
+		// Add metric to exportedMetrics.
+		self.exportedMetrics[request.Name] = request
+	}
+	return err
 }
 
 // Adds the specified metrics or updates them if they already exist.
@@ -130,18 +181,16 @@ func (self *gcmSink) Register(metrics []sink_api.MetricDescriptor) error {
 				ValueType:  metric.ValueType.String(),
 			},
 		}
-		if err := self.addMetric(metric, request); err != nil {
+		if err := self.addMetric(request); err != nil {
 			return err
 		}
 
-		if rateMetric, ok := getRateMetric(metric.Name); ok {
+		if rateMetric, exists := gcmRateMetrics[metric.Name]; exists {
 			request.Name = fullMetricName(rateMetric.name)
 			request.Description = rateMetric.description
 			request.TypeDescriptor.MetricType = sink_api.MetricGauge.String()
-			originalMetric := metric
-			originalMetric.Name = rateMetric.name
-			originalMetric.Description = rateMetric.description
-			if err := self.addMetric(originalMetric, request); err != nil {
+			request.TypeDescriptor.ValueType = sink_api.ValueDouble.String()
+			if err := self.addMetric(request); err != nil {
 				return err
 			}
 		}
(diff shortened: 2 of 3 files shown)


vishh:
@EranGabber: This PR will ensure the existing cumulative metrics are preserved.

sinks/gcm/driver.go
```diff
@@ -191,6 +161,7 @@ type timeseriesDescriptor struct {
 type point struct {
 	Start       time.Time `json:"start,omitempty"`
 	End         time.Time `json:"end,omitempty"`
+	Int64Value  int64     `json:"int64Value"`
```
vmarmol: Sadly this won't work for now. Since it is not omitempty we'd generate both types which will confuse the GCM API. We can't make them omitempty since 0 is a valid value. We need to have two structs we define for this (probably a base embedded into an int or double type). Easiest thing may be to just use double for all the metrics for now. It should not reduce the accuracy of our numbers I believe.

vmarmol:
Outside of that problem, it looks good!

sinks/gcm/driver.go
```diff
@@ -83,67 +85,71 @@ type metricDescriptor struct {
...
+	for idx, m := range response.Metrics {
+		if strings.Contains(m.Name, customApiPrefix) && strings.Contains(m.Name, metricDomain) {
+			self.exportedMetrics[m.Name] = response.Metrics[idx]
+		}
+
+	}
+	return nil
+}
+
+func (self *gcmSink) deleteMetric(metricName string) error {
```
vishh: @vmarmol: This never seems to work. Deletion works via the API explorer. Can you try this code once and see if it works for you?

sinks/gcm/driver.go
vishh: Ok. Updated all metric to be doubles. The only downside is that now all the existing metrics are invalid. To solve this, I have added code to query existing metrics, check if they differ from requirements and delete if they do, and re-add the metric.

vishh:
@vmarmol: Give this branch a spin if possible. I am not able to figure out why the DELETE call fails :(

EranGabber:
Thank you for fixing it.

sinks/api/supported_metrics.go
```diff
@@ -84,12 +84,12 @@ var statMetrics = []SupportedStatMetric{
 			return []internalPoint{{value: int64(stat.Memory.WorkingSet)}}
 		},
 	},
-	{
+	/*	{
```
vmarmol: commented out?

sinks/gcm/driver.go
vmarmol: You have to send the fullMetricName() I believe

sinks/gcm/driver.go
```diff
@@ -83,67 +85,71 @@ type metricDescriptor struct {
...
 	},
 }
 
+type listMetricsResponse struct {
+	Metrics []metricDescriptor `json:"metrics,omitempty"`
+}
+
+func (self *gcmSink) listMetrics() error {
+	response := listMetricsResponse{}
+	err := sendRequest("GET", fmt.Sprintf("https://www.googleapis.com/cloudmonitoring/v2beta2/projects/%s/metricDescriptors", self.project), self.token, nil, &response)
```
vmarmol: I think we can move the URL to a separate function (or a const) since we use it in multiple places

sinks/gcm/driver.go
```diff
@@ -83,67 +85,71 @@ type metricDescriptor struct {
...
+	if err != nil {
+		glog.Errorf("[GCM} list metrics failed %v", err)
```
vmarmol: nit: `[GCM]` second is a bracket right now

sinks/gcm/driver.go
```diff
@@ -366,27 +395,23 @@ func fullLabelName(name string) string {
...
-	// This will ease the transition to cumulative metrics when those come.
-	suffix := ""
-	if metricType == sink_api.MetricCumulative {
-		suffix = "_delta"
-	}
-
-	if !strings.Contains(name, "custom.cloudmonitoring.googleapis.com/") {
-		return fmt.Sprintf("custom.cloudmonitoring.googleapis.com/%s/%s%s", metricDomain, name, suffix)
+func fullMetricName(name string) string {
+	if !strings.Contains(name, customApiPrefix) {
```
vmarmol: Check prefix instead of contains?

sinks/gcm/rate_metrics.go
```diff
@@ -0,0 +1,57 @@
...
+		name:        "network/tx_rate",
+		description: "Rate of bytes transmitted over the network in bytes per second",
+	},
+	"network/tx_errors": rateMetric{
+		name:        "network/tx_errors_rate",
+		description: "Rate of errors transmitting over the network in errors per second",
+	},
+}
+
+func getRateMetric(baseMetric string) (rateMetric, bool) {
```
vmarmol: Why the function redirect? Fine either way :)

sinks/gcm/driver.go
```diff
@@ -83,67 +85,71 @@ type metricDescriptor struct {
...
+
+func (self *gcmSink) listMetrics() error {
+	response := listMetricsResponse{}
+	err := sendRequest("GET", fmt.Sprintf("https://www.googleapis.com/cloudmonitoring/v2beta2/projects/%s/metricDescriptors", self.project), self.token, nil, &response)
+	if err != nil {
+		glog.Errorf("[GCM} list metrics failed %v", err)
+		return err
+	}
+	for idx, m := range response.Metrics {
+		if strings.Contains(m.Name, customApiPrefix) && strings.Contains(m.Name, metricDomain) {
```
vmarmol: We should check the metric domain as well. Also: prefix instead of contains?

sinks/gcm/driver.go
vmarmol: Shouldn't we not need to do this btw? We should know if the metrics coming in are ours or not no?

sinks/gcm/driver.go
vmarmol: That looks like it is still not sufficient. It is interesting that we're getting a 404.

sinks/api/supported_metrics.go
vishh: WIP :) Fixed it.

sinks/gcm/driver.go
vishh: I am relying on the `kubernetes.io` namespace to isolate just our metrics. The prefix check makes sense.

sinks/gcm/driver.go
vishh: Done

sinks/gcm/rate_metrics.go
vishh: Deleted it

sinks/gcm/driver.go
vishh: Done

sinks/gcm/driver.go
vishh: Done

vishh:
PTAL @vmarmol. I updated the docs as well. I tested manually and it works. An integration test for GCM would reduce that anxiety :)

docs/storage-schema.md
```diff
@@ -59,6 +59,14 @@ Heapster adds timestamp and sequence number to every metric.
 
 ### Google Cloud Monitoring
 
-Metrics mentioned above are stored along with corresponding labels as custom metrics in Google Cloud Monitoring.
-Each metric is pushed with an additonal namespace prefix `kubernetes.io`.
-TODO: Add more information about GCM support.
+Metrics mentioned above are stored along with corresponding labels as [custom metrics](https://cloud.google.com/monitoring/custom-metrics/) in Google Cloud Monitoring.
+1. Metrics are collected every 2 minutes by default and pushed with a 1 minute precision.
+2. Each metrics has a custom metric prefix - `custom.cloudmonitoring.googleapis.com`
```
vmarmol: nit: each "metric"

docs/storage-schema.md
```diff
@@ -59,6 +59,14 @@ Heapster adds timestamp and sequence number to every metric.
...
+3. Each metric is pushed with an additonal namespace prefix - `kubernetes.io`.
+4. GCM does not support visualizing cumulative metrics yet. To work around that, heapster exports an equivalent gauge metric for all cumulative metrics mentioned above.
+   The gauge metrics use their parent cumulative metric name as the prefix, followed by a "_rate" suffix. 
+   E.x.: "cpu/usage", which is cumulative, will have a corresponding gauge metric "cpu/usage_rate"
+   NOTE: The gauge metrics will be deprecated as soon as GCM supports visualizing cumulative metrics.
+
+TODO: Add a detailed user guide for GCM.
```
vmarmol: What are we missing?

sinks/gcm/driver.go
```diff
@@ -81,69 +83,85 @@ type metricDescriptor struct {
...
+	if err != nil {
+		return err
+	}
+	err = sendRequest("GET", self.token, url, nil, &response)
+	if err != nil {
+		glog.Errorf("[GCM] list metrics failed %v", err)
+		return err
+	}
+	for idx, m := range response.Metrics {
+		apiPrefix := fmt.Sprintf("%s/%s", customApiPrefix, metricDomain)
```
vmarmol: can build outside the for loop?

sinks/gcm/driver.go
```diff
@@ -81,69 +83,85 @@ type metricDescriptor struct {
...
+		if strings.HasPrefix(m.Name, apiPrefix) {
+			self.exportedMetrics[m.Name] = response.Metrics[idx]
+		}
+	}
+
+	return nil
+}
+
+func (self *gcmSink) deleteMetric(metricName string) error {
+	url := &url.URL{
```
vmarmol: lol :)

sinks/gcm/driver.go
```diff
@@ -191,7 +214,8 @@ type timeseriesDescriptor struct {
 type point struct {
 	Start       time.Time `json:"start,omitempty"`
 	End         time.Time `json:"end,omitempty"`
-	DoubleValue float64   `json:"doubleValue"`
+	DoubleValue *float64  `json:"doubleValue,omitempty"`
```
vmarmol: we can't do omitempty unfortunately since zero is a valid value :( We need two different structs one for each type.

docs/storage-schema.md
vishh: How to access and graph custom metrics; A snapshot of all the metrics stored in gcm would help.

sinks/gcm/driver.go
vishh: The pointer should do the trick ;)

sinks/gcm/driver.go
vishh: Done

docs/storage-schema.md
vishh: Done

vishh:
Addressed comments

docs/storage-schema.md
vmarmol: The first one is in the GCM started guide no? and the second is this document?

sinks/gcm/driver.go
vmarmol: Ah, sneaky, I like it :)

sinks/gcm/driver.go
```diff
@@ -366,46 +417,51 @@ func fullLabelName(name string) string {
...
+		glog.Errorf("request creation failed: method: %q, url: %q, error: %v", method, url, err)
 		return err
 	}
+	req.URL = url
 	req.Header.Add("Content-Type", "application/json")
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
 
 	resp, err := (&http.Client{}).Do(req)
 	if err != nil {
+		glog.Errorf("req %+v failed with error: %v", req, err)
```
vmarmol: nit: remove

sinks/gcm/driver.go
```diff
@@ -366,46 +417,51 @@ func fullLabelName(name string) string {
...
+		if err != nil {
+			return err
+		}
+		rawRequest = bytes.NewReader(jsonRequest)
 	}
-
-	req, err := http.NewRequest("POST", url, bytes.NewReader(rawRequest))
+	req, err := http.NewRequest(method, url.String(), rawRequest)
 	if err != nil {
+		glog.Errorf("request creation failed: method: %q, url: %q, error: %v", method, url, err)
```
vmarmol: nit: remove

sinks/gcm/driver.go
```diff
@@ -81,69 +83,85 @@ type metricDescriptor struct {
...
+	}
+	return err
+}
+
+func (self *gcmSink) addMetric(request metricDescriptor) error {
+	if err := self.listMetrics(); err != nil {
+		return err
+	}
+	if existingMetric, found := self.exportedMetrics[request.Name]; found {
+		if existingMetric.TypeDescriptor != request.TypeDescriptor ||
```
vmarmol: You can check for only type, if the description is different the add below would update it.

sinks/gcm/driver.go
```diff
@@ -81,69 +83,85 @@ type metricDescriptor struct {
...
+
+	err := sendRequest("DELETE", self.token, url, nil, nil)
+	if err != nil {
+		glog.V(2).Infof("[GCM] Deleting metric %q failed: %v", metricName, err)
+	}
+	return err
+}
+
+func (self *gcmSink) addMetric(request metricDescriptor) error {
+	if err := self.listMetrics(); err != nil {
```
vmarmol: We should only need to do this once in register no? Since it builds the map of supported metrics.

docs/storage-schema.md
vishh: Agreed!

vmarmol: I guess I mean that we have both already :)

sinks/gcm/driver.go
vishh: Good point. I meant to do it and somehow forgot to do it.

sinks/gcm/driver.go
vishh: Ah! I did not know that. updated!

sinks/gcm/driver.go
vishh: Done

sinks/gcm/driver.go
vishh: Done

vishh:
Thanks for the review @vmarmol !

docs/storage-schema.md
vishh: I would prefer posting a full list just to make it easy. May be just list all the rate metrics that are specific to GCM.

vmarmol:
LGTM, thanks @vishh!

PR #255, opened 2015-04-23
Avoid duplicate kube events in InfluxDB

### LINKED ISSUE #250: Avoid duplicate kube events in InfluxDB
We can use the UID of kube events as the sequence number while writing to influxdb. That combined with the timestamp in the event will help avoid duplicates in Influxdb.

saad-ali:
Avoid duplicate kube events in InfluxDB by deterministically setting `sequence_number` value for each event.

Closes #250


vishh:
LGTM

PR #258, opened 2015-04-24
Fix default kubeletPort in docs

luksa:


vishh:
Thanks for the fix @luksa! 

PR #261, opened 2015-04-25
Configure multiple sinks

vishh:
Docs will come soon, either in this PR or in a later PR.

cc @vmarmol @saad-ali @jimmidyson

### DIFF
--- a/deploy/kube-config/influxdb/v1beta3/heapster-controller.json
+++ b/deploy/kube-config/influxdb/v1beta3/heapster-controller.json
@@ -28,7 +28,7 @@
             "env": [
               {
                 "name": "INFLUXDB_HOST",
-                "value": "monitoring-influxdb"
+                "value": "http://monitoring-influxdb:80"
               },
               {
                 "name": "SINK",
--- a/sinks/influxdb/driver.go
+++ b/sinks/influxdb/driver.go
@@ -18,10 +18,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"hash/fnv"
+	"net/url"
+	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/GoogleCloudPlatform/heapster/extpoints"
 	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api"
 	kube_api "github.com/GoogleCloudPlatform/kubernetes/pkg/api"
 	"github.com/golang/glog"
@@ -36,19 +40,28 @@ type influxDBClient interface {
 }
 
 type influxdbSink struct {
-	hostname  string
-	database  string
 	client    influxDBClient
-	dbName    string
 	stateLock sync.RWMutex
 	// TODO(rjnagal): switch to atomic if writeFailures is the only protected data.
 	writeFailures int // guarded by stateLock
-	avoidColumns  bool
 	seqNum        metricSequenceNum
+	c             config
+}
+
+type config struct {
+	user         string
+	password     string
+	host         string
+	dbName       string
+	avoidColumns bool
 }
 
 const (
-	EventsSeriesName = "log/events"
+	eventsSeriesName = "log/events"
+	// Attempt database creation maxRetries times before quitting.
+	maxRetries = 20
+	// Sleep for waitDuration between database creation retries.
+	waitDuration = 30 * time.Second
 )
 
 func (self *influxdbSink) Register(metrics []sink_api.MetricDescriptor) error {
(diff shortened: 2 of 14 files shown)


deploy/kube-config/influxdb/v1beta3/heapster-controller.json
```diff
@@ -28,7 +28,7 @@
             "env": [
               {
                 "name": "INFLUXDB_HOST",
-                "value": "monitoring-influxdb"
+                "value": "http://monitoring-influxdb"
```
jimmidyson: You'll need to add port 8086 in too.

sinks/factory.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, err
+	}
+	glog.Infof("creating influxdb sink with options: %v", defaultConfig)
+
+	return []sink_api.ExternalSink{sink}, nil
+}
+
+func CreateGCMSink(_ string, _ map[string][]string) ([]sink_api.ExternalSink, error) {
+	sink, err := gcm.New()
```
jimmidyson: Is there any way we could use the options map instead of the flags to keep it consistent? How about using the first param as the client id?

sinks/factory.go
```diff
@@ -0,0 +1,89 @@
...
+	glog.Infof("creating GCM sink")
+	return []sink_api.ExternalSink{sink}, err
+}
+
+func CreateGCLSink(_ string, _ map[string][]string) ([]sink_api.ExternalSink, error) {
+	sink, err := gcl.New()
```
jimmidyson: Same for GCM sink.

sinks/influxdb/driver.go
```diff
@@ -281,31 +290,23 @@ func createDatabase(databaseName string, client *influxdb.Client) error {
...
-		Username: username,
-		Password: password,
-		Database: databaseName,
-		IsSecure: false,
+	iConfig := &influxdb.ClientConfig{
+		Host:     c.Host,
+		Username: c.User,
+		Password: c.Password,
+		Database: c.DbName,
+		IsSecure: c.AvoidColumns,
```
jimmidyson: `IsSecure` = `c.AvoidColumns`? Typo?

sinks/factory.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/GoogleCloudPlatform/heapster/extpoints"
+	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api"
+	"github.com/GoogleCloudPlatform/heapster/sinks/gcl"
+	"github.com/GoogleCloudPlatform/heapster/sinks/gcm"
+	"github.com/GoogleCloudPlatform/heapster/sinks/influxdb"
+	"github.com/golang/glog"
+)
+
+func init() {
+	extpoints.SinkFactories.Register(CreateInfluxdbSink, "influxdb")
```
jimmidyson: Could these registrations be moved into their respective packages to keep consistent registration pattern?

sinks/factory.go
```diff
@@ -0,0 +1,89 @@
...
+	extpoints.SinkFactories.Register(CreateGCMSink, "gcm")
+	extpoints.SinkFactories.Register(CreateGCLSink, "gcl")
+}
+
+func CreateInfluxdbSink(uri string, options map[string][]string) ([]sink_api.ExternalSink, error) {
```
jimmidyson: Same with the factory methods - can these be moved into their respective patterns to keep things cleaner?

sinks/factory.go
vishh: By flags are you referring to the opaque part of the url? These drivers will require more options once they are enhanced to work outside of GCE as well.

sinks/factory.go
vishh: Can you elaborate a bit more?

deploy/kube-config/influxdb/v1beta3/heapster-controller.json
vishh: Done. Its actually port '80' since its a kubernetes service.

sinks/influxdb/driver.go
vishh: Good catch! Done

sinks/factory.go
vishh: Done

sinks/factory.go
vishh: Done

vishh:
Ping @vmarmol @rjnagal

rjnagal:
LGTM

PR #262, opened 2015-04-25
Update extension of kube config files from '.js' to '.json'.

vishh:
cc @vmarmol


vmarmol:
LGTM

vishh:
Self merging since it has been reviewed already!

PR #264, opened 2015-04-25
Change the logging scope from 'admin' to 'write' for GCL driver.

vishh:
Updated the log file name to 'kubernetes.io/events'


vishh:
I tested this manually with just write scope enabled for logging and it works.

vmarmol:
LGTM

PR #270, opened 2015-05-01
Remove v1beta1 configs and make v1beta3 configs the default.

vishh:


vmarmol:
LGTM

vishh:
Self merging this PR with the LGTM from @vmarmol

PR #272, opened 2015-05-04
Adding support for Bosun

vishh:
Bosun will let users configure and generate alerts.
Bosun can store metrics in InfluxDB. 
The Sink now stores metrics only.

Next steps are to cook up a kube script for deploying Bosun alongside influxdb and post a guide for setting up alerts!

screen shot 2015-05-03 at 7 53 29 pm


piosz:
@vishh do you still want to get this in? If yes please re-open and rebase.

Issue #269, opened 2015-05-01
Update heapster to use kubeconfig credentials on kubernetes


erictune:
In the file where you currently create a client using kubernetes_auth file, do something like this:

```go
import (	kclientcmd "github.com/GoogleCloudPlatform/kubernetes/pkg/client/clientcmd" )
...

kubecfg_file          = flag.String("kubecfg_file", "", "Location of kubecfg file for access to kubernetes service")
...

func newKubeClient() (*client.Client, error) {
  var config *client.Config
  if *kubecfg_file != "" {
    var err error
    if config, err = kclientcmd.NewNonInteractiveDeferredLoadingClientConfig(
      &kclientcmd.ClientConfigLoadingRules{ExplicitPath: *kubecfg_file},
      &kclientcmd.ConfigOverrides{}).ClientConfig(); err != nil {
        return nil, err
    }
  } else {
    // however you handle this case
  }
  return client.New(config), nil
}
```

And in your pod template add:
```yaml
volumeMounts:
- name: monitoring-token
  mountPath: /etc/kubernetes/kubeconfig
  readOnly: true
```

and in the heapster container of the pod template add:
```yaml
volumes:
 - name: dns-token
   source:
    secret:
     target:
       kind: Secret
       namespace: default
       name: token-system-monitoring
```

And add this argument to heapster `--kubeconfig=/etc/kubernetes/kubeconfig`

erictune:
@vishh asked for this issue

PR #276, opened 2015-05-05
Added support for connecting to kubelets through https (using the same certificates as for kubernetes master)

luksa:


vishh:
ok to test

sources/kube_factory.go
```diff
@@ -94,10 +95,28 @@ func CreateKubeSources(uri string, options map[string][]string) ([]api.Source, e
...
+	}
 	glog.Infof("Using Kubernetes client with master %q and version %s\n", kubeConfig.Host, kubeConfig.Version)
 	glog.Infof("Using kubelet port %d", kubeletPort)
-	kubeletApi := datasource.NewKubelet()
-	kubePodsSource := NewKubePodMetrics(kubeletPort, nodesApi, newPodsApi(kubeClient), kubeletApi)
+
+	kubeletConfig := &kube_client.KubeletConfig{
+		Port:            uint(kubeletPort),
+		EnableHttps:     kubeletHttps,
+		TLSClientConfig: kubeConfig.TLSClientConfig,
```
vishh: I assume #269 is required for this PR to just work?

vishh:
Thanks for the PR @luksa! I wonder if adding access to kubeConfig is a prerequisite for this PR?

sources/kube_factory.go
luksa: No, I wasn't even aware of #269. 

I'm runnin heapster like this: heapster --source="kubernetes:https://ip:port?auth=clientauth&kubeletPort=10250", where clientauth is a json file that defines paths to the cert files (docs on this are here: https://github.com/GoogleCloudPlatform/heapster/blob/master/docs/source-configuration.md)

vishh: #279 takes care of kube auth plumbing. Once that goes in, can you rebase this PR on HEAD and then this will be good to go!

vishh: #279 is merged.

vishh:
Ping @luksa

luksa:
I've force-pushed the rebased commits.

vishh:
LGTM.

sources/kube_factory.go
glanzi: @luksa: how does the clientauth json file exactly look. i need to define ca, cert and key for accessing my protected kubelets on port 10250 over https (including --anonymous-auth=false). thanks a lot for your answer. the doc say nothing about the format of the json..

PR #278, opened 2015-05-05
Adding secrets as a volume.

vishh:
A PR to use the auth files will show up soon. 

Also removed kube auth version flag.


vmarmol:
LGTM

PR #284, opened 2015-05-07
Support for Hawkular-Metrics sink.

burmanm:
This PR provides support for pushing numerical metrics to the Hawkular-Metrics (http://www.hawkular.org/) for monitoring purposes. The Hawkular-Metrics endpoints can be used standalone or with a full Hawkular 

github.com/hawkular/hawkular-client-go/metrics is a required dependency.

### DIFF
--- a/sinks/hawkular/driver.go
+++ b/sinks/hawkular/driver.go
@@ -1,0 +1,284 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hawkular
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+
+	"github.com/GoogleCloudPlatform/heapster/extpoints"
+	"github.com/golang/glog"
+	"github.com/hawkular/hawkular-client-go/metrics"
+
+	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api"
+	kube_api "github.com/GoogleCloudPlatform/kubernetes/pkg/api"
+)
+
+const (
+	unitsTag       string = "units"
+	typeTag        string = "type"
+	descriptionTag string = "_description"
+	descriptorTag  string = "descriptor_name"
+	groupTag       string = "group_id"
+	separator      string = "/"
+)
+
+type hawkularSink struct {
+	client  *metrics.Client
+	models  map[string]metrics.MetricDefinition // Model definitions
+	regLock sync.Mutex
+	reg     map[string]*metrics.MetricDefinition // Real definitions
+}
+
+// START: ExternalSink interface implementations
+
+func (self *hawkularSink) Register(mds []sink_api.MetricDescriptor) error {
+	self.regLock.Lock()
+	defer self.regLock.Unlock()
+
+	// Create model definitions based on the MetricDescriptors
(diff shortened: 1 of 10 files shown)


jimmidyson:
Could you please add the required godeps so builds are repeatable?

vishh:
Thanks for the PR @burmanm. Can you add all new go dependencies using `godep save`?

sinks/hawkular/driver.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/GoogleCloudPlatform/heapster/extpoints"
+	"github.com/golang/glog"
+	"github.com/hawkular/hawkular-client-go/metrics"
+
+	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api"
+	kube_api "github.com/GoogleCloudPlatform/kubernetes/pkg/api"
+)
+
+const (
+	units_tag string = "units"
```
vishh: nit: Take a look at the go style guide [here](https://code.google.com/p/go-wiki/wiki/CodeReviewComments#Mixed_Caps) and update the code.

sinks/hawkular/driver.go
```diff
@@ -0,0 +1,236 @@
...
+			}
+
+			// Add to the lookup table
+			self.reg[key] = &md
+			glog.Infof("Registered new metric definition: %s", key)
+		} else {
+			return fmt.Errorf("Could not find definition model with name %s", t.MetricDescriptor.Name)
+		}
+	}
+	// TODO Compare the definition tags and update if necessary? Quite expensive operation..
```
vishh: Since this is a one time operation, I think we can disregard the expense. WDYT? Have updates would make user experience nicer. Very soon, we hope to lock down the current storage schema with appropriate versioning.

sinks/hawkular/driver.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	// Fetch currently known metrics from Hawkular-Metrics and cache them
+	prev, err := self.client.Definitions(metrics.Gauge)
+	if err != nil {
+		return err
+	}
+
+	for _, p := range prev {
+		self.reg[p.Id] = p
```
vishh: Why are you not performing the actual registration here?

sinks/hawkular/driver.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	m := metrics.Datapoint{
+		Value:     value,
+		Timestamp: metrics.UnixMilli(p.End),
+	}
+
+	// At the moment all the values are converted to gauges.
```
vishh: Why?

sinks/hawkular/driver.go
burmanm: Actually, yes, this would be doable in the register phase (here it isn't as we run this process on every occasion of metric store - I'll explain in other comment why) -  compare the metricDescriptor labels with the stuff we've already stored and mark stuff to be updated. Versioning would make this a lot simpler process as we could store the version on the Hawkular side.

sinks/hawkular/driver.go
burmanm: In this model, we would store the metric with a name: containerName/containerId/metricName and only the last one is available on the Register() phase. If the register would be called when new container is detected, then it would make sense to register here. But as we now want to create a metric per each container+metric combination it can't be done here, so that's why we store the model only here.

sinks/hawkular/driver.go
burmanm: Our gauge support fits better these currently collected metrics than our counter model would. This might or might not change in the future, but for now that's the case. Also, there weren't any Boolean values yet so modeling them would be quite premature.

vishh: Currently we have gauges and cumulative metrics. Wouldn't storing cumulative metrics as gauges be confusing?

vishh:
Overall LGTM. Can you add a section in `docs/storage-schema.md` for hawkular that describes the differences between base heapster storage schema and the one you are currently adding for hawkular?

Do Squash your commits at the end.

burmanm:
Squashed and added the documentation to storage-schema.md

sinks/hawkular/driver.go
vishh: @burmanm: Ping

PR #286, opened 2015-05-08
Update integration tests to use gcloud instead of gcutil.

vishh:


vishh:
Ping @vmarmol. e2e is green. I think we can ignore travis.

rjnagal:
LGTM

PR #291, opened 2015-05-14
Fixed kube master url handling

vishh:
Updated docker image to not use ubuntu.
Removed the startup script.
Update the kube config files.


vishh:
ping @vmarmol @rjnagal

vmarmol:
LGTM

PR #294, opened 2015-05-19
Remove extraneous comma

vmarmol:
/cc @vishh 

We should have json validator run for these. Will file an issue.


vishh:
+1 for the validator. There is a pending PR to switch back to yaml format. So we will need a yaml validator.

Thanks for fixing this @vmarmol. Wonder how I missed this while testing manually.

PR #297, opened 2015-05-26
Fixed URL in influxdb kube-config and updated readme

huliz:
I noticed that the kubectl deploy for influxdb did not work - the external and internal URL and version of API was incorrect.


huliz:
I signed it!

vishh:
ok to test

vishh:
Thanks for the fix @huliz!

PR #300, opened 2015-05-28
Adding an in memory cache for all heapster data.

vishh:
As of now the cache is populated by a housekeeping loop.
This cache can serve REST APIs in the future.
Sources will directly write to the cache in the future.
This is a work in progress.

cc @vmarmol


sinks/cache/cache.go
```diff
@@ -0,0 +1,52 @@
...
+
+type Metadata struct {
+	Name      string
+	Namespace string
+	UID       string
+	Hostname  string
+	Labels    map[string]string
+}
+
+type ContainerElem struct {
```
vmarmol: nit: Element here and below.

sinks/cache/cache.go
```diff
@@ -0,0 +1,52 @@
...
+	Metadata
+	ContainerSpec  TimeStore
+	ContainerStats TimeStore
+}
+
+type PodElem struct {
```
vmarmol: Any reason not to use the source API structs?

sinks/cache/cache.go
```diff
@@ -0,0 +1,52 @@
...
+	// map of container name to container element.
+	Containers map[string]*ContainerElem
+	// TODO: Cache history of Spec and Status.
+}
+
+type NodeElem struct {
+	Node *ContainerElem
+	// FreeContainers refers to all the containers in a node
+	// that do not belong to a pod.
+	FreeContainers map[string]*ContainerElem
```
vmarmol: We call them SystemContainers elsewhere.

sinks/cache/cache_impl.go
```diff
@@ -0,0 +1,129 @@
...
+
+	source_api "github.com/GoogleCloudPlatform/heapster/sources/api"
+	cadvisor_api "github.com/google/cadvisor/info/v1"
+)
+
+// TODO: Add unit tests
+type realCache struct {
+	bufferDuration time.Duration
+	// Map of pod UIDs to pod cache entry.
+	Pods map[string]*PodElem
```
vmarmol: Why public?

sinks/cache/cache_impl.go
```diff
@@ -0,0 +1,129 @@
...
+	// Map of node hostnames to node cache entry.
+	Nodes map[string]*NodeElem
+}
+
+const gcDuration = time.Second
+
+func (rc *realCache) newContainerElem() *ContainerElem {
+	return &ContainerElem{
+		ContainerSpec:  NewGCStore(NewTimeStore(), rc.bufferDuration, gcDuration),
```
vmarmol: GCStore starts a thread for garbage collection, this would be a thread per container we are storing which could get quite huge. WDYT if we enforce the eviction age on the TimedStore in an insert and we then GC ContainerElements?

sinks/cache/cache_impl.go
```diff
@@ -0,0 +1,129 @@
...
+			pe.Metadata = Metadata{
+				Name:      pod.Name,
+				Namespace: pod.Namespace,
+				UID:       pod.ID,
+				Hostname:  pod.Hostname,
+				Labels:    pod.Labels,
+			}
+			rc.Pods[pod.ID] = pe
+		}
+		for _, cont := range pod.Containers {
```
vmarmol: Should we also populate the metadata for these containers?

sinks/cache/cache_impl.go
```diff
@@ -0,0 +1,129 @@
...
+				ce = rc.newContainerElem()
+				ce.Metadata = Metadata{
+					Name:     cont.Name,
+					Hostname: cont.Hostname,
+				}
+				ne.FreeContainers[cont.Name] = ce
+			}
+		}
+
+		lastSpec := ce.ContainerSpec.Last().(*cadvisor_api.ContainerSpec)
```
vmarmol: We may be able to share some of the logic for the container insertion.

sinks/cache/gc_store.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/golang/glog"
+)
+
+type gcStore struct {
+	bufferDuration time.Duration
+	store          TimeStore
+}
+
+func (gcs *gcStore) Put(timestamp time.Time, data interface{}) error {
+	if err := gcs.store.Put(timestamp, data); err != nil {
```
vmarmol: nit: return gcs.store.Put(...)

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,116 @@
...
+	"container/list"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+type entry struct {
+	Timestamp time.Time
```
vmarmol: Why public fields?

sinks/cache/memory.go~
```diff
@@ -0,0 +1,100 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
vmarmol: Temp file.

sinks/cache/time_store.go
```diff
@@ -0,0 +1,33 @@
...
+// limitations under the License.
+
+// Package cache implements a cache for time series data.
+package cache
+
+import "time"
+
+// TimeStore represents a timeseries storage.
+// Implementations are expected to be thread safe.
+type TimeStore interface {
```
vmarmol: Wonder whether we should unify this with the cAdvisor one?

vmarmol:
Sent comments, but I will patch this and build off of it :)

sinks/cache/cache.go
```diff
@@ -0,0 +1,52 @@
...
+	// TODO: Cache history of Spec and Status.
+}
+
+type Cache interface {
+	StorePod([]source_api.Pod) error
```
vmarmol: nit: StorePods and StoreContainers since we're giving a slice

sinks/cache/cache_impl.go
```diff
@@ -0,0 +1,129 @@
...
+			ce.ContainerSpec.Put(time.Now(), &cont.Spec)
+		}
+		for idx := range cont.Stats {
+			ce.ContainerStats.Put(cont.Stats[idx].Timestamp, cont.Stats[idx])
+		}
+	}
+	return nil
+}
+
+func NewCache(bufferDuration time.Duration) Cache {
```
vmarmol: Unused bufferDuration?

sinks/cache/cache.go
vishh: I intend to get rid of the source API structs and make the sources write PodElements directly to the cache.

sinks/cache/cache.go
vishh: What about raw cgroups? They are not system containers in the literal sense right?

vmarmol: Not a huge fan of that either, ExternalContainers was another suggestion but that was shut down. It doesn't matter as long as we're consistent though :)

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	ts.rwLock.Lock()
+	defer ts.rwLock.Unlock()
+	if ts.buffer.Len() == 0 {
+		glog.V(5).Infof("put pushfront: %v, %v", timestamp, data)
+		ts.buffer.PushFront(entry{timestamp: timestamp, data: data})
+		return nil
+	}
+	for elem := ts.buffer.Front(); elem != nil; elem = elem.Next() {
+		if timestamp.After(elem.Value.(entry).timestamp) {
```
vmarmol: Should this be the other way around?

```
# List has
[1, 2, 4]

# I do:
Put(3)
1: 3 is after 1? yes

InsertAfter()
```

sinks/cache/cache.go
```diff
@@ -0,0 +1,70 @@
...
+	Containers []*ContainerElement
+	// TODO: Cache history of Spec and Status.
+}
+
+type Cache interface {
+	StorePods([]source_api.Pod) error
+	StoreContainers([]source_api.Container) error
+
+	// GetPods returns a list of pod elements in the cache between 'start' and 'end'.
+	// If 'start' is null, it returns all the elements up until 'end'.
```
vmarmol: Not null, but zero. Here and below.

sinks/cache/gc_store.go
vmarmol: ping

sinks/cache/memory.go~
vmarmol: ping

sinks/cache/in_memory.go
vishh: Discussed offline. This is working as intended. Needs documentation though.

sinks/cache/cache_impl.go
vishh: Fixed

sinks/cache/cache.go
vishh: Done

sinks/cache/time_store.go
vishh: Yeah. As discussed offline, I will try that once the basic framework is in.

sinks/cache/gc_store.go
vishh: Moved real gc to here. Ignoring this nit.

sinks/cache/in_memory.go
vishh: Not necessary. Fixed

sinks/cache/cache_impl.go
vishh: Done

sinks/cache/cache_impl.go
vishh: Makes sense. Done

sinks/cache/cache_impl.go
vishh: Makes sense done.

sinks/cache/cache_impl.go
vishh: Not needed. Fixed.

sinks/cache/cache.go
vishh: Ack.

sinks/cache/cache.go
vishh: Done

vishh:
Unit tests for the cache is on its way. Otherwise this PR is good to go.

sinks/cache/cache.go
vishh: Done

sinks/cache/in_memory.go
vmarmol: Can we document that the buffer is a linked list in reverse chronological order? :)

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	return result
+}
+
+func (ts *timeStore) Get(start, end time.Time) []interface{} {
+	ts.rwLock.RLock()
+	defer ts.rwLock.RUnlock()
+	if ts.buffer.Len() == 0 {
+		return nil
+	}
+	nullTime := time.Time{}
```
vmarmol: nit: zeroTime

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	if start == nullTime {
+		start = time.Unix(0, 0)
```
vmarmol: You can do `time.Time{}` as you do above

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	result := []interface{}{}
+	for elem := ts.buffer.Back(); elem != nil; elem = elem.Prev() {
+		entry := elem.Value.(entry)
+		if entry.timestamp.Before(start) || (end != nullTime && entry.timestamp.After(end)) {
```
vmarmol: I think there is an issue here with start time.

```
Take: [3, 2, 1]
Get(2, 5)

I would expect it to return [2, 3], but it actually returns []

We check:
- is 1 before 2? Yes, break.
```
Issue with [4, 3, 2, 1,] I want 2 - 3?

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (ts *timeStore) Delete(start, end time.Time) error {
+	ts.rwLock.Lock()
+	defer ts.rwLock.Unlock()
+	if ts.buffer.Len() == 0 {
+		return nil
+	}
+	if !end.After(start) {
+		return fmt.Errorf("End time %v is not after Start time %v", end, start)
```
vmarmol: nit: s/End/end/ s/Start/start/

sinks/cache/in_memory.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	elem := ts.buffer.Back()
+	for elem != nil {
+		entry := elem.Value.(entry)
+		if entry.timestamp.Before(start) || entry.timestamp.After(end) {
```
vmarmol: I *think* this has the same start problem as above.

sinks/cache/in_memory_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+func TestNilInsert(t *testing.T) {
+	store := NewTimeStore()
+	assert.Error(t, store.Put(time.Now(), nil))
+}
+
+func TestInsert(t *testing.T) {
+	store := NewTimeStore()
+	now := time.Now()
+	assert.NoError(t, store.Put(now, 2))
```
vmarmol: FYI you can do:

```
assert := assert.New(t)
// use assert without t
```

sinks/cache/in_memory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	assert.NoError(t, store.Put(now.Add(-time.Second), 1))
+	assert.NoError(t, store.Put(now.Add(time.Second), 3))
+	assert.NoError(t, store.Put(now.Add(-2*time.Second), 0))
+	actual := store.Get(time.Time{}, now.Add(time.Second))
+	require.Len(t, actual, 4)
```
vmarmol: You can probably write a helper that is like:

```
func expectElements(t *testing.T, actual interface{}, expected []int)
```

sinks/cache/in_memory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	for i := 0; i < len(actual); i++ {
+		assert.Equal(t, i, actual[i].(int))
+	}
+	actual = store.Get(time.Time{}, time.Time{})
+	require.Len(t, actual, 4)
+	for i := 0; i < len(actual); i++ {
+		assert.Equal(t, i, actual[i].(int))
+	}
+}
```
vmarmol: We should add a test that specifies a start time

sinks/cache/in_memory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	store := NewTimeStore()
+	now := time.Now()
+	assert.NoError(t, store.Put(now, 2))
+	assert.NoError(t, store.Put(now.Add(-time.Second), 1))
+	assert.NoError(t, store.Put(now.Add(time.Second), 3))
+	assert.NoError(t, store.Delete(now.Add(-time.Second), now))
+	actual := store.Get(now.Add(-time.Second), time.Time{})
+	require.Len(t, actual, 1)
+	assert.Equal(t, 3, actual[0].(int))
+}
```
vmarmol: Add a test that specifies a start time in the middle of the range too

sinks/cache/cache_impl.go
```diff
@@ -0,0 +1,230 @@
...
+	lock  sync.RWMutex
+}
+
+const (
+	gcDuration    = time.Second
+	rootContainer = "/"
+	machine       = "machine"
+)
+
+func (rc *realCache) newcontainerElement() *containerElement {
```
vmarmol: nit: newContainerElement, same below with the other three. The function is still internal so it probably differentiates itself from a possible NewContainerElement

sinks/cache/gc_store.go
```diff
@@ -0,0 +1,58 @@
...
+func (gcs *gcStore) Get(start, end time.Time) []interface{} {
+	return gcs.store.Get(start, end)
+}
+
+func (gcs *gcStore) Delete(start, end time.Time) error {
+	return gcs.store.Delete(start, end)
+}
+
+func (gcs *gcStore) reapOldData() {
+	end := time.Now().Add(-gcs.bufferDuration)
```
vmarmol: Should this be from the new point and not from now? The difference in behavior is: "the last 2min of data that has been exported" vs "the data exported in the last 2min of walltime". The first one would keep 2, 1min data points whereas the second will probably only keep one if the delay from collection to storing is high-ish.

sinks/cache/gc_store.go
```diff
@@ -0,0 +1,58 @@
...
+	start := time.Unix(0, 0)
+	if err := gcs.store.Delete(start, end); err != nil {
+		glog.Fatalf("failed to delete old data")
```
vmarmol: nit: add error to the log.

Also, not sure this should be fatal? Maybe just error?

sinks/cache/gc_store_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	time.Sleep(time.Second)
+	// Perform a put to invoke GC.
+	assert.NoError(t, gcStore.Put(time.Now(), struct{}{}))
+	data := gcStore.Get(now, time.Now())
+	assert.Len(t, data, 0)
+}
+
+func TestGCDetail(t *testing.T) {
+	gcStore := NewGCStore(NewTimeStore(), time.Second, time.Microsecond)
```
vmarmol: This could be race, since we control the timestamp going in, why not customize it? Take now and subtract an hour, make the gc time an hour.

sinks/cache/in_memory.go
vishh: Done

sinks/cache/in_memory.go
vishh: Done

sinks/cache/in_memory.go
vishh: Yeah. Thats true

sinks/cache/in_memory.go
vishh: Good catch. There is an issue. I think I have fixed it this time.

sinks/cache/in_memory.go
vishh: Done

sinks/cache/in_memory.go
vishh: True. Fixed.

sinks/cache/in_memory_test.go
vishh: Ah. Thats very helpful. Done

sinks/cache/in_memory_test.go
vishh: Done

sinks/cache/in_memory_test.go
vishh: There is one as part of TestDelete()

sinks/cache/in_memory_test.go
vishh: Done

sinks/cache/gc_store.go
vishh: We can store points out of order. Based on how we end up storing data, we might end up clearing more points that we intended to. We can take the most recent point's timestamp for GC. I wonder if it is necessary to keep data points that are older than the buffer duration?

sinks/cache/in_memory_test.go
vmarmol: Seems weird to have that in TestDelete though...instead of TestInsert. Wouldn't be where you would expect to find it.

vmarmol:
Note the new comments btw :)

sinks/cache/gc_store.go
vmarmol: Yeah using the newest point makes sense. What do you mean about keeping the
older data points? Isn't that what we want to gc?

sinks/cache/gc_store_test.go
vishh: Done

sinks/cache/gc_store.go
vishh: Ah! Missed that. Fixed

sinks/cache/cache_impl.go
vishh: Done

vishh:
Fixed the new set of comments as well. Thanks for the through review @vmarmol. I apologize for the size of this PR

sinks/cache/gc_store.go
vmarmol: This is the last remaining commend, otherwise looks good :D

vishh: Fixed.

vmarmol:
LGTM, thanks for fixing all the nits @vishh!

PR #303, opened 2015-05-30
Metrics Export REST API

vmarmol:
Building off of #300 with some fixes and the plumbing for the REST API. Still working on the metrics conversion.


vmarmol:
This should be ready for review. It is the interface and the logic to get the data from the cache. But there is no cache so we don't return useful information today :)

/cc @vishh @a-robinson

sinks/external.go
```diff
@@ -26,6 +27,7 @@ import (
 type externalSinkManager struct {
 	decoder       sink_api.Decoder
 	externalSinks []sink_api.ExternalSink
+	cache         cache.Cache
```
vishh: I was thinking of keeping the cache as part of heapster manager instead of here since it holds data that is not in metrics form.

vmarmol: Done.

vmarmol:
@a-robinson that doesn't seem unreasonable. I changed the data structure PTAL. We not expose all the metrics for the same target schema together.

api/v1/api.go
```diff
@@ -0,0 +1,86 @@
...
+	return &Api{
+		manager: m,
+	}
+}
+
+// Register the Api on the specified endpoint.
+func (a *Api) Register(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/api/v1/metric-export").
```
vishh: Can this API be "metrics" instead of "metric-export"?

vmarmol: I think this is purpose-built for this usecase, I'd leave metrics for a more generic API we build for querying. Worst case, we symlink them in the future.

vishh:
LGTM

api/v1/api.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (a *Api) exportMetrics(request *restful.Request, response *restful.Response) {
+	points, err := a.manager.ExportMetrics()
+	if err != nil {
+		response.WriteError(http.StatusInternalServerError, err)
+		return
+	}
+
+	// Group points by labels.
```
a-robinson: Does this only group by the target labels (the ones that identify a container), or will it also group by the metric-specific labels, like device_name and fault_type?

a-robinson:
Thanks for the adjustment, @vmarmol. De-duping the label values like this should also reduce the data sent over the wire by quite a bit.

vishh:
LGTM

api/v1/api.go
vmarmol: It will also do by metric-specific labels, we can do this way or by target labels only if you prefer.

vmarmol: Also, fault_type we expose as different metrics today, so we'll need to sync up on that.

a-robinson: Target labels only would definitely be preferable, but we could live with this if grouping only by target labels is too difficult.

vmarmol: It shouldn't be, we know what labels are for the target schema. Will update and ping again.

vmarmol:
PTAL at the API @a-robinson, still working through integration issues, but the API should be good to go.

api/v1/types.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"time"
+)
+
+// Timeseries represents a set of metrics for the same target object
+// (typically a container).
+type Timeseries struct {
+	// Map of metric names to their value.
+	Metrics map[string]Point `json:"metrics"`
```
a-robinson: Doesn't this have to be a map from metric names to more than just a single point? A single target-metric pair can have more than one point if the metric has additional labels. For example, if I have one disk-bytes-used data point for device_name=/dev/sda1 and one point for device_name=/dev/sda2, both from the same container, I'm not sure how this schema supports exposing both?

vmarmol: Ah good catch, I was overwriting them today. Should be fixed.

a-robinson:
API LGTM, thanks!

vmarmol:
Rebased and retested. All is working as expected and we return metrics as expected. PTAL @vishh

vishh:
Thanks @vmarmol! LGTM

PR #311, opened 2015-06-04
Update fleetEndpoint of heapster.

gosharplite:


vishh:
LGTM. Thanks for the fix @gosharplite!

PR #313, opened 2015-06-05
Adding unit tests for the cache.

vishh:
Fixed a few bugs in the cache.

PR #314, opened 2015-06-05
Removed go vet dependency

afein:
Vet is now part of the golang distribution, through the go.tools subdirectory. 
Therefore, the dependency should be satisfied by the go distribution itself, rather than go get.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ all: build
 deps:
 	go get github.com/tools/godep
 	go get github.com/progrium/go-extpoints
-	go get golang.org/x/tools/cmd/vet
 
 build: clean deps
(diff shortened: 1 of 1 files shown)


vishh:

vishh:
We can ignore the travis failure. It will be fixed in #313. LGTM

rjnagal:
Doesn't that mean we need to update travis to 1.5

For Go 1.4 you still need to go install golang.org/x/tools/cmd/vet. For Go 1.5 this may not be necessary.

vishh:
@rjnagal: Travis seems to be happy with not having a vet dependency. The current failure is due to a flaky test.

vishh:
LGTM

PR #315, opened 2015-06-05
Use the cache for external sinks.

vishh:
Events needs to be pushed to the cache before this PR can be functional.


vishh:
@afein: Still a WIP. But managed to rebase.

vishh:
@afein: Can I get a quick review? I would like for this to go in before hitting any more merge conflicts.

afein:
LGTM

sinks/external_test.go
```diff
@@ -129,28 +132,50 @@ func TestSetSinksRegisterAgain(t *testing.T) {
...
+	f.Fuzz(&pods)
+	f.Fuzz(&containers)
+	f.Fuzz(&events)
+	c.StorePods(pods)
+	c.StoreContainers(containers)
+	c.StoreEvents(events)
+	stopChan = m.Sync()
+	as.NotNil(stopChan)
+	time.Sleep(time.Second)
 	as.Equal(2, s1.StoredTimeseries)
```
afein: @vishh: I've seen some test runs where s1.StoredTimeseries is 1, instead of 2.

mvdan: FWIW, I fixed a race condition in https://github.com/kubernetes/heapster/pull/405 - there might be another one, and if similar it could be causing something like this. Since `Store()` would return before it's actually finished.

vishh: @afein: Care to send a PR to invoke the local `sync` method instead of the one that spawns a go thread?

PR #317, opened 2015-06-08
Add major page fault metric

vmarmol:
/cc @vishh

### DIFF
--- a/docs/storage-schema.md
+++ b/docs/storage-schema.md
@@ -2,18 +2,19 @@
 
 Heapster exports the following metrics to its backends.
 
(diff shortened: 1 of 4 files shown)


vishh:
Wonder why the integration test is failing.

vmarmol:
Passed, although I forgot to update the storage schema, will do that now.

vmarmol:
Pushed the new commit with the doc updates.

vishh:
I don't see the docs @vmarmol

vmarmol:
Ooops my bad, should be updated now @vishh

vishh:
LGTM

PR #322, opened 2015-06-09
Adding metrics for limits in CPU, memory, and disk

vmarmol:
This simply adds the metrics and a field that lets us specify to only push them if the value has changed. We'll need a fair amount of logic to keep track of old values and not export them if they have changed. I considered it an optimization and left it out of this PR. Let me know if you disagree :)

/cc @vishh

### DIFF
--- a/sinks/api/supported_metrics.go
+++ b/sinks/api/supported_metrics.go
@@ -56,6 +56,22 @@ var statMetrics = []SupportedStatMetric{
 	},
 	{
 		MetricDescriptor: MetricDescriptor{
+			Name:        "cpu/limit",
+			Description: "CPU limit in millicores",
+			Type:        MetricGauge,
+			ValueType:   ValueInt64,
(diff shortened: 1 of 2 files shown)


sinks/api/supported_metrics.go
```diff
@@ -56,6 +56,22 @@ var statMetrics = []SupportedStatMetric{
 	},
 	{
 		MetricDescriptor: MetricDescriptor{
+			Name:        "cpu/limit",
+			Description: "CPU limit in millicores",
+			Type:        MetricGauge,
+			ValueType:   ValueInt64,
+			Units:       UnitsNanoseconds,
```
vishh: This needs to be UnitsCount since cores is not in nanoseconds.

vishh:
Exporting only when it changes is an optimization. I am fine with this PR as-is for now.

vishh:
Closing this PR in favor of #326

Issue #320, opened 2015-06-08
Trying to run heapster on openshift3


shilpapadgaonkar:
Hi,
I am trying to run heapster on openshift3. I followed the documentation link https://github.com/GoogleCloudPlatform/heapster/blob/master/docs/influxdb.md

I also made changes to remove the environment variable INFLUXDB_EXTERNAL_URL from InfluxDB Grafana RC config and added a route so that Grafana service become reachable as an external service.

All services were created. InfluxDB and grafana started without issues. I am able to see the Grafana UI.

The monitoring-heapster-controller does not start though. It keeps showing me a pending state. I am not sure why..

I have read in the docu somewhere that I could provide heapster a kubernetes source explicitly. Do I need to do this while trying to run it on openshift3 ? Could you please let me know how I could do this?

Thanks in advance.

Regards,
Shilpa

vishh:
Can you post logs from heapster pod? (`kubectl logs <heapster pod name>`)

shilpapadgaonkar:
I could not get pod logs as the pod is not running yet. See attached pic
![heapster-pic](https://cloud.githubusercontent.com/assets/3367612/8051752/a92063a8-0e80-11e5-8d31-511fb50ead08.PNG)


The openshift master logs showed some errors though...
![openshift-master-logs](https://cloud.githubusercontent.com/assets/3367612/8051757/b73e810e-0e80-11e5-9147-b0655b6f601e.PNG)

detiber:
@shilpapadgaonkar I've been working on this myself.  I've gotten a bit further than you have, but am still hitting a roadblock on the correct service URL to use internal to the cluster.  Here is my modified replication controller:
```
[openshift@ip-172-19-47-10 ~]$ cat heapster-rc.json 
{
    "apiVersion": "v1beta3",
    "kind": "ReplicationController",
    "metadata": {
	"labels": {
	    "name": "heapster"
	},
	"name": "monitoring-heapster-controller"
    },
    "spec": {
	"replicas": 1,
	"selector": {
	    "name": "heapster"
	},
	"template": {
	    "metadata": {
		"labels": {
		    "name": "heapster"
		}
	    },
	    "spec": {
		"containers": [
		    {
			"image": "kubernetes/heapster:v0.13.0",
			"name": "heapster",
			"command": [
			    "/heapster",
			    "--source=kubernetes:http://kubernetes-ro.default.local?auth=",
[...]

PR #324, opened 2015-06-09
Fixes #323: Enable use of service accounts

### LINKED ISSUE #323: Support for Kubernetes service accounts
As detailed at https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/design/service_accounts.md.

Should be a simple case of reusing bearer token at `/var/run/secrets/kubernetes.io/servicaccount/token`.

jimmidyson:


jimmidyson:
Fixes #320 too, or at least makes it easier to run on OpenShift.

vishh:

vishh:
@jimmidyson: Can you add some documentation to go along with this PR? This will help support new users on OpenShift.

sources/kube_factory.go
```diff
@@ -31,11 +32,13 @@ import (
...
-	defaultKubeletPort    = 10255
-	defaultKubeConfigFile = "/etc/kubernetes/kubeconfig/kubeconfig"
-	defaultKubeletHttps   = false
+	defaultApiVersion         = "v1beta3"
+	defaultInsecure           = false
+	defaultKubeletPort        = 10255
+	defaultKubeConfigFile     = "/etc/kubernetes/kubeconfig/kubeconfig"
+	defaultKubeletHttps       = false
+	defaultUseServiceAccount  = false
+	defaultServiceAccountFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
vishh: Will `/var/run/secrets' be auto mounted for the heapster pod?

jimmidyson: Yeah that's how service accounts work. You create a service account, a token is generated, then you can reference that service account in your pod spec & token is automounted.

vishh: How about adding an example config for running heapster in OpenShift? Does it already exist elsewhere?

jimmidyson: I'm working on it for OpenShift Origin docs & will either link to it from here or add it here too. There are some problems to work through as OpenShift is locked down right now so accessing /stats endpoint on the kubelet isn't possible... working on that too,

docs/source-configuration.md
```diff
@@ -28,6 +28,20 @@ To run without auth file, use the following config:
...
+```
+cat <EOF | kubectl create -f -
+apiVersion: v1beta3
+kind: ServiceAccount
+metadata:
+  name: heapster
+EOF
+```
+
+You will then need to reference the service account in your Heapster pod spec.
```
vishh: Can you provide a pointer to an example PodSpec or may be some pod that uses service accounts?

vishh:
LGTM. Thanks @jimmidyson

PR #334, opened 2015-06-11
Additional node metrics for autoscaling.

jszczepkowski:
Added additional node metrics (tracking cpu and mem usage) together with additional lables. They are in format compatible with cloud autoscaler, and they can be used to autoscale the numer of nodes in a cluster.

### DIFF
--- a/sinks/api/supported_labels.go
+++ b/sinks/api/supported_labels.go
@@ -15,14 +15,16 @@
 package api
 
 const (
-	LabelPodId           = "pod_id"
-	LabelPodName         = "pod_name"
-	LabelPodNamespace    = "pod_namespace"
-	LabelPodNamespaceUID = "namespace_id"
-	LabelContainerName   = "container_name"
-	LabelLabels          = "labels"
-	LabelHostname        = "hostname"
-	LabelResourceID      = "resource_id"
+	LabelPodId               = "pod_id"
+	LabelPodName             = "pod_name"
+	LabelPodNamespace        = "pod_namespace"
+	LabelPodNamespaceUID     = "namespace_id"
+	LabelContainerName       = "container_name"
+	LabelLabels              = "labels"
+	LabelHostname            = "hostname"
+	LabelResourceID          = "resource_id"
+	LabelComputeResourceID   = "compute.googleapis.com/resource_id"
+	LabelComputeResourceType = "compute.googleapis.com/resource_type"
 )
 
 // TODO(vmarmol): Things we should consider adding (note that we only get 10 labels):
@@ -60,6 +62,14 @@ var allLabels = []LabelDescriptor{
 	{
 		Key:         LabelResourceID,
 		Description: "Identifier(s) specific to a metric",
+	},
+	{
+		Key:         LabelComputeResourceID,
+		Description: "Resource id for machines (GCE, used for auto-scaling)",
+	},
+	{
+		Key:         LabelComputeResourceType,
+		Description: "Resource type for machines (GCE, used for auto-scaling)",
 	},
 }
 
--- a/sinks/api/supported_metrics.go
+++ b/sinks/api/supported_metrics.go
@@ -17,6 +17,7 @@ package api
 import (
 	"time"
 
+	source_api "github.com/GoogleCloudPlatform/heapster/sources/api"
 	cadvisor "github.com/google/cadvisor/info/v1"
 )
 
@@ -32,10 +33,10 @@ var statMetrics = []SupportedStatMetric{
 			ValueType:   ValueInt64,
 			Units:       UnitsMilliseconds,
 		},
-		HasValue: func(spec *cadvisor.ContainerSpec) bool {
+		HasValue: func(spec *source_api.ContainerSpec) bool {
 			return !spec.CreationTime.IsZero()
 		},
-		GetValue: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []internalPoint {
+		GetValue: func(spec *source_api.ContainerSpec, stat *cadvisor.ContainerStats) []internalPoint {
 			return []internalPoint{{value: timeSince(spec.CreationTime).Nanoseconds() / time.Millisecond.Nanoseconds()}}
 		},
 	},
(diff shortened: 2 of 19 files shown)


jszczepkowski:
@vishh @piosz

sinks/api/supported_labels.go
```diff
@@ -15,13 +15,15 @@
...
-	LabelHostname      = "hostname"
-	LabelResourceID    = "resource_id"
+	LabelPodId               = "pod_id"
+	LabelPodName             = "pod_name"
+	LabelPodNamespace        = "pod_namespace"
+	LabelContainerName       = "container_name"
+	LabelLabels              = "labels"
+	LabelHostname            = "hostname"
+	LabelResourceID          = "resource_id"
+	LabelComputeResourceID   = "compute.googleapis.com/resource_id"
```
vishh: Any reason why the 'compute.googleapis.com' prefix is necessary? We can add that when we push to GCM. The labels here are meant to be generic to monitoring backends.

sinks/api/supported_metrics.go
```diff
@@ -255,7 +256,50 @@ var statMetrics = []SupportedStatMetric{
...
+	{
+		MetricDescriptor: MetricDescriptor{
+			Name:        "cpu/node_usage",
+			Description: "Cumulative CPU usage on all cores on a node",
+			Type:        MetricCumulative,
+			ValueType:   ValueInt64,
+			Units:       UnitsNanoseconds,
+		},
+		HasValue: func(spec *source_api.ContainerSpec) bool {
+			return spec.HasCpu && spec.HasResourceId
```
vishh: The heapster core is in a bit of a flux. I apologize for the inconvenience. 
We currently push node usage as part of `cpu/usage` metric, with `container_name` label set to `machine`. Will that suffice your use case? 
If not, this change will not expose node_usage if resourceId is not set, which might be the case in baremetal deployments. Should we instead annotate ContainerSpec with either a Container name or isNode boolean and use that here?
If resource ID is available we can push that label as part of node_usage.

sinks/api/supported_metrics.go
```diff
@@ -255,7 +256,50 @@ var statMetrics = []SupportedStatMetric{
...
+					LabelComputeResourceID:   spec.ResourceId,
+					LabelComputeResourceType: "instance",
+				},
+			}
+			return []internalPoint{result}
+		},
+	},
+	{
+		MetricDescriptor: MetricDescriptor{
+			Name:        "memory/node_usage",
```
vishh: As mentioned above, 'memory/usage' should include node stats as well. Label "container_name=machine" is the identifier. If that is inconvenient, we can avoid including node metrics in `memory/usage`. WDYT?

sources/nodes/types.go
```diff
@@ -28,6 +28,8 @@ type Info struct {
 	// use to communicate with the host sinc Public IP access is usually
 	// restricted behind firewalls.
 	InternalIP string `json:"internal_ip,omitempty"`
+	// External ID of the node (e.g. assigned by a cloud provider)
```
vishh: Why not call it `ExternalID`? That way it remains generic.

vishh:

sinks/api/supported_labels.go
jszczepkowski: We need labels of our metrics to be in a format compatible with google cloud autoscaler. So, we need to have prefix "compute.googleapis.com" for them. I'm afraid we cannot change the labels later when pushing to GCM. Maybe @tkulczynski can comment on this.

sinks/api/supported_metrics.go
jszczepkowski: We don't want to use regular `cpu/usage` metric, as it contains info from both machines and containers. We want our metric to contain info for **machines only**.

I agree that annotating `ContainerSpec` with `isNode` is better idea than basing on `resourceId` to be not empty. We will apply it.

sinks/api/supported_metrics.go
jszczepkowski: The same as above: `memory/usage` is polluted with container stats.

sources/nodes/types.go
jszczepkowski: I agree, we should change it.

jszczepkowski:
This PR will be taken over by piosz@, as I'm on vacation for the next two weeks.

sinks/api/supported_metrics.go
vishh: Essentially then, `node/usage` is the kernel overhead?

sinks/api/supported_labels.go
vishh: The GCM driver under `sinks/gcm/` already prefixes individual metrics. Can we move this logic there?

sinks/api/supported_metrics.go
vishh: `stat.Memory.Usage` is the entire's node's usage right? If your goal is to calculate system overhead, then it has to be `node_usage - usage of all containers under '/'`

vishh:
@jszczepkowski: Can we sync up offline on the specific goals of this PR? Depending on the goals, there could be alternate pathways to getting metrics plumbed into Google Cloud Monitoring.

sinks/api/supported_metrics.go
piosz: Nope. It's kernel overhead + kubernetes overhead + user's pods usage.

sinks/api/supported_metrics.go
piosz: Same as above.

sinks/api/supported_labels.go
piosz: I can do it, but there will be name collision. There is already label named "resource_id". The solution is to change the value of LabelResourceID. WDYT?

sinks/api/supported_metrics.go
piosz: Changed logic to use IsNode istead of HasResourceId.

sources/nodes/types.go
piosz: Done

sinks/api/supported_labels.go
vishh: Ah! Naming is hard :( Lets not overload "resource_id", since it is required for filesystem metrics.
Does "instance id" have to be called "resource_id"? Can we call it "host_id" instead in GCM side?

sinks/api/supported_metrics.go
vishh: Can you elaborate your comment?

vishh:
@jszczepkowski: Are you continuing to work on this PR? If so can you rebase to HEAD?

sinks/api/supported_labels.go
piosz: Nope. According to https://cloud.google.com/compute/docs/autoscaler/scaling-cloud-monitoring-metrics#custom_metrics_beta it has to be exactly:
compute.googleapis.com/resource_type
compute.googleapis.com/resource_id

sinks/api/supported_metrics.go
piosz: See https://github.com/GoogleCloudPlatform/heapster/pull/334#discussion_r32456815

vishh:
Closing this PR in favor of #375

PR #343, opened 2015-06-15
Updating deps

vishh:
This is big PR. I had to do it in one go since godep was mad with cadvisor and kubernetes both including libcontainer.


vishh:
cc @rjnagal

vishh:
@saad-ali can you review this PR?
Rohit might be busy.

vishh:
Self merging since a release needs to be cut soon.

saad-ali:
Way too many files to look at. But I assume you did some manual testing to verify things are still working. If so, LGTM.

rjnagal:
LGTM

PR #345, opened 2015-06-16
Adding External ID as a label

vishh:
Fixed issue where metric specific labels were not being populated in the API.

### DIFF
--- a/sinks/api/supported_labels.go
+++ b/sinks/api/supported_labels.go
@@ -23,7 +23,7 @@ const (
 	LabelLabels          = "labels"
 	LabelHostname        = "hostname"
 	LabelResourceID      = "resource_id"
-	LabelExternalID      = "external_id"
+	LabelHostID          = "host_id"
 )
 
 // TODO(vmarmol): Things we should consider adding (note that we only get 10 labels):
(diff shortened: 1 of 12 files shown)


vishh:
Depends on #343

vishh:
cc @rjnagal

sinks/api/supported_labels.go
```diff
@@ -22,6 +22,7 @@ const (
 	LabelLabels        = "labels"
 	LabelHostname      = "hostname"
 	LabelResourceID    = "resource_id"
+	LabelExternalID    = "external_id"
```
a-robinson: It's not my API, but I'd much prefer a name that makes it clear that it's the external ID of the host. Otherwise it's not clear which object the ID belongs to, and the default that I'd expect if I didn't know better is the container.

Could be any of host_external_id, host_id, node_external_id, node_id, etc.

a-robinson:
LGTM other than the naming issue

rjnagal:
Agree w/ the naming comments. Applies to ResourceID too?

sinks/api/supported_labels.go
vishh: Good point. Done. Settled with 'host_id'.

vishh:
Thanks for the quick review. Updated the label name. PTAL

vishh:
@rjnagal: E2e is green

rjnagal:
LGTM

PR #349, opened 2015-06-16
[Schema 1] Definitions for exportable types

afein:
@rjnagal @vishh This is part 1 of #342 
The presented types are intended for use from both within the module and from external REST API consumers. They represent a hierarchical in-memory structure of the cluster's historical state.

Open Issues:
- Limitations on the historical data length of []*MetricTimeseries
- Processing of Kubernetes' event stream for real-time information regarding container/pod/node deletions

### DIFF
--- a/api/schema/types.go
+++ b/api/schema/types.go
@@ -1,0 +1,79 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+import (
+	"github.com/GoogleCloudPlatform/heapster/sinks/cache"
+	"sync"
+	"time"
+)
+
+type Cluster interface {
+	// The Update operation populates the Cluster from a cache
+	Update(*cache.Cache) error
+
+	// The Get operations extract internal types from the Cluster
+	GetAllClusterData() (*ClusterInfo, time.Time, error)
+	GetAllNodeData(string) (*NodeInfo, time.Time, error)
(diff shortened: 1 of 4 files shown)


rjnagal:
ok to test

api/schema/types.go
```diff
@@ -0,0 +1,79 @@
...
+	lock      *sync.RWMutex
+	ClusterInfo
+}
+
+// Internal Types
+// REST consumption requires conversion to the corresponding versioned API types
+
+type InfoType struct {
+	// TODO: replace []*MetricTimeSeries with a BoundedTimeStore
+	Metrics map[string][]*MetricTimeseries // key: Metric name
```
vishh: You can start with the existing GcStore which is not bounded and then replace that with a Bounded one.

api/schema/types.go
```diff
@@ -0,0 +1,79 @@
...
+	Labels  map[string]string              // key: Label
+}
+
+type ClusterInfo struct {
+	InfoType
+	Namespaces map[string]*NamespaceInfo // key: Namespace Name
+	Nodes      map[string]*NodeInfo      // key: Hostname
+}
+
+type NamespaceInfo struct {
```
vishh: Should we keep the kubernetes versioned namespace objects around as well?

api/schema/types.go
```diff
@@ -0,0 +1,79 @@
...
+	// Implementation of Cluster
+	timestamp time.Time // Marks the last update from a cache.
+	lock      *sync.RWMutex
+	ClusterInfo
+}
+
+// Internal Types
+// REST consumption requires conversion to the corresponding versioned API types
+
+type InfoType struct {
```
vishh: This is essentially a MetricsStore right?

api/schema/types.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
vishh: Create a separate package under heapster and place all the derived logic in there. Let the 'api' package only deal with REST endpoints.

vishh:
Just a few comments. Overall LGTM

api/schema/types.go
afein: Yes, along with Labels. Labels are only provided at the Pod level at the moment, and maintaining them inside the InfoType will be helpful in propagating them up and down the hierarchy.

afein:
@vishh addressed comments in the latest commit

vishh:
LGTM. Will merge once CI passes.

vishh:
FYI: You can an IDE hook for running gofmt on save.

PR #350, opened 2015-06-16
Fix issue with exporting host_id for pod containers.

### LINKED ISSUE #348: host_id label only populated some of the time
In the heapster 0.14.1 HTTP API, only some of the timeseries I'm receiving come with host_id populated, even though hostname is always populated.

In a default kubernetes cluster with no custom pods running, it looks like host_id is populated for the "/system", "/kubelet", "/kube-proxy", and "/docker-daemon", but not for anything else (i.e. "machine" or any actual pods).

vishh:
Fixes #348 

cc @a-robinson @rjnagal

### DIFF
--- a/sinks/api/decoder_v2.go
+++ b/sinks/api/decoder_v2.go
@@ -93,7 +93,6 @@ func (self *v2Decoder) getContainerMetrics(container *cache.ContainerElement, la
 		return nil
 	}
 	labels[LabelContainerName] = container.Name
-	labels[LabelHostID] = container.ExternalID
 	// One metric value per data point.
 	var result []Timeseries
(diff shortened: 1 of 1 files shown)


rjnagal:
What other endpoint do we export host_id at?

vishh:
The monitoring sinks like InfluxDB should have it.

vishh:
Here is a running instance of heapster with this change: http://104.197.27.4:8082/api/v1/metric-export

a-robinson:
I can't hit that endpoint, but LGTM.

rjnagal:
LGTM

vishh:
@a-robinson: Try again.

a-robinson:
Much better, thanks!

PR #354, opened 2015-06-17
wrong docker image label of kubernetes/heapster

njuicsgz:
wrong docker image label of kubernetes/heapster


rjnagal:
Thanks for catching this, @njuicsgz 

We should just push a new image tag v0.14.2 to be consistent with the other tags.
@vishh Can you do a push today?

vishh:
I updated the hub to have an image with v0.14.2 tag. Thanks for this PR!

njuicsgz:
good to me.

PR #357, opened 2015-06-17
Move TimeStore definition and implementations outside of cache module

afein:
This PR moves the definition and implementations of TimeStores outside of the cache package, along with related references. 

Sidenote: There might be confusion between store.gcStore and utils.gcstore.GCStore


vishh:
ok to test

vishh:
LGTM

vishh:
Added you to the jenkins server. PRs should be built automatically henceforth.

PR #358, opened 2015-06-17
[Schema 2] Constructors and Getters

alexmavr:
Part 2 of #342 
Added constructors and implementation for three basic "get" methods.

As this series of PRs is generated in order of dependencies, Part 3 will include the addPod, addNode, addNamespace and addContainer methods.
Unit tests for Part 2 will be provided in later parts (estimate: part 6)

### DIFF
--- a/schema/impl.go
+++ b/schema/impl.go
@@ -88,5 +88,6 @@ func (rc *realCluster) GetAllPodData(namespace string, pod_name string) (*PodInf
 }
 
 func (rc *realCluster) Update(c *cache.Cache) error {
+	// TODO(afein): Unimplemented
 	return nil
 }
--- a/schema/impl.go
+++ b/schema/impl.go
@@ -57,7 +57,7 @@ func (rc *realCluster) GetAllNodeData(hostname string) (*NodeInfo, time.Time, er
 
 	res, ok := rc.Nodes[hostname]
 	if !ok {
-		return nil, zeroTime, fmt.Errorf("Unable to find node with hostname: %s", hostname)
+		return nil, zeroTime, fmt.Errorf("unable to find node with hostname: %s", hostname)
 	}
 
 	return res, rc.timestamp, nil
@@ -71,17 +71,17 @@ func (rc *realCluster) GetAllPodData(namespace string, pod_name string) (*PodInf
 	defer lock.RUnlock()
 
 	if len(rc.Namespaces) == 0 {
-		return nil, zeroTime, fmt.Errorf("Unable to find pod: no namespaces in cluster")
+		return nil, zeroTime, fmt.Errorf("unable to find pod: no namespaces in cluster")
 	}
 
 	ns, ok := rc.Namespaces[namespace]
 	if !ok {
-		return nil, zeroTime, fmt.Errorf("Unable to find namespace with name: %s", namespace)
+		return nil, zeroTime, fmt.Errorf("unable to find namespace with name: %s", namespace)
 	}
 
 	pod, ok := ns.Pods[pod_name]
 	if !ok {
-		return nil, zeroTime, fmt.Errorf("Unable to find pod with name: %s", pod_name)
+		return nil, zeroTime, fmt.Errorf("unable to find pod with name: %s", pod_name)
 	}
 
 	return pod, rc.timestamp, nil
--- a/schema/impl.go
+++ b/schema/impl.go
@@ -15,7 +15,7 @@
 package schema
 
 import (
-	"errors"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


schema/impl.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/GoogleCloudPlatform/heapster/sinks/cache"
+)
+
+func NewCluster() Cluster {
```
vishh: Why is this wrapper needed? Can NewRealCluster() be private?

schema/impl.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func NewRealCluster() *realCluster {
+	cinfo := ClusterInfo{
+		InfoType:   newInfoType(nil, nil),
+		Namespaces: make(map[string]*NamespaceInfo),
+		Nodes:      make(map[string]*NodeInfo),
+	}
+	cluster := &realCluster{
+		timestamp:   time.Time{},
```
vishh: Should the timestamp be time.Now()?

schema/impl.go
```diff
@@ -0,0 +1,99 @@
...
+		lock:        new(sync.RWMutex),
```
vishh: Is this required? Why cant the lock be a variable?

schema/impl.go
afein: The timestamp of the realCluster signifies the last update from the cache. Therefore, for the first invocation we would like to invoke the Cache methods with parameters (start=zeroTime, end=zeroTime) to extract all the contents

vishh: Ack

schema/impl.go
```diff
@@ -0,0 +1,100 @@
...
+	var res *NodeInfo
+	var stamp time.Time
+	var err error
+
+	if val, ok := rc.Nodes[hostname]; ok {
+		res = val
+		stamp = rc.timestamp
+		err = nil
+	} else {
+		res = nil
```
vishh: Why not `return nil, stamp, fmt.Errorf(...)` instead? 
Another style nit: The pattern that we usually follow is that of handling the failure case first and then moving on with the rest of the logic.

vishh:
Overall LGTM. Just a couple of comments.

schema/impl.go
afein: addressed on the next commit

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/GoogleCloudPlatform/heapster/sinks/cache"
+)
+
+var lock sync.RWMutex
```
rjnagal: why is this a package global? It only seems to be protecting realCluster?

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+func (rc *realCluster) GetAllNodeData(hostname string) (*NodeInfo, time.Time, error) {
+	// Returns the corresponding NodeInfo object
+	var zeroTime time.Time
+
+	lock.RLock()
+	defer lock.RUnlock()
+
+	res, ok := rc.Nodes[hostname]
+	if !ok {
+		return nil, zeroTime, fmt.Errorf("Unable to find node with hostname: %s", hostname)
```
rjnagal: s/Unable/unable

go convention for errors is to start with lower case. Here and below.

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+		Nodes:      make(map[string]*NodeInfo),
+	}
+	cluster := &realCluster{
+		timestamp:   time.Time{},
+		ClusterInfo: cinfo,
+	}
+	return cluster
+}
+
+func (rc *realCluster) GetAllClusterData() (*ClusterInfo, time.Time, error) {
```
rjnagal: why do we use "All" in every Get*Data()? It makes me think we'll add interfaces to pull out partial data (field-specific). Is that true?

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+	// Returns the entire ClusterInfo object
+	lock.RLock()
+	defer lock.RUnlock()
+	return &rc.ClusterInfo, rc.timestamp, nil
+}
+
+func (rc *realCluster) GetAllNodeData(hostname string) (*NodeInfo, time.Time, error) {
```
rjnagal: Are there any restrictions on hostname string? Is this internal hostname from heapster, or does it have to be fqdn? Doesn't matter if it is called after getting hostname from heapster, but we should document it.

rjnagal: Document the return types - what is the returned time - time of data collection? Is that relevant to caller of GetAll*Data()? Why do they need that?

Adding header comments to methods would help the users understand how to use the interface.

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (rc *realCluster) GetAllPodData(namespace string, pod_name string) (*PodInfo, time.Time, error) {
+	// Returns the corresponding NodeInfo object
+	var zeroTime time.Time
+
+	lock.RLock()
+	defer lock.RUnlock()
+
+	if len(rc.Namespaces) == 0 {
```
rjnagal: Is the default namespace also part of rc.Namespaces? In that case, this shouldn't ever happen if we have any data?

rjnagal: Is this check necessary? If we didn't have this one, user would get a better error message that the namespace they asked for does not exist.

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	ns, ok := rc.Namespaces[namespace]
+	if !ok {
+		return nil, zeroTime, fmt.Errorf("Unable to find namespace with name: %s", namespace)
+	}
+
+	pod, ok := ns.Pods[pod_name]
+	if !ok {
+		return nil, zeroTime, fmt.Errorf("Unable to find pod with name: %s", pod_name)
```
rjnagal: Also add namespace we are looking at in this error message.

schema/impl.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return pod, rc.timestamp, nil
+}
+
+func (rc *realCluster) Update(c *cache.Cache) error {
+	return nil
```
rjnagal: Add  //TODO(afein): Unimplemented to indicate that this is WIP.

This is a general pattern that will help us easily figure out which pieces are coming.

schema/impl.go
rjnagal: Is returning pointers for Node and Pods etc safe in case the caller mutates it? We should add a comment if it is.

schema/impl.go
afein: My initial thinking was to have two methods for each exposable resource, "All" and "New". I don't believe there is any reason to explore field-specific interfaces, at least not at this stage.
GetAllXData() -> returns all metrics and structure under X
GetNewXData(start time.Time) -> returns structure under X and metrics with timestamp > start

"Get" functions return the cluster timestamp at the time of the request. Therefore, the user will be able to get only fresh data by calling the GetNewX methods with that timestamp as a parameter. This can prove useful for AJAX calls

Note: The timestamp signifies the latest metric "end" time, as provided by cadvisor, rather than time.Now(). This allows us to skip timezone conversions and calculations of the time difference from metric generation on cadvisor, until consumption through the heapster API

schema/util.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"time"
+
+	"github.com/GoogleCloudPlatform/heapster/store"
+)
+
+func maxTimestamp(first time.Time, second time.Time) time.Time {
+	if first.After(second) {
+		return first
+	} else {
```
rjnagal: optional nit: remove the else to remove a branch.

schema/impl.go
afein: Hostname string is the internal hostname from heapster,
For the Timestamp, please look at my answer at the previous inline comment
Returning pointers is NOT safe, the interface for getters is planned to change to return copies. I will make that change on the same PR as the unit tests of the getters.

For everything else, comments incoming on the next commit

schema/impl.go
vishh: +1. This needs to be part of the cluster object. Is the intention to not return the lock as well when Cluster object is returned via APIs? If so, we need an internal and an external object. How about having an internal object that wraps the external object and adds a lock?

vishh:
@rjnagal: Assigned this PR to you!

schema/impl.go
afein: @vishh @rjnagal The lock was moved outside the internal cluster object in commit https://github.com/afein/heapster/commit/32b54bd90a9f4a731ed5f449f9ca617e098a70eb , after  comment https://github.com/GoogleCloudPlatform/heapster/pull/358#discussion_r32685465 and the followup offline discussion.

Due to a dependency with the upcoming [Schema 3] PR, the lock will be re-inserted into the realCluster as part of [Schema 3].

schema/impl.go
afein: As mentioned offline, this check represents only the case where GetAllPodData is called before any data has been inserted in the cluster.

rjnagal:
Let's merge this now, assuming we'll fix all comments in the followup PRs.

PR #360, opened 2015-06-19
[Schema 3] Adders and Unit Tests

afein:
Part 3 of #342 
Includes the following:
- Add methods for Namespaces, Nodes and Pods
- Unit tests for adders and the existing util functions

Changes based on comments from #358:
- lock moved inside the cluster object again
- Get methods have more verbose comments, clearly explaining inputs and outputs

@vishh @rjnagal the next PR, Schema 4, will include utility functions that operate on TimeStores, along with their unit tests.

### DIFF
--- a/schema/impl.go
+++ b/schema/impl.go
@@ -167,7 +167,7 @@ func (rc *realCluster) addPod(pod_name string, pod_uid string, namespace *Namesp
 	if in_ns && in_node {
 		// Pod already in Namespace and Node maps, return pointer
 		pod_ptr, _ = node.Pods[pod_name]
-	} else if !in_ns && !in_node {
+	} else {
 		// Create new Pod and point from node and namespace
 		pod_ptr = &PodInfo{
 			InfoType:   newInfoType(nil, nil),
--- a/schema/impl.go
+++ b/schema/impl.go
@@ -16,13 +16,10 @@ package schema
 
 import (
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/heapster/sinks/cache"
 )
-
-var lock sync.RWMutex
 
 func NewCluster() Cluster {
 	return newRealCluster()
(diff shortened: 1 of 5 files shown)


schema/util.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/GoogleCloudPlatform/heapster/store"
 )
 
+// maxTimestamp returns its largest time.Time argument
```
vishh: latest instead of max may be?

schema/impl.go
```diff
@@ -41,19 +38,23 @@ func newRealCluster() *realCluster {
 	return cluster
 }
 
+// GetAllClusterData returns a pointer to the ClusterInfo, along with all of its metrics.
+// GetAllClusterData also returns the latest cluster timestamp, for reuse in GetNew* methods.
 func (rc *realCluster) GetAllClusterData() (*ClusterInfo, time.Time, error) {
```
vishh: Change to GetClusterData(start, end time.Time) instead?

schema/impl.go
```diff
@@ -91,3 +95,88 @@ func (rc *realCluster) Update(c *cache.Cache) error {
...
+	}
+
+	if _, ok := node.Pods[pod_name]; ok {
+		in_node = true
+	}
+
+	if in_ns && in_node {
+		// Pod already in Namespace and Node maps, return pointer
+		pod_ptr, _ = node.Pods[pod_name]
+	} else if !in_ns && !in_node {
```
vishh: Can we remove the 'if' condition here?

vishh:
Just a few comments. Overall LGTM

schema/impl.go
afein: Interface-level methods will be redesigned alongside the REST endpoints.
The existing Getters serve mostly as proof-of-concept endpoints. After the redesign, these methods will probably be renamed to internal methods used only for testing.
Estimate: [Schema 6]

Fundamental ideas:
- Different methods for historical and derived data
- (start, end) parameters for historical data
- Getters return (time, error) and accept a callback function which performs the appropriate conversion and copy of the internal types to external types.

schema/util.go
afein: done

schema/impl.go
afein: done

schema/impl.go
vishh: Ack.

vishh:
LGTM.

PR #365, opened 2015-06-22
[Schema 4] Timestore-related functionality and unit tests

afein:
Part 4 of #342 
This PR addresses the population of an InfoType element from a cache.ContainerElement.
This involves all the TimeStore-related functionality of the implementation.

Additions (at 100% test coverage):
- updateInfoType: populates an InfoType from a ContainerElement
- parseMetric: populates a map[string]*TimeStore from a ContainerMetricElement
- addMetricToMap: adds a single metric to a map[string]*TimeStore

Fundamental change - Feedback required:
 realCluster now contains a tsConstructor: func() (store.TimeStore) field, meant to be populated by the caller. 
+ This effectively decouples the TimeStore implementation and its duration from the realCluster.
- The three new functions (updateInfoType, parseMetric, addMetricToMap) need to be realCluster methods, in order to have access to tsConstructor.

cc @rjnagal @vishh , please do not be discouraged by the PR size, 70% of the additions are unit tests

### DIFF
--- a/schema/impl.go
+++ b/schema/impl.go
@@ -18,22 +18,28 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/golang/glog"
+
 	"github.com/GoogleCloudPlatform/heapster/sinks/cache"
+	"github.com/GoogleCloudPlatform/heapster/store"
 )
 
-func NewCluster() Cluster {
-	return newRealCluster()
+// NewCluster returns a new Cluster, given a TimeStore constructor function.
+func NewCluster(tsConstructor func() store.TimeStore) Cluster {
+	return newRealCluster(tsConstructor)
 }
 
-func newRealCluster() *realCluster {
+// newRealCluster returns a realCluster, given a TimeStore constructor.
+func newRealCluster(tsConstructor func() store.TimeStore) *realCluster {
 	cinfo := ClusterInfo{
 		InfoType:   newInfoType(nil, nil),
 		Namespaces: make(map[string]*NamespaceInfo),
 		Nodes:      make(map[string]*NodeInfo),
 	}
 	cluster := &realCluster{
-		timestamp:   time.Time{},
-		ClusterInfo: cinfo,
+		timestamp:     time.Time{},
+		ClusterInfo:   cinfo,
+		tsConstructor: tsConstructor,
 	}
 	return cluster
 }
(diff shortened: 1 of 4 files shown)


schema/impl.go
```diff
@@ -180,3 +196,123 @@ func (rc *realCluster) addPod(pod_name string, pod_uid string, namespace *Namesp
...
+	}
+	if dict == nil {
+		return zeroTime, fmt.Errorf("cannot populate nil map")
+	}
+
+	timestamp := cme.Stats.Timestamp
+	if cme.Spec.HasCpu {
+		// Add CPU Limit metric
+		cpu_limit := cme.Spec.Cpu.Limit
+		err := rc.addMetricToMap("cpu/limit", timestamp, cpu_limit, dict)
```
vishh: nit: Can we have consts for resource types and "limits" or "usage" or "%tage"?

vishh: Better to have sub-structure, which will help with derived stats. Store spec and stats separately in a composite struct.

vishh:
Just one comment. Otherwise LGTM

schema/impl.go
afein: constants nit addressed in the latest commit.

I'm already planning to change the structure of the existing InfoType object at [Schema 8], to facilitate derived stats. Let's discuss offline and postpone structural concerns related to derived stats for a future PR.

afein:
@rjnagal @vishh ready to merge

Regarding the comment, please look at my line note reply: https://github.com/GoogleCloudPlatform/heapster/pull/365#discussion_r33193039

schema/types.go
```diff
@@ -77,3 +78,12 @@ type PodInfo struct {
 type ContainerInfo struct {
 	InfoType
 }
+
+// Supported metric names, used as keys for all map[string]*store.TimeStore
+const cpuLimit = "cpu/limit"
```
vishh: How about a generic resource struct { usage int, limit int } and have one such struct for every resource type? Essentially we will have a static struct with all the metrics defined. Ok to tackle in a future PR.

afein: That is the eventual goal, however there are a few issues to consider, such as:
- time resolution of limits vs time resolution of usage.
- mismatch of limit/usage timestamps.
- input resolution versus output resolution for historical data.
- instantaneous derived stats (e.g. current utilization percentage) and the required conversions of limits/usage values.

For those reasons, I believe it's better if we address the format of the resource struct together with the design of the derived stats.

vishh: All your points are valid! One option is to have a map of resource type to
a timeseries of resource metrics (spec + stats). This is closer to the
kubernetes model.
As long as this is not what we expose via APIs, I am totally fine deferring
the refactoring to a later stage.

PR #366, opened 2015-06-23
Added possibility to export node only metrics for GCM.

piosz:
This is for Kubernetes cluster size autoscaling purposes due to limits of
custom metrics. More details https://cloud.google.com/monitoring/quota-policy.


piosz:
cc @vishh

vishh:
ok to test

vishh:
Can we drop this in favor of a autoscaling specific sink?

piosz:
Yes.

PR #371, opened 2015-06-25
Add option to use in cluster config

mikedanese:
extracted all kubeconfig creation logic to makeKubeConfig and added a return on inClusterConfig option at the top of makeKubeConfig.

```
➜  cluster-monitoring git:(master) ✗ cat influxdb/heapster-controller.yaml 
apiVersion: v1
kind: ReplicationController
metadata:
  name: monitoring-heapster-v4
  namespace: default 
  labels: 
    k8s-app: heapster
    version: v4
    kubernetes.io/cluster-service: "true"
spec: 
  replicas: 1
  selector: 
    k8s-app: heapster
    version: v4
  template: 
    metadata: 
      labels: 
        k8s-app: heapster
        version: v4
        kubernetes.io/cluster-service: "true"
    spec: 
      containers: 
        - image: gcr.io/mikedanese-k8s/heapster:test
          name: heapster
          resources:
            limits:
              cpu: 200m
              memory: 2.5Gi
          command:
            - /heapster
            - --source=kubernetes:''?inClusterConfig=true
            - --sink=influxdb:http://monitoring-influxdb:8086
➜  cluster-monitoring git:(master) ✗ k logs monitoring-heapster-v4-gs5n6             
I0625 21:34:39.622669       1 heapster.go:52] /heapster --source=kubernetes:''?inClusterConfig=true --sink=influxdb:http://monitoring-influxdb:8086
I0625 21:34:39.622751       1 heapster.go:53] Heapster version test
I0625 21:34:39.623075       1 kube_factory.go:104] using in cluster config
I0625 21:34:39.623315       1 kube_factory.go:75] Using Kubernetes client with master "https://10.0.0.1:443" and version "v1beta3"
I0625 21:34:39.623342       1 kube_factory.go:76] Using kubelet port 10255
W0625 21:34:39.624930       1 request.go:308] field selector: v1beta3 - pods - spec.host - : need to check if this is versioned correctly.
I0625 21:34:39.630766       1 driver.go:376] created influxdb sink with options: {root root monitoring-influxdb:8086 k8s false}
I0625 21:34:39.631296       1 heapster.go:64] Starting heapster on port 8082
W0625 21:34:39.770255       1 request.go:308] field selector: v1beta3 - pods - spec.host - : need to check if this is versioned correctly.
```


mikedanese:
@vishh

vishh:
ok to test

sources/kube_factory.go
```diff
@@ -76,7 +76,18 @@ func getConfigOverrides(uri string, options map[string][]string) (*kubeClientCmd
...
+		if err != nil {
+			return nil, fmt.Errorf("error creating inClusterConfig: %v", err)
+		}
+		glog.Infof("using in cluster config")
+		return kubeConfig, nil
+	}
+
 	configOverrides, err := getConfigOverrides(uri, options)
 
 	authFile := defaultKubeConfigFile
```
vishh: As of now, heapster expects an auth file at `defaultKubeConfigFile` path to exist. Is that required moving forward with service accounts? Can we make auth an optional item - user cluster without auth, can run vanilla heapster without any issues?

mikedanese: Yes. I left the rest of makeKubeConfig in for backwards compatibility, but eventually we should be able to replace all of it with kube_client.InClusterConfig() or a kubeconfig file without the extra decoration that makeKubeConfig() and getConfigOverrides() currently does.

vishh: My understanding is that service accounts are created by default on all clusters. If thats true, I think breaking backwards compatibility is fine, since the new model will work off the box.
Essentially I am suggesting getting rid of the `auth` config completely, since service accounts do the right thing.

mikedanese: Ok. I can do that. Service accounts are on by default but the tls crt is only distributed if the optional `--root-ca-cert` flag is set on the kube-controller-manager. Is that okay?

mikedanese:
@vishh I've made requested changes and ran `make test-unit` and `make test-integration`. Is there anything else the jenkins config is doing?

PR #372, opened 2015-06-25
Make the system container names user friendly

vishh:
Essentially this PR removes the "/" prefix from system containers. Over time, we will have to update this logic to handle system containers in systemd systems.
The e2e test has been upgraded to capture the presence of system containers on all the hosts.


vishh:
cc @vmarmol @a-robinson

vmarmol:
LGTM

PR #378, opened 2015-06-26
Cumulative Moving Average TimeStore

afein:
Added a CMA TimeStore implementation, cma_store. 

All TimePoints in the cma_store have unique timestamps. When two or more TimePoints with the same timestamp are Put in the store, they are converted to a single TimePoint with a Value that represents the average of all TimePoints with the same timestamp.

cma_store can be used as the basis for a TimeStore that can maintain derived stats, as it is already making the assumption that Value is a uint64.

100% Test Coverage

cc @rjnagal @vishh

### DIFF
--- a/store/cma_store.go
+++ b/store/cma_store.go
@@ -1,0 +1,141 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package store
+
+import (
+	"container/list"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// cmaStore maintains the cumulative moving average of values at a specific Timestamp.
+// cmaStore is an implementation of TimeStore where the values are casted as uint64.
+// TODO(alex): avoid duplication with in_memory.
+type cmaStore struct {
+	// buffer is a list of tpContainers that is sequenced in a time-descending order.
+	buffer *list.List
+	// rwLock protects all operations on the buffer.
+	rwLock sync.RWMutex
+}
+
+// tpContainer is the actual struct that is being stored in the buffer that implements cmaStore.
+// tpContainer contains a TimePoint and the count of TimePoints with the same timestamp that
+// have been averaged to a single tpContainer. The TimePoint.Value field contains the average
+// of all these TimePoints.
+type tpContainer struct {
+	TimePoint
+	count int
+}
+
+func (ts *cmaStore) Put(tp TimePoint) error {
+	if tp.Value == nil {
+		return fmt.Errorf("cannot store TimePoint with nil data")
+	}
+	if (tp.Timestamp == time.Time{}) {
+		return fmt.Errorf("cannot store TimePoint with zero timestamp")
+	}
+	ts.rwLock.Lock()
+	defer ts.rwLock.Unlock()
+	newTPC := tpContainer{
+		TimePoint: tp,
+		count:     1,
+	}
+	if ts.buffer.Len() == 0 {
+		glog.V(5).Infof("put pushfront: %v, %v", tp.Timestamp, tp.Value)
+		ts.buffer.PushFront(newTPC)
+		return nil
+	}
+	for elem := ts.buffer.Front(); elem != nil; elem = elem.Next() {
+		curr := elem.Value.(tpContainer)
+		if tp.Timestamp.Equal(curr.Timestamp) {
+			// If an element with that timestamp exists, update its average and count
+			newVal := tp.Value.(uint64)
+			n := uint64(curr.count)
+			oldAvg := curr.Value.(uint64)
+			curr.Value = uint64((newVal + (n * oldAvg)) / (n + 1))
+			curr.count = curr.count + 1
+			elem.Value = curr
+			return nil
+		} else if tp.Timestamp.After(curr.Timestamp) {
+			glog.V(5).Infof("put insert before: %v, %v, %v", elem, tp.Timestamp, tp.Value)
+			ts.buffer.InsertBefore(newTPC, elem)
+			return nil
+		}
+	}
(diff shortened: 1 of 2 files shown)


store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// cmaStore maintains the cumulative moving average of values at a specific Timestamp.
+// cmaStore is an implementation of TimeStore where the values are casted as uint64.
+type cmaStore struct {
+	buffer *list.List
```
vmarmol: nit: document these please

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+	rwLock sync.RWMutex
+}
+
+type TPContainer struct {
```
vmarmol: Any particular reason this one is public?

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+	TimePoint
```
vmarmol: document here too :)

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+	count uint16
```
vmarmol: I'd say to just use an int, it is generally simpler and we don't need the optimization

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+func (ts *cmaStore) Put(tp TimePoint) error {
+	if tp.Value == nil {
+		return fmt.Errorf("cannot store TimePoint with nil data")
+	}
+	if (tp.Timestamp == time.Time{}) {
+		return fmt.Errorf("cannot store TimePoint with zero timestamp")
+	}
+	ts.rwLock.Lock()
+	defer ts.rwLock.Unlock()
+	new_tpc := TPContainer{
```
vmarmol: Go uses camelCase, here and elsewhere

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+		TimePoint: tp,
+		count:     uint16(1),
+	}
+	if ts.buffer.Len() == 0 {
+		glog.V(5).Infof("put pushfront: %v, %v", tp.Timestamp, tp.Value)
+		ts.buffer.PushFront(new_tpc)
+		return nil
+	}
+	for elem := ts.buffer.Front(); elem != nil; elem = elem.Next() {
+		tpc := elem.Value.(TPContainer)
```
vmarmol: maybe rename to current? tp vs tpc gets hard for readability

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		// Add all entries whose timestamp is not after end.
+		if (end == time.Time{}) || !entry.Timestamp.After(end) {
+			result = append(result, entry)
+		}
+	}
+	return result
+}
+
+func (ts *cmaStore) Delete(start, end time.Time) error {
```
vmarmol: Not a huge fan of the duplication between here and `in_memory.go`. If we find a bug in one we'll have to update both and I'm sure we'd forget to do so. We should try to merge the implementations more, but we don't have to do that now. Can we add a TODO to look into merging them?

store/cma_store.go
```diff
@@ -0,0 +1,135 @@
...
+		if !tpc.Timestamp.Before(start) {
+			ts.buffer.Remove(oldElem)
+		}
+	}
+	return nil
+}
+
+func NewCMAStore() TimeStore {
+	return &cmaStore{
+		rwLock: sync.RWMutex{},
```
vmarmol: nit: you shouldn't need to do this

store/cma_store_test.go
```diff
@@ -0,0 +1,138 @@
...
+	// Get Invocation with values 2 and 4
+	actual := store.Get(zeroTime, now)
+	fmt.Println(actual)
+	require.Len(t, actual, 1)
+	assert.Equal(actual[0].Value, uint64(4))
+
+	// Get Invocation with values 0, 20, 10 and 1000
+	actual = store.Get(zeroTime, zeroTime)
+	require.Len(t, actual, 2)
+	assert.Equal(actual[0].Value, uint64(257)) // 257.5 rounded down
```
vmarmol: How about we use a value that averages out to a round number? Easier to reason about and less comment :) {0, 30, 10, 1000} = 1040 / 4 = 260. / 8 = 130

vmarmol:
Looks good overall, just minor comments and suggestions.

afein:
addressed most comments, squashed with previous commit. Please view line note answers for nits that were not addressed in the new commit.

store/cma_store_test.go
afein: The problem in that case is that (0+30+10)/3 is not round, as well as 1040/5, 1040/6 etc. As this is a cumulative average with unsigned values, the number of testcases that will actually match the exact average of all values is very limited, due to consecutive rounding for each individual average calculation. Even if we find such a testcase, that behavior is part of the the uint64 implementation and should require testing.

store/cma_store.go
afein: in_memory.go is used only for the cache implementation, it is barely documented and it is not as well-tested as cma_store. As the cache might (or might not) be eventually deprecated, in_memory might be removed as well, or just replaced by cma_store. Duplication was used now to avoid taking such deprecation decisions. 

I expect that the functionality of in_memory.go, if not deprecated, can be implemented within cma_store with a simple configuration flag and a conditional statement.

store/cma_store_test.go
vmarmol: Ah I see what you mean. Long term, we may want to look at how this affects the accuracy of the data. SGTM for now.

store/cma_store.go
vmarmol: Agree, but we should make a note of this. Many times these "temporary" things outlive us all.

vmarmol: Saw the TODO you added, looks good!

store/cma_store.go
```diff
@@ -0,0 +1,141 @@
...
+	if ts.buffer.Len() == 0 {
+		glog.V(5).Infof("put pushfront: %v, %v", tp.Timestamp, tp.Value)
+		ts.buffer.PushFront(newTPC)
+		return nil
+	}
+	for elem := ts.buffer.Front(); elem != nil; elem = elem.Next() {
+		curr := elem.Value.(tpContainer)
+		if tp.Timestamp.Equal(curr.Timestamp) {
+			// If an element with that timestamp exists, update its average and count
+			new_val := tp.Value.(uint64)
```
vmarmol: nit: camelCase (and in `old_avg`)

vmarmol:
One nit, LGTM otherwise

afein:
ready to merge, merge build fail is irrelevant to this PR

vmarmol:
Thanks @afein! Merging.

PR #383, opened 2015-07-01
Fix Kubernetes Grafana Default Dashboard

saad-ali:
Currently the default Kubernetes Grafana dashboard does not show metrics correctly.
With this fix the dashboard looks like this:
defaultkubernetesgrafanadashboard


saad-ali:
CC @vishh @vmarmol @rjnagal

vmarmol:
LGTM, thanks @saad-ali!

PR #384, opened 2015-07-03
Added exporting nodes reservation metrics for Autoscaling sink

piosz:
cc @vmarmol @vishh @jszczepkowski

### DIFF
--- a/sinks/gcmautoscaling/driver.go
+++ b/sinks/gcmautoscaling/driver.go
@@ -15,6 +15,8 @@
 package gcmautoscaling
 
 import (
+	"time"
+
 	"github.com/GoogleCloudPlatform/heapster/extpoints"
 	"github.com/GoogleCloudPlatform/heapster/sinks/gcm"
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


vmarmol:
ok to test

sinks/gcmautoscaling/driver.go
```diff
@@ -52,11 +52,19 @@ type utilizationMetric struct {
 var autoscalingMetrics = map[string]utilizationMetric{
 	"cpu/usage": {
 		name:        "cpu/node_utilization",
-		description: "Cpu utilization as a percentage of node capacity",
+		description: "Cpu utilization as a share of node capacity",
```
vmarmol: What do you mean by shares? What units?

piosz: I mean total usage divided by node capacity. It's the number from range [0, 1] without any unit.

vmarmol: Just curious about the units :)

sinks/gcmautoscaling/driver.go
```diff
@@ -88,29 +108,90 @@ func isNode(metric *sink_api.Point) bool {
...
+		LabelGCEResourceType.Key: "instance",
+	}
+}
+
+func (self *gcmAutocalingSink) getNewValue(metric *sink_api.Point, ts *gcm.Timeseries) *float64 {
+	host := metric.Labels[sink_api.LabelHostname.Key]
+
+	var val float64
+	switch metric.Name {
+	case "cpu/usage":
```
vmarmol: Seems like we should start using defined values for these strings. We can do in another PR if we'd like.

sinks/gcmautoscaling/driver.go
```diff
@@ -88,29 +108,90 @@ func isNode(metric *sink_api.Point) bool {
...
+			if metric.Name == "cpu/limit" {
+				self.cpuReservation[hostTime{host, metric.End}] += value
+			} else if metric.Name == "memory/limit" {
+				self.memReservation[hostTime{host, metric.End}] += value
+			}
+		}
+	}
+}
+
+func getLabels(metric *sink_api.Point) map[string]string {
```
vmarmol: Can we document this function and the one below it?

vmarmol:
LGTM, sorry for the delay. Just minor edits. Will merge to unblock you, please send another PR for the fixes :)

piosz:
No problem. Thanks for review! Added follow up PR.

PR #283, opened 2015-05-07
Update Kubernetes deployment configs

saad-ali:
To match configs currently checked in to Kubernetes HEAD

CC @vishh

PR #387, opened 2015-07-06
Migrating to Kubernets API v1

krousey:
Kubernetes is going to stop exposing the v1beta3 API by default, so this need to switch to the new version.

See https://github.com/GoogleCloudPlatform/kubernetes/issues/10755

### DIFF
--- a/deploy/kube-config/google/heapster-service.json
+++ b/deploy/kube-config/google/heapster-service.json
@@ -1,6 +1,6 @@
 {
   "kind": "Service",
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "metadata": {
     "name": "monitoring-heapster",
     "labels": {
--- a/deploy/kube-config/influxdb/grafana-service.json
+++ b/deploy/kube-config/influxdb/grafana-service.json
@@ -1,5 +1,5 @@
 {
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "kind": "Service",
   "metadata": {
     "labels": {
--- a/deploy/kube-config/influxdb/heapster-service.json
+++ b/deploy/kube-config/influxdb/heapster-service.json
@@ -1,5 +1,5 @@
 {
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "kind": "Service",
   "metadata": {
     "labels": {
--- a/deploy/kube-config/influxdb/influxdb-grafana-controller.json
+++ b/deploy/kube-config/influxdb/influxdb-grafana-controller.json
@@ -1,5 +1,5 @@
 {
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "kind": "ReplicationController",
   "metadata": {
     "labels": {
@@ -38,7 +38,7 @@
             "env": [
               {
                 "name": "INFLUXDB_EXTERNAL_URL",
-                "value": "/api/v1beta3/proxy/namespaces/default/services/monitoring-grafana/db/"
+                "value": "/api/v1/proxy/namespaces/default/services/monitoring-grafana/db/"
               },
               {
                 "name": "INFLUXDB_HOST",
--- a/deploy/kube-config/influxdb/influxdb-service.json
+++ b/deploy/kube-config/influxdb/influxdb-service.json
@@ -1,5 +1,5 @@
 {
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "kind": "Service",
   "metadata": {
     "labels": {
--- a/deploy/kube-config/influxdb/influxdb-ui-service.json
+++ b/deploy/kube-config/influxdb/influxdb-ui-service.json
@@ -1,5 +1,5 @@
 {
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "kind": "Service",
   "metadata": {
     "labels": null,
--- a/deploy/kube-config/standalone/heapster-service.json
+++ b/deploy/kube-config/standalone/heapster-service.json
@@ -1,5 +1,5 @@
 {
-  "apiVersion": "v1beta3",
+  "apiVersion": "v1",
   "kind": "Service",
   "metadata": {
     "labels": {
--- a/docs/influxdb.md
+++ b/docs/influxdb.md
@@ -10,7 +10,7 @@ _Warning: Virtual machines need to have at least 2 cores for InfluxDB to perform
 $ kubectl.sh create -f deploy/kube-config/influxdb/
 ```
 
-Grafana will be accessible at `https://<masterIP>/api/v1beta3/proxy/namespaces/default/services/monitoring-grafana/`. Use the master auth to access Grafana.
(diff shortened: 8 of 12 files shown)


sinks/gcl/driver_test.go
```diff
@@ -127,7 +127,7 @@ func TestStoreEventsSingleEventInput(t *testing.T) {
 	assert.Equal(t, "POST" /* expected */, fakeSink.fakeHttpClient.capturedSendRequests[0].Method /* actual */)
 	assert.Equal(t, "https" /* expected */, fakeSink.fakeHttpClient.capturedSendRequests[0].URL.Scheme /* actual */)
 	assert.Equal(t, "logging.googleapis.com" /* expected */, fakeSink.fakeHttpClient.capturedSendRequests[0].URL.Host /* actual */)
-	assert.Equal(t, "/v1beta3/projects/fakeProjectId/logs/kubernetes.io%2Fevents/entries:write" /* expected */, fakeSink.fakeHttpClient.capturedSendRequests[0].URL.Opaque /* actual */)
+	assert.Equal(t, "/v1/projects/fakeProjectId/logs/kubernetes.io%2Fevents/entries:write" /* expected */, fakeSink.fakeHttpClient.capturedSendRequests[0].URL.Opaque /* actual */)
```
bgrant0607: This is not our v1beta3 API.

sinks/gcl/driver_test.go
```diff
@@ -172,7 +172,7 @@ func TestStoreEventsMultipleEventInput(t *testing.T) {
...
```
bgrant0607: Ditto

rjnagal:
ok to test

sinks/gcl/driver_test.go
bgrant0607: See:
https://cloud.google.com/logging/docs/api/ref/rest/index

bgrant0607:
Note: We should still update them, but we don't run using the json files here. There was an earlier unmerged (and now out-of-date) PR #283 to re-sync the entire files with Kubernetes.

sinks/gcl/driver_test.go
bgrant0607: Please revert this file.

sinks/gcl/driver.go
```diff
@@ -39,7 +39,7 @@ const (
 	eventLoggingLogName      = "kubernetes.io/events"
 	logEntriesWriteURLScheme = "https"
 	logEntriesWriteURLHost   = "logging.googleapis.com"
-	logEntriesWriteURLFormat = "/v1beta3/projects/%s/logs/%s/entries:write"
+	logEntriesWriteURLFormat = "/v1/projects/%s/logs/%s/entries:write"
```
bgrant0607: Please revert this file.

sinks/gcl/driver_test.go
krousey: Done

sinks/gcl/driver.go
krousey: Done.

bgrant0607:
LGTM

dchen1107:
LGTM

vmarmol:
LGTM, @krousey have you had a chance to try this out on a Kubernetes cluster manually? Our integration setup is a bit broken at the moment...

krousey:
@vmarmol can you refresh my memory on how?

krousey:
@vmarmol I'm running some Kubernetes e2e tests against krousey/heapster:canary right now. Hopefully this is what you wanted? Also it seems the Jenkins build timed out on the integration tests. It ran fine locally.

bgrant0607:
Oh, you need to update Kubernetes packages in godeps

krousey:
@bgrant0607 I don't understand. Doesn't this change explicitly ask for the v1 API?

krousey:
Or are we defaulting to whatever kubernetes wants somewhere?

bgrant0607:
Only in the integration test.

krousey:
@bgrant0607 it seems weird to have heapster not tied to a concrete version. Shouldn't it be?

As far as updating Kubernetes, I have heapster in a separate GOPATH. Should I clone kubernetes directly, sync to tag v.20.0 and godep restore kubernetes then godep save heapster?

bgrant0607:
Heapster looks like it attempted to specify v1 already:
https://github.com/GoogleCloudPlatform/heapster/blob/master/sources/kube_factory.go#L35

We need to understand why that didn't work.

cc @nikhiljindal 

I'm not an expert on godeps, unfortunately. Perhaps @lavalamp could provide some advice. 

As for what version of Kubernetes -- I believe there were fixes more recent than v.20.0 that we need. I'd grab head.

bgrant0607:
Heapster looks like it should be logging which api version it thinks it's using:
https://github.com/GoogleCloudPlatform/heapster/blob/master/sources/kube_factory.go#L166

nikhiljindal:
I could be wrong, but I believe you can do:
```
$ cd <heapster_dir>
$ godep restore
$ go get -u github.com/GoogleCloudPlatform/kubernetes
$ godep update github.com/GoogleCloudPlatform/kubernetes
```
This should update https://github.com/GoogleCloudPlatform/heapster/blob/master/Godeps/Godeps.json#L34

https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/devel/development.md#using-godep has more details

lavalamp:
What @nikhiljindal says is correct, but you need to make a dedicated GOPATH first, or you risk cross-contamination.

krousey:
I think I'm giving up on trying to get this working tonight. I will be in tomorrow to ask for help, because what I did isn't working.

krousey:
@vmarmol I added a change to make heapster use the v1 API and ignore the preferred version in the client in the godeps version of the client. Is this ok?

krousey:
Also https://github.com/GoogleCloudPlatform/kubernetes/pull/10785 tests the kubernetes e2e with a docker image I built from this. It passed and did not use the v1beta3 API.

nikhiljindal:
Awesome Thanks @krousey !!

bgrant0607:
cc @jlowdermilk, in case he can help with the kubeconfig in any way

krousey:
Integration also passed locally.

rjnagal:
LGTM 

I think this needs to wait for #389 though

krousey:
@rjnagal This does not depend on that. The second commit in this PR forces v1.

jlowdermilk:
LGTM. Regarding using config overrides to explicitly set apiversion, it's fine medium term solution, but should put a TODO in to remove it after updating client deps.

rjnagal:
Thanks @krousey, merging.

krousey:
@jlowdermilk Why should this be removed? Heapster is a client of the API and should only use versions it supports. In fact, this should go in before #389 because it would break if heapster defaulted to v1 without switching it's node query label from spec.host to spec.nodeName. I think it should use the version it was coded to use.

jlowdermilk:
@krousey, I stand corrected. If it's using/referring to api objects directly then explictly setting apiversion in client config is correct.

PR #388, opened 2015-07-07
[Schema 6] Metric Aggregation

afein:
Part 6 of #342 
This is #380, reopened due to ghost commits in the old PR.

This PR includes a full implementation of metric aggregation (summation) throughout the cluster structure.

New realCluster methods:
- aggregationStep: performs full aggregation of all entities in the structure, holds the cluster lock.
- aggregateNodeMetrics: sums Node metrics to generate Cluster metrics
- aggregateNamespaceMetrics: sums Pod metrics to generate Namespace metrics
- aggregatePodMetrics: sums Container metrics to generate Pod metrics
- aggregateMetrics: populates an InfoType by summing values from an []*InfoType

New utility functions:
- addTimePoints: adds the values of two TimePoints, returns a new TimePoint with the added value.
- popTPSlice: pops the first element of a []TimePoint.
- addMatchingTimeseries: adds two []TimePoint with matching timestamps, returns a new []TImePoint.

Misc updates:
- Removed the existing Get method from the implementation to avoid confusion. Getters are being redesigned at [Schema 7]

All methods that are not at 100% coverage contain internal error handling branches.

cc @rjnagal  @vmarmol @vishh

PR #394, opened 2015-07-08
Added aligning stats timestamps to multiple of resolution

piosz:
This is required to add a possibility of aggregating stats for node/cluster
needed for autoscaling purposes.

### DIFF
--- a/heapster.go
+++ b/heapster.go
@@ -35,6 +35,7 @@ import (
 var (
 	argPollDuration    = flag.Duration("poll_duration", 10*time.Second, "The frequency at which heapster will poll for stats")
 	argStatsResolution = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats. Acceptable values are in the range [1 second, 'poll_duration')")
+	argAlignStats      = flag.Bool("align_stats", false, "Whether to align timestamps of metrics to multiplicity of 'stats_resolution'")
 	argPort            = flag.Int("port", 8082, "port to listen to")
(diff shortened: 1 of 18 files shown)


piosz:
@vmarmol, @vishh are you ok with such change? If yes I'll add unit test.

vmarmol:
@afein you were working on something similar no?

afein:
Hey @piosz, indeed I am working on something very similar which involves both stat alignment and aggregation. 

Feel free to take a look at how aggregation and alignment is implemented in #388 and let us know if that satisfies your use cases. After that PR is merged, the next PR involves REST endpoints for all metrics.

Here are some example REST endpoints:

/api/v1/model/cluster/cpu-usage
/api/v1/model/pod/monitoring-heapster-xk12d/memory-usage
/api/v1/model/namespace/default/cpu-limit

The JSON response for these endpoints would be a single timeseries, represented as a list of timestamp-value pairs. These endpoints also have an optional starting time parameter, so you can only get metrics that are newer than your specified time.

You can take a look at the core of the next PR here:
https://github.com/GoogleCloudPlatform/heapster/compare/master...afein:schema7

Let me know if that model would work in your use case, in which case we can speed up the merge process to let you access that functionality faster!

piosz:
@afein maybe you can review this PR?

afein:
LGTM, but @vmarmol should review it as well

vmarmol:
LGTM

piosz:
Thanks!

PR #395, opened 2015-07-08
Properly override host as passed by --source

mvdan:


mvdan:
Maybe I should have clarified in the commit message that this is for the `kubernetes` source.

sources/kube_factory.go
```diff
@@ -102,6 +102,7 @@ func CreateKubeSources(uri string, options map[string][]string) ([]api.Source, e
 			return nil, err
 		}
 
+		kubeConfig.Host = configOverrides.ClusterInfo.Server
```
vmarmol: This should probably only override if it was provided no?

mvdan: Fair point.

mvdan:
@vmarmol just updated the patch with an empty string check. Made sure both the default and overriding cases work.

vmarmol:
LGTM, thanks for being thorough @mvdan!

PR #396, opened 2015-07-08
Fix typo in Heapster InfluxDB Container creation command

saad-ali:
CC @vmarmol @vishh


vmarmol:
LGTM

vmarmol:
Thanks @saad-ali!

PR #397, opened 2015-07-09
Fixed time range for collecting stats for every pass

piosz:
This should be merged and then pushed to Kubernetes ASAP because it may cause:
* exporting duplicated data
* increasing cpu/mem usage by Heapster due to increasing amount of data to process

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -80,7 +80,7 @@ func (rm *realManager) Housekeep() {
 	var sd syncData
 	start := rm.lastSync
 	end := time.Now()
-	rm.lastSync = start
+	rm.lastSync = end
 	glog.V(2).Infof("starting to scrape data from sources")
 	for idx := range rm.sources {
 		s := rm.sources[idx]


piosz:
@vmarmol

vmarmol:
I know @afein had some work in this area as well

vmarmol:
Ah no, disregard that @afein :)

vmarmol:
ok to test

vmarmol:
Hmmm this code is 2 months old, what issues have you found from it? Or was it just by inspection?

vmarmol:
Ah I see, this will eventually get to cAdvisor which only keeps ~2min of data anyways. This should be safe. Have you had a chance to manually test?

vmarmol:
we shouldn't duplicate the exported data (we remove it before export), but it is using more resources than necessary to get this information.

piosz:
I've found this bug while developing #394. There seems to be no issue probably because of what you wrote about cAdvisor, but it always requests stats from the interval begin on the time of Heapster creation.

What do you mean by manually testing? I ran Heapster with this change and seems to work but I may perform more advances tests tomorrow.

vmarmol:
Thanks @piosz! That's what I meant. We'll merge now and try to get into Kubernetes. We'll need extra testing for that anyways.

PR #398, opened 2015-07-09
Add support for modifying sinks at runtime

mvdan:
Added a new API endpoint, /api/v1/sinks, that returns the sink uris on GET and
lets you set them via a POST. The sink uris work just like when you pass them
as flags.

Had to split the ExternalSinkManager interface into Sink for actual sinks and
ExternalSinkManager for the actual managers (only sinks/external.go for now).
The latter adds a new method, SetSinks(), which is what we use here.

Since sinks may now be removed at runtime, an Unregister() function in the
ExternalSink interface was added. The implementations have been left as TODOs
for now - what each one will do depends on the implementation, just like
Regsiter().

CC @vmarmol @vishh

### DIFF
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -59,6 +59,21 @@ func (a *Api) Register(container *restful.Container) {
 		Operation("exportmetricsSchema").
 		Writes(TimeseriesSchema{}))
 	container.Add(ws)
+	ws = new(restful.WebService)
+	ws.Path("/api/v1/sinks").
+		Doc("Configuration for Heapster sinks for exporting data").
+		Produces(restful.MIME_JSON)
+	ws.Route(ws.POST("").
+		To(a.setSinks).
+		Doc("set the current sinks").
+		Operation("setSinks").
+		Reads([]string{}))
+	ws.Route(ws.GET("").
+		To(a.getSinks).
+		Doc("get the current sinks").
+		Operation("getSinks").
+		Writes([]string{}))
+	container.Add(ws)
 }
 
 func compressionFilter(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
@@ -178,4 +193,24 @@ func (a *Api) exportMetrics(request *restful.Request, response *restful.Response
 	}
 
 	response.WriteEntity(timeseries)
+}
+
+func (a *Api) setSinks(req *restful.Request, resp *restful.Response) {
+	sinkUris := new([]string)
+	if err := req.ReadEntity(sinkUris); err != nil {
+		resp.WriteError(http.StatusBadRequest, err)
+		return
+	}
+	if err := a.manager.SetSinkUris(*sinkUris); err != nil {
+		resp.WriteError(http.StatusInternalServerError, err)
+		return
+	}
+}
+
+func (a *Api) getSinks(req *restful.Request, resp *restful.Response) {
+	sinks := a.manager.SinkUris()
+	if sinks == nil {
+		sinks = make([]string, 0)
+	}
+	resp.WriteEntity(sinks)
 }
--- a/docs/sink-configuration.md
+++ b/docs/sink-configuration.md
@@ -6,7 +6,7 @@ via the `--sink` flag. The flag takes an argument of the form `PREFIX:CONFIG[?OP
 Options (optional!) are specified as URL query parameters, separated by `&` as normal.
 This allows each source to have custom configuration passed to it without needing to
 continually add new flags to Heapster as new sinks are added. This also means
-heapster can store data intomultiple sinks at once.
+heapster can store data into multiple sinks at once.
 
 ## Current sinks
 ### InfluxDB
@@ -75,3 +75,9 @@ If `HAWKULAR_SERVER_URL` includes any path, the default `hawkular/metrics` is ov
 The following options are available:
 
 * `tenant` - Hawkular-Metrics tenantId (default: `heapster`)
+
+## Modifying the sinks at runtime
+
+Using the `/api/v1/sinks` endpoint, it is possible to fetch the sinks
+currently in use via a GET request or to change them via a POST request. The
+format is the same as when passed via command line flags.
--- a/manager/sinks.go
+++ b/manager/sinks.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package manager
 
 import (
 	"fmt"
@@ -22,9 +22,9 @@ import (
 	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api/v1"
 )
 
-func newSinks() ([]sink_api.ExternalSink, error) {
+func newSinks(sinkUris []string) ([]sink_api.ExternalSink, error) {
 	var sinks []sink_api.ExternalSink
-	for _, sinkFlag := range argSinks {
+	for _, sinkFlag := range sinkUris {
 		uri, err := url.Parse(sinkFlag)
 		if err != nil {
 			return nil, err
--- a/sinks/external.go
+++ b/sinks/external.go
@@ -17,6 +17,7 @@ package sinks
 import (
 	"fmt"
 	"strings"
+	"sync"
 
 	sink_api_old "github.com/GoogleCloudPlatform/heapster/sinks/api"
 	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api/v1"
@@ -27,11 +28,10 @@ import (
 type externalSinkManager struct {
 	decoder       sink_api_old.Decoder
 	externalSinks []sink_api.ExternalSink
+	sinkMutex     sync.RWMutex // Protects externalSinks
 }
 
-// NewExternalSinkManager returns an external sink manager that will manage pushing data to all
-// the sinks in 'externalSinks', which is a map of sink name to ExternalSink object.
-func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSinkManager, error) {
+func supportedMetricsDescriptors() []sink_api.MetricDescriptor {
 	// Get supported metrics.
 	supportedMetrics := sink_api.SupportedStatMetrics()
 	for i := range supportedMetrics {
@@ -43,21 +43,23 @@ func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSink
 	for _, supported := range supportedMetrics {
 		descriptors = append(descriptors, supported.MetricDescriptor)
 	}
+	return descriptors
+}
 
-	for _, externalSink := range externalSinks {
-		err := externalSink.Register(descriptors)
-		if err != nil {
+// NewExternalSinkManager returns an external sink manager that will manage pushing data to all
+// the sinks in 'externalSinks', which is a map of sink name to ExternalSink object.
+func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSinkManager, error) {
+	m := &externalSinkManager{
+		decoder: sink_api_old.NewDecoder(),
+	}
+	if externalSinks != nil {
+		if err := m.SetSinks(externalSinks); err != nil {
 			return nil, err
 		}
 	}
-	decoder := sink_api_old.NewDecoder()
-	return &externalSinkManager{
-		externalSinks: externalSinks,
-		decoder:       decoder,
-	}, nil
+	return m, nil
 }
 
-// TODO(vmarmol): Paralellize this.
 func (self *externalSinkManager) Store(input interface{}) error {
 	data, ok := input.(source_api.AggregateData)
 	if !ok {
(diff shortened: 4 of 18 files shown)


mvdan:
FWIW I'll update this PR with integration tests for the new endpoint. And we might want to fix the `Unregister()` TODOs before merging too.

sinks/types.go
```diff
@@ -15,8 +15,15 @@
...
+import "github.com/GoogleCloudPlatform/heapster/sinks/api/v1"
+
 // TODO: Once in-memory storage is implemented, make the manager extract data from the in-memory store periodically.
-type ExternalSinkManager interface {
+type Sink interface {
 	Store(input interface{}) error
 	DebugInfo() string
 }
+
+type ExternalSinkManager interface {
```
vmarmol: Just SinkManager?

sinks/types.go
```diff
@@ -15,8 +15,15 @@
...
+	Sink
+	SetSinks([]v1.ExternalSink) error
```
vmarmol: We should document this (don't follow the bad example we set above :) ). Lets also documen tthe ExternalSinkManager

api/v1/api.go
```diff
@@ -59,6 +59,21 @@ func (a *Api) Register(container *restful.Container) {
 		Operation("exportmetricsSchema").
 		Writes(TimeseriesSchema{}))
 	container.Add(ws)
+	ws = new(restful.WebService)
+	ws.Path("/api/v1/sinks").
+		Doc("Sinks currently in use").
```
vmarmol: "Configuration for Heapster sinks for exporting data"

api/v1/api.go
```diff
@@ -59,6 +59,21 @@ func (a *Api) Register(container *restful.Container) {
...
+		Produces(restful.MIME_JSON)
+	ws.Route(ws.POST("").
+		To(a.setSinks).
+		Doc("set the current sinks").
```
vmarmol: We should document this format and provide examples. Something more simple here, but we should add something to `docs/`

vmarmol:
Lets to the integration test in another PR

sinks/external.go
```diff
@@ -43,18 +43,19 @@ func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSink
...
+}
 
-	for _, externalSink := range externalSinks {
-		err := externalSink.Register(descriptors)
-		if err != nil {
-			return nil, err
-		}
+// NewExternalSinkManager returns an external sink manager that will manage pushing data to all
+// the sinks in 'externalSinks', which is a map of sink name to ExternalSink object.
+func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSinkManager, error) {
```
vmarmol: NewSinkManager

sinks/types.go
mvdan: ExternalSinkManager was already there, so I didn't touch it. Not sure what the difference between Sink and ExternalSink is, so I didn't change that.

sinks/types.go
mvdan: Sure.

api/v1/api.go
mvdan: Sure.

api/v1/api.go
mvdan: What do you mean by "something more simple here"?

By examples, you mean examples with json data? Should they include curl commands, or is json enough?

manager/sinks.go
```diff
@@ -22,9 +22,9 @@ import (
 	sink_api "github.com/GoogleCloudPlatform/heapster/sinks/api/v1"
 )
 
-func newSinks() ([]sink_api.ExternalSink, error) {
+func newSinks(sinkUris []string) ([]sink_api.ExternalSink, error) {
```
vmarmol: How we do this is so nasty... (not your code). This whole overloading of a URI and using a string instead of a strongly-typed thing really doesn't sit well with me. WDYT of making this a slightly stronger and friendlier API? At the very least we can do:

```Go
type SinkConfig struct {
    // Name of the sink.
    Name string
    // Arguments for configuring the sink.
    ConfigArgs string
}
```

It won't be as nice for flags, but it'll work much better for a REST API. I'd even argue to make the SinkArgs more concrete. Maybe make this a struct with configs for all the sinks?

```Go
type SinkConfig struct {
    GCM GCMConfig `json:"gcm,omitempty"`
    InfluxDB InfluxDBConfig `json:"gcm,omitempty"`
    ...
}
```

sinks/external.go
mvdan: Then we'd have to change lots more, like `externalSinkManager` to `sinkManager`. This goes hand in hand with "don't know the difference between Sink and ExternalSink".

sinks/types.go
vmarmol: We can make the interface SinkManager and have the implementation be ExternalSinkManager

api/v1/api.go
vmarmol: I guess we don't have to add anything here, we're as verbose as we are elsewhere. We should have more docs for it though. Questions always come :)

vmarmol: more docs -> `docs/`

sinks/external.go
vmarmol: I think this was meant as an `ExternalSink` being a sink of other sinks. That didn't quite pan out, it is more of a manager of sinks.

sinks/types.go
mvdan: SGTM

manager/sinks.go
mvdan: Agreed on the string stuff. I saw that `Uris` was just `[]string`, so I went with that. I've worked with the `flag.Value` interface before, I could get that working.

Haven't looked into sink configs yet, but I guess this is for exporting them on a `GET /api/v1/sinks`, right?

I'd actually prefer to do all of this in a separate PR if that's OK. Since this is just continuing with what we were doing already, really.

vmarmol: I'm OK with doing in another PR. I don't think we should change the flags though. Writing JSON to flags is not the nicest...

sinks/external.go
mvdan: So then we're keeping `Sink`, `ExternalSink` and `SinkManager`, correct?

manager/sinks.go
mvdan: Hadn't noticed that you were describing the same struct in two different ways. What about it being some sort of interface with the underlying type being the actual config, be it `GCMConfig` or whatever?

The only problem then would be the json key names.

sinks/external.go
vmarmol: I'd say: `s/ExternalSink/sinkManager/`

mvdan: Right, so `sinkManager` is for now the only implementation of `SinkManager`.

mvdan:
Pushed the four changes in four extra commits, will amend into original commit once you're OK with them.

sinks/external.go
```diff
@@ -17,6 +17,7 @@ package sinks
 import (
```
vmarmol: rename this file to `manager.go`?

sinks/external.go
```diff
@@ -43,22 +43,23 @@ func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSink
...
+	return descriptors
+}
 
-	for _, externalSink := range externalSinks {
-		err := externalSink.Register(descriptors)
-		if err != nil {
-			return nil, err
-		}
+// NewSinkManager returns an external sink manager that will manage pushing data to all
+// the sinks in 'externalSinks', which is a map of sink name to ExternalSink object.
```
vmarmol: comment is stale

sinks/external.go
```diff
@@ -24,14 +25,13 @@ import (
 	"github.com/golang/glog"
 )
 
-type externalSinkManager struct {
+type sinkManager struct {
 	decoder       sink_api_old.Decoder
 	externalSinks []sink_api.ExternalSink
+	sync.RWMutex
```
vmarmol: Don't embed the Mutex since it makes its functions public on the object

vmarmol: What does this mutex protect? Let's document it :)

sinks/external.go
```diff
@@ -43,22 +43,23 @@ func NewExternalSinkManager(externalSinks []sink_api.ExternalSink) (ExternalSink
...
 	}
-	decoder := sink_api_old.NewDecoder()
-	return &externalSinkManager{
-		externalSinks: externalSinks,
-		decoder:       decoder,
-	}, nil
+	return m, nil
 }
 
 // TODO(vmarmol): Paralellize this.
```
vmarmol: unrelated: Can remove this TODO

sinks/external.go
```diff
@@ -110,9 +113,45 @@ func (self *externalSinkManager) DebugInfo() string {
...
+	self.RLock()
+	defer self.RUnlock()
 	for _, externalSink := range self.externalSinks {
 		desc += fmt.Sprintf("\n\t\t%s", externalSink.DebugInfo())
 	}
 
 	return desc
 }
+
+func inSlice(sink sink_api.ExternalSink, sinks []sink_api.ExternalSink) bool {
```
vmarmol: Document both functions please

vmarmol: SetSinks is probably fine actually, since it implements the interface

sinks/external.go
vmarmol: Actually, now that I read more of the code on the side it looks like the code is currently transitioning to `ExternalSink` as the interface. So yes, let's not rename ExternalSinkManager -> SinkManager. Sorry :) at least you have the separate commit!

vmarmol:
Also take a look at the code for the existing ExternalSinks and lets see if any of them should do something in their `Unregister()`

sinks/external.go
mvdan: OK!

sinks/external.go
mvdan: Yep.

sinks/external.go
mvdan: Good point.

sinks/external.go
mvdan: Sure.

sinks/external.go
mvdan: Sure.

mvdan:
I'll try :)

mvdan:
Seems like `gcm` needs code in its `Unregister()`, since its `Register()` then calls `addMetric()` in `gcm/core.go` which sends a `POST`. We probably want to send another request for `Unregister()`.

`hawkular` also does save some stuff in `self.reg` and `self.models`, so we probably want to do something there as well.

`gcmautoscaling` looks very much like `gcm`. Although it ignores the metrics passed via `Register()`, so I'm guessing nothing needs to be done? Or perhaps just unregistering the autoscaling metrics?

The rest all have no-ops, so I'm betting on no-ops for their counterparts too.

mvdan:
I'll be back soon with another commit for the unregister bits, think I got the hang of it.Sure.

mvdan:
Unit tests added and I'm pretty confident with the `Unregister()` code for the `gcm` and `gcmautoscaling` sinks.

Don't know how the `hawkular` one works so I've tried putting something together and left a TODO. @burmanm you seem to have written the `Register()` code, any ideas?

burmanm:
@mvdan I think it might be simpler if you just recreate &HawkularSInk, as the Register() takes care of the initialization process (by fetching definitions from the server), so there's nothing valuable left. I assume at least that the GC would then take care of removing old maps from the memory.

mvdan:
@burmanm Cool, thanks! I can't do that with GCM because of the GET/POST requests, but since as you say it's all local here, the GC will just wipe it :)

mvdan:
@vmarmol done. Thanks again @burmanm!

mvdan:
FWIW, I'm having issues with gofmt because I'm running 1.4.2 and my gofmt makes lots more changes that travis isn't complaining about.

mvdan:
@vmarmol just added the last unit test. If you confirm everything is ok I'll rebase the commits and they'll be ready for merging.

mvdan:
@vmarmol last issues fixed and commits rebased.

vmarmol:
Looks like we need a rebase, otherwise LGTM!

mvdan:
Rebased and opened another PR for the integration test.

vmarmol:
Thanks @mvdan! Will wait for green and merge.

docs/sink-configuration.md
```diff
@@ -75,3 +75,9 @@ If `HAWKULAR_SERVER_URL` includes any path, the default `hawkular/metrics` is ov
 The following options are available:
 
 * `tenant` - Hawkular-Metrics tenantId (default: `heapster`)
+
+## Modifying the sinks at runtime
```
vishh: Can you add an example for dynamically adding a sink?

mvdan: Something like the following?

```markdown
echo '["gcm", "influxdb:http://monitoring-influxdb:8086"]' | curl \
    --insecure -u admin:<password> -X POST -d @- \
    -H "Accept: application/json" -H "Content-Type: application/json" \
    https://<master-ip>/api/v1/proxy/namespaces/kube-system/services/monitoring-heapster/api/v1/sinks
```

mvdan: Note that `--insecure` is necessary for curl to work with the self-signed SSL certificate.

vishh: Yes please.

mvdan: See #427

PR #400, opened 2015-07-10
Makefile fixes

mvdan:
These don't change much really, but bugged me :)

### DIFF
--- a/deploy/docker/build.sh
+++ b/deploy/docker/build.sh
@@ -1,9 +1,16 @@
 #!/bin/bash
 
+IMAGE=${1-heapster:canary}
+
 set -e
 
 pushd $( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+pushd ../..
+make generate
+popd
+
 godep go build -a github.com/GoogleCloudPlatform/heapster
 
-docker build -t heapster:canary .
+docker build -t $IMAGE .
 popd
--- a/integration/heapster_api_test.go
+++ b/integration/heapster_api_test.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
 	"strings"
 	"testing"
 	"time"
@@ -37,7 +36,6 @@ const (
 	kTestZone        = "us-central1-b"
 	targetTags       = "kubernetes-minion"
 	heapsterBuildDir = "../deploy/docker"
-	makefile         = "../Makefile"
 )
 
 var (
@@ -97,9 +95,6 @@ func buildAndPushHeapsterImage(hostnames []string) error {
 		return err
 	}
 	if err := os.Chdir(heapsterBuildDir); err != nil {
-		return err
-	}
-	if err := make("build", filepath.Join(curwd, makefile)); err != nil {
 		return err
 	}
 	if err := buildDockerImage(*heapsterImage); err != nil {
--- a/integration/utils.go
+++ b/integration/utils.go
@@ -22,16 +22,8 @@ import (
 	"path"
 )
 
-func make(command, makefile string) error {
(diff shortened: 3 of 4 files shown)


vmarmol:
ok to test

vmarmol:
LGTM, thanks!

mvdan:
Hang on a sec, the integration tests don't work with this. Don't know why though.

mvdan:
The code in the integration tests that runs docker and make is kinda broken, currently fixing it. Will be back with a larger patch.

mvdan:
Fixed and updated.

vmarmol:
Needs a rebase btw

deploy/docker/build.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
+IMAGE=${1-heapster:canary}
```
vmarmol: just heapster?

integration/heapster_api_test.go
```diff
@@ -99,9 +97,6 @@ func buildAndPushHeapsterImage(hostnames []string) error {
 	if err := os.Chdir(heapsterBuildDir); err != nil {
 		return err
 	}
-	if err := make("build", path.Join(curwd, makefile)); err != nil {
```
vmarmol: Won't we need the makefile for the generate stuff?

deploy/docker/build.sh
mvdan: That's what was in the original build.sh:

```
-docker build -t heapster:canary .
+docker build -t $IMAGE .
```

What would you call it? The Makefile uses `$PREFIX/heapster:$TAG`, but I didn't know what prefix to use here.

integration/heapster_api_test.go
mvdan: Right yes. I'll add a Makefile target for that and fix the PR.

mvdan:
Rebased (it conflicted with another PR of mine, duh) and fixed the generate issue.

mvdan:
Yet another timeout :( but other than the container name thing this should be ready to merge.

deploy/docker/build.sh
vmarmol: Ah maybe my bash is weak :) I intended for this to be "heapster:canary", but this seems like we're selecting a var with that name?

mvdan: `${1-heapster:canary}` means `$1 if $1 is empty (or undefined, same thing) otherwise "heapster:canary"`. There is an alternate and more popular notation, `${1:-heapster:canary}`, which does the same.

vmarmol: Ah thanks! I've seen the second but not the first notation.

integration/utils.go
```diff
@@ -22,16 +22,8 @@ import (
...
-	out, err := exec.Command("make", "-f", makefile, command).CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("failed to build go binary (%q) - %q", err, out)
-	}
-	return nil
-}
-
 func buildDockerImage(imageName string) error {
-	out, err := exec.Command("docker", "build", "-t", imageName, ".").CombinedOutput()
+	out, err := exec.Command("./build.sh", imageName).CombinedOutput()
```
vmarmol: Why change to this instead of what we did before with the makefile?

mvdan: Firstly, because both the script and this go func do essentially the same. So why duplicate code? For example, we found out that the bash code was wrong (it was missing the `go generate` stuff).

Secondly, because now that the root Makefile no longer uses absolute build paths, running make in the root dir will create a `heapster` binary in the root directory. So we'd have to copy it over. Which complicates the Go code in this function even more.

So it's more a question of whether to reimplement all of `build.sh` in Go, which seemed unnecessary.

vmarmol: SGTM, I see you made the fixes to the Makefile as well.

vmarmol:
LGTM

PR #401, opened 2015-07-10
Properly do all kubeVersions in the integration tests

mvdan:
If we always return at each iteration, we'll only ever test the first version.
Seems like a typo or silly mistake.


vmarmol:
ok to test

vmarmol:
LGTM

PR #402, opened 2015-07-10
Use filepaths for local file paths

mvdan:
This would definitely be an issue for running the integration tests on
non-unix-like systems.


vmarmol:
ok to test

vmarmol:
LGTM

PR #403, opened 2015-07-13
Added exporting cpu/mem limits based on Kubernetes rather than cadvisor

### LINKED ISSUE #399: Export the machine container with the proper CPU limit
Today we convert shares to CPU limit, but this doesn't work for the root container since it always has 1024 shares even though it corresponds to all cores on the machine. We should take the NumCores from the machine and export that instead.

piosz:
Fixes #399

### DIFF
--- a/sources/nodes/types.go
+++ b/sources/nodes/types.go
@@ -30,6 +30,10 @@ type Info struct {
 	InternalIP string
 	// An optional ID assigned to nodes by either the cloud provider or user.
 	ExternalID string
+	// Cpu capacity of node in millicores.
+	CpuCapacity uint64
+	// Memory capacity of node in bytes.
+	MemCapacity uint64
 }
 
 // NodeList contains the nodes that an instance of heapster is required to
--- a/sources/kube_nodes.go
+++ b/sources/kube_nodes.go
@@ -82,6 +82,9 @@ func (self *kubeNodeMetrics) updateStats(host nodes.Host, info nodes.Info, start
 		hostContainer = &hostCopy
 		containers = append(containers[:hostIndex], containers[hostIndex+1:]...)
 	}
+	// This is temporary workaround for #399. To make unit consistent with cadvisor normalize to a conversion factor of 1024.
+	hostContainer.Spec.Cpu.Limit = info.CpuCapacity * 1024 / 1000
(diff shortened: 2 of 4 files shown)


piosz:
@vmarmol

#399 was only about fixing cpu, but I've also fixed memory to make both resources exported consistently.

sources/nodes/types.go
```diff
@@ -30,6 +30,10 @@ type Info struct {
 	InternalIP string
 	// An optional ID assigned to nodes by either the cloud provider or user.
 	ExternalID string
+	// Cpu capacity of node as number of 1/1024 share of core.
+	CpuCapacity uint64
```
vmarmol: This is exporting in shares? At the Heapster level (and for nodes) we should probably export in whole cores. Shares is hard to reason about at a node level.

vmarmol:
Outside the issue with shares vs. cores, LGTM

sources/nodes/types.go
piosz: Just wanted to make it consistent with the current unit, but it will be definitely easier to use millicores instead.

piosz:
Changed to use millicores internally. We probably don't want to use whole core (not millicores) to support also machines with fractional part of virtual CPU. Please let me know if it addressed your comment.

sinks/api/v1/supported_metrics.go
```diff
@@ -66,8 +66,7 @@ var statMetrics = []SupportedStatMetric{
 			return spec.HasCpu && (spec.Cpu.Limit > 0)
 		},
 		GetValue: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []InternalPoint {
-			// Normalize to a conversion factor of 1000.
-			return []InternalPoint{{Value: int64(spec.Cpu.Limit*1000) / 1024}}
+			return []InternalPoint{{Value: int64(spec.Cpu.Limit)}}
```
vmarmol: I would have expected we still do this? We fixed it for node containers but not for general containers no?

piosz: I see the problem. Fixed.

piosz:
ping

vmarmol:
LGTM, sorry this dropped out of my radar!

piosz:
No worries. Thanks you!

PR #405, opened 2015-07-13
External save block

mvdan:


mvdan:
This is really weird. I couldn't find a reference to this issue anywhere, so to confirm that indeed reading from an empty buffered channel doesn't block, I set up a play.golang.org example: https://play.golang.org/p/abMIOJa0FJ

And it does indeed block. So this issue seems to come from someplace else.

vmarmol:
LGTM

mvdan:
@vmarmol now it is ready to merge :)

vmarmol:
Thanks :) LGTM

PR #406, opened 2015-07-13
[Schema->Model] Rename the Schema package to Model

afein:
This PR renames the schema package to model, as it is a more fitting name and avoids confusion with actual schema references in the codebase.

From now on, all further PRs will be prefixed with [Model X]. Specifically, the next PR will be [Model 7], which will implement some basic cluster-level API endpoints.

\cc @rjnagal @vmarmol


rjnagal:
LGTM

PR #409, opened 2015-07-14
Fix onlyKeepLatestStat to return the latest rather than the oldest stat

a-robinson:
This was causing the REST API to return nearly 10 minute old data due to the default 10 minute buffer in the GC store.

As a pretty strong disclaimer, I haven't actually tested this in a cluster and I didn't see any unit tests that it'd be easy to add a check for this to. The unit tests do all pass, however. I can spend some time trying to test it in a live cluster tomorrow if you'd like.

@vmarmol


vmarmol:
ok to test

vmarmol:
LGTM, thanks @a-robinson!

PR #413, opened 2015-07-14
Uris type

mvdan:


manager/flags_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUriString(t *testing.T) {
+	for _, c := range [...]struct {
```
afein: nit: creating the struct array before the for statements would help a lot with readability, both here and in the other test cases.

sinks/gcmautoscaling/driver.go
```diff
@@ -263,7 +265,10 @@ func init() {
 	extpoints.SinkFactories.Register(CreateGCMScalingSink, "gcmautoscaling")
 }
 
-func CreateGCMScalingSink(_ string, _ map[string][]string) ([]sink_api.ExternalSink, error) {
+func CreateGCMScalingSink(uri *url.URL) ([]sink_api.ExternalSink, error) {
+	if *uri != (url.URL{}) {
+		return nil, errors.New("gcmautoscaling sinks don't take arguments")
```
afein: nit: we primarily use fmt.Errorf for new errors in Heapster.

afein:
overall LGTM, just two nits

manager/flags_test.go
mvdan: Fair enough.

sinks/gcmautoscaling/driver.go
mvdan: Even if they don't have any formatting? I'm guessing this is to be consistent and to not add the `errors` import, correct?

mvdan:
FWIW, the old code used to do an `url.Parse()` to get the type, opaque url and options for the sink/source. The new code simply splits by `:` and makes a `url.URL` out of the part in the right. This results in simpler code, IMHO, and doesn't break anything.

mvdan:
PR updated with the two fixes.

manager/flags_test.go
```diff
@@ -0,0 +1,172 @@
...
+					RawQuery: "key=val&key2=val2",
+				},
+			},
+			"influxdb:http://monitoring-influxdb:8086?key=val&key2=val2",
+		},
+	}
+	for _, c := range tests {
+		got := c.in.String()
+		if got != c.want {
+			t.Fatalf("%v.String() got %v, wanted %v", c.in, got, c.want)
```
afein: For consistency with the rest of the codebase, can you use assert.Equal instead?

manager/flags_test.go
```diff
@@ -0,0 +1,172 @@
...
+					Val: url.URL{Path: "foo"},
+				},
+			},
+			"[gcm influxdb:foo]",
+		},
+	}
+	for _, c := range tests {
+		got := c.in.String()
+		if got != c.want {
+			t.Fatalf("%v.String() got %v, wanted %v", c.in, got, c.want)
```
afein: same here

afein:
just one nit on consistency, sorry I missed it before.

mvdan:
Amended.

manager/flags.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"net/url"
+	"os"
+	"strings"
+)
+
+type Uri struct {
```
vmarmol: Hmmm I tend to like the alternative we mentioned better: 

```Go
type SinkConfig struct {
    GCM GCMConfig `json:"gcm,omitempty"`
    InfluxDB InfluxDBConfig `json:"gcm,omitempty"`
    ...
}
```

It is stronger typed, self-documented, and easier on a REST API. It totally sucks for a flag interface though. WDYT? I am not against having the flag interface be different from the REST API.

mvdan: I don't think this is good for the flag interface, yeah. Gets too much done too soon.

I didn't do the REST API part yet, what I did for now is simply change the underlying type to be stronger without changing any of the behaviour.

I seem to recall you said that you'd like the endpoint to use something other than the strings that the flags use. Would that be for GET, POST or both? In a way I'd rather keep it the way it is now, since it's more intuitive and it uses the same format everywhere.

Although at the same time, a json detailed description of each sink would be useful. Perhaps a separate endpoint, something like `/api/v1/sink-configs`?

PR #417, opened 2015-07-14
[Model 7] REST endpoints

afein:
Part 7 of #342 

This PR implements the initial set of Getter methods for the model, as well as their corresponding REST endpoints. 

Suggested reading order:
- model/types.go
- api/v1/api.go
- api/v1/model_types.go
- api/v1/model_handlers.go
- models/impl.go
- models/impl_test.go
- manager/manager.go
- heapster.go
- models/util.go
- models/util_test.go
- store/cma_store.go
- store/cma_store_test.go

The rework of parseMetrics and InfoType is still a WIP, as it was aiming to correctly calculate instantaneous cpu usage from the cumulative metric counterpart. This is not achieved with the existing implementation, but might be addressed in a future commit on this same PR (unless merged early!)

\cc @rjnagal @vmarmol


api/v1/model_handlers.go
```diff
@@ -0,0 +1,304 @@
...
+// These metric names can be used to extract metrics from the various model entities.
+func (a *Api) availableMetrics(request *restful.Request, response *restful.Response) {
+	cluster := a.manager.GetCluster()
+	result := cluster.GetAvailableMetrics()
+	response.WriteEntity(result)
+}
+
+// clusterMetrics returns a metric timeseries for a metric of the Cluster entity.
+func (a *Api) clusterMetrics(request *restful.Request, response *restful.Response) {
+	var err error
```
mvdan: nit: you can probably remove this line

mvdan: And also the others in this file.

heapster.go
```diff
@@ -36,10 +36,12 @@ var (
 	argPollDuration    = flag.Duration("poll_duration", 10*time.Second, "The frequency at which heapster will poll for stats")
 	argStatsResolution = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats. Acceptable values are in the range [1 second, 'poll_duration')")
 	argAlignStats      = flag.Bool("align_stats", false, "Whether to align timestamps of metrics to multiplicity of 'stats_resolution'")
+	argCacheDuration   = flag.Duration("cache_duration", 10*time.Minute, "The total duration of the historical data that will be cached by heapster.")
+	argUseModel        = flag.Bool("use_model", false, "When true, the internal model representation will be used.")
+	argModelResolution = flag.Duration("model_resolution", 20*time.Second, "The resolution of the timeseries stored in the model. Applies only if use_model is true.")
```
mvdan: I just noticed that the other flags don't end with a period. Perhaps do the same?

model/impl.go
```diff
@@ -46,6 +47,175 @@ func newRealCluster(tsConstructor func() store.TimeStore, resolution time.Durati
...
+
+// GetNodeMetric returns a metric of a node entity, along with the latest timestamp.
+// GetNodeMetric receives as arguments the name of a metric and a starting timestamp.
+// GetNodeMetric returns a slice of TimePoints for that metric, with times starting AFTER the starting timestamp.
+func (rc *realCluster) GetNodeMetric(hostname string, metric_name string, start time.Time) ([]store.TimePoint, time.Time, error) {
+	rc.lock.RLock()
+	defer rc.lock.RUnlock()
+
+	if len(rc.Nodes) == 0 {
+		return nil, time.Time{}, fmt.Errorf("the model is not populated yet")
```
mvdan: I've seen people use "zero" variables, like `zt := time.Time{}` at the beginning of the function, so that you don't repeat the typed zero value many times. Reduces columns and makes it easier to change the zero value or type if it ever needs to change.

api/v1/model_handlers.go
afein: right! The var definition was left over before I ported some code to parseRequestStartParam

model/impl.go
afein: Good call. We generally define a zeroTime variable if there are multiple instances of time.Time{} within the same scope

afein:
nits addressed in the latest commit (squash before merge pending)

mvdan:
Cool! That was just a quick read, haven't gone through the code extensively FWIW. Also, you probably need to rebase since there are a couple of merge conflicts due to #413. Easy to fix though.

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+
+	"github.com/GoogleCloudPlatform/heapster/store"
+)
+
+// Register the Model API endpoints
+func (a *Api) RegisterModel(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/api/v1/model").
+		Doc("Exports all aggregated Cluster-level metrics since the specified start time").
```
rjnagal: specified start time implies there are params that we should describe here?

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		Consumes("*/*").
+		Produces(restful.MIME_JSON)
+
+	// The /endpoint returns a list of all the metrics that are available in the model
```
rjnagal: space between / and endpoint.

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		Operation("clusterMetrics"))
+
+	// The /cluster endpoint exposes aggregated metrics for the Cluster entity of the model.
+	ws.Route(ws.GET("/cluster/{metric-name}").
+		To(a.clusterMetrics).
+		Filter(compressionFilter).
+		Doc("export all aggregated cluster-level metrics").
+		Operation("clusterMetrics").
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metrics").DataType("string")).
```
rjnagal: what type is the start time?

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		Writes(MetricResult{}))
+
+	// The /node/ endpoint exposes metrics for the Node entity of the model.
+	ws.Route(ws.GET("/node/{node-name}/{metric-name}").
```
rjnagal: does /node give data about all nodes? or list available nodes? How does the caller discover node names?

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		To(a.nodeMetrics).
+		Filter(compressionFilter).
+		Doc("export all node-level metrics").
+		Operation("nodeMetrics").
+		Param(ws.PathParameter("node-name", "The name of the node to lookup").DataType("string")).
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metrics").DataType("string")).
+		Writes(MetricResult{}))
+
+	// The /namespace/ endpoint exposes aggregated metrics for the Namespace entity of the model.
```
rjnagal: same comment for namespace and pods: how do you find all available namespaces and pods? Does the /namespace endpoint gives data for all namespaces?

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+	ws.Route(ws.GET("/namespace/{namespace-name}/{metric-name}").
+		To(a.namespaceMetrics).
+		Filter(compressionFilter).
+		Doc("export all aggregated namespace-level metrics").
+		Operation("namespaceMetrics").
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metrics").DataType("string")).
```
rjnagal: For all endpoints, why don't we need an end time? It looks like some of the calls can return too much data and ability to exactly chunk might be useful?

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		Writes(MetricResult{}))
+
+	// The /namespace/{}/pod endpoint exposes aggregated metrics for the Pod entity of the model.
```
rjnagal: pod between {}

rjnagal: ah, you meant /namespace/{namespace-name}/pod .

api/v1/model_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		Doc("export a container-level metric for a pod container").
+		Operation("podContainerMetrics").
+		Param(ws.PathParameter("namespace-name", "The name of the namespace to use").DataType("string")).
+		Param(ws.PathParameter("pod-name", "The name of the pod to use").DataType("string")).
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metrics").DataType("string")).
+		Writes(MetricResult{}))
+
+	// The /node/{}/container endpoint exposes metrics for the Free Container entity of the model.
+	ws.Route(ws.GET("/node/{node-name}/container/{container-name}/{metric-name}").
```
rjnagal: that this calls returns only free containers and not all containers is confusing. rename?

manager/manager.go
```diff
@@ -59,11 +66,21 @@ type syncData struct {
 	mutex sync.Mutex
 }
 
-func NewManager(sources []source_api.Source, sinkManager sinks.ExternalSinkManager, res, bufferDuration time.Duration, align bool) (Manager, error) {
+func NewManager(sources []source_api.Source, sinkManager sinks.ExternalSinkManager, res, bufferDuration time.Duration, useModel bool, modelRes time.Duration, align bool) (Manager, error) {
+	// TimeStore constructor passed to the cluster implementation.
+	// TODO(alex): determine default analogy of cache duration to Timestore durations.
```
rjnagal: use your github id in TODOs for better name resolution :)

rjnagal: what does 'default analogy' mean?

api/v1/model_handlers.go
afein: Most use cases of the API are meant for fresh data (scheduling, analytics, resource recommendations). As the model only stores a short total duration by default, exporting all the data  is not particularly wasteful.
Nonetheless, adding an end time to the APIs and getters is less than trivial, should I do it?.

manager/manager.go
afein: default analogy means the ratio of model_duration to cache_duration (passed as bufferDuration to NewManager). As of now, model_duration = 5 * cache_duration for no specific reason. However, other options include using a flag to determine the exact duration, while defaulting to a multiple of cache_duration.

afein:
Addressed all points in the latest commit. Also, renamed the /pod, /namespace, /node, /container endpoints to plural (/pods, /namespaces, /nodes, /containers), so that hitting api/v1/model/nodes would present a list of all the available hostnames and so on.

The actual implementation of these endpoints will be sent out on a different PR (otherwise this one will bloat up to around 1.5k LOC)

api/v1/model_handlers.go
```diff
@@ -0,0 +1,305 @@
...
+		Operation("podContainerMetrics").
+		Param(ws.PathParameter("namespace-name", "The name of the namespace to use").DataType("string")).
+		Param(ws.PathParameter("pod-name", "The name of the pod to use").DataType("string")).
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metrics").DataType("string")).
+		Writes(MetricResult{}))
+
+	// The /nodes/{node-name}/freecontainers/{container-name}/{metric-name} endpoint exposes
+	// a metric for a free Container entity of the model.
+	ws.Route(ws.GET("/node/{node-name}/container/{container-name}/{metric-name}").
```
rjnagal: update to freecontainers.

rjnagal:
Handling end time and listing in separate PR is fine. It'll be easier to review :)

The only remaining nit I have around API is that the doc part of most endpoints feel under-specified. eg does nodename in the endpoint correspond to a fqdn, kube host id, or maybe ip?

Would it also help if we bundled all params for a request into a single structure, eg NodeRequest for the implementation methods. That way we can have a single well-documented types.go and addition of new params would be a bit cleaner. WDYT?

afein:
squashed all commits, resolved merge conflict and addressed the last inline note (update to freecontainers).

The hostname is the exact hostname that is reported by cadvisor, which I believe is the kube host id. I will document this more in the followup PR, but any confusion caused by that shouldn't be a problem as the API consumers are expected to retrieve a list of available node names at the /nodes endpoint.

I agree that bundling params into Request structs would be cleaner, I will address that in the followup PR.

rjnagal:
ok, let's merge this and get going with the follow on changes.

PR #418, opened 2015-07-15
Fix fmt format issue introduced in 67415bd031b

mvdan:


afein:
LGTM
FYI, you can use a commit link to point to a specific commit: https://github.com/mvdan/heapster/commit/67415bd031b1191e0fa0a9902d6d1c8e1cf47b73

sinks/external.go
```diff
@@ -106,19 +106,19 @@ func (self *externalSinkManager) DebugInfo() string {
...
+		fmt.Fprintf(b, "\t\t%s: %s", supported.Name, supported.Description)
 	}
 
 	// Add labels being used.
 	fmt.Fprintln(b, "\tExported labels:")
 	for _, label := range sink_api.SupportedLabels() {
-		fmt.Fprintln(b, "\t\t%s: %s", label.Key, label.Description)
+		fmt.Fprintf(b, "\t\t%s: %s", label.Key, label.Description)
 	}
 	fmt.Fprintln(b, "\tExternal Sinks:")
```
afein: aren't you missing a newline before \tExternal Sinks?

sinks/external.go
```diff
@@ -106,19 +106,19 @@ func (self *externalSinkManager) DebugInfo() string {
...
 	self.sinkMutex.RLock()
 	defer self.sinkMutex.RUnlock()
 	for _, externalSink := range self.externalSinks {
-		fmt.Fprintln(b, "\t\t%s", externalSink.DebugInfo())
```
afein: same here, the format string was "\n\t\t%s" in the Sprintf version of the file

sinks/external.go
mvdan: Damn, forgot about that when moving from `ln` to `f`.

mvdan:
You mean using a github commit http link on a commit message? That's way too long IMHO. For a pull request name/description sure, but for a git commit summary/message I've never seen it done.

I assumed github would pick up the hash and link to it. I could've used the link for the PR title, yeah.

mvdan:
Amended. Thanks for picking that up!

afein:
LGTM. For the commit link, I meant only for the description, so whoever reviews your PR can just click through to the previous commit. You could mention a PR number in a commit message, though, like #418

mvdan:
FWIW, the PR page doesn't pick up the link but the commit history page does pick it up.

PR #424, opened 2015-07-21
Add StatStore interface and first implementation

mvdan:
A StatStore is a layer on top of a TimeStore that provides stats over it. For
now, it does average, max and three percentiles. The stats are cached and
recalculated every time the data is changed.

CC @afein @vmarmol @vishh

I'll add unit tests tomorrow.

### DIFF
--- a/store/stat_store.go
+++ b/store/stat_store.go
@@ -19,6 +19,8 @@ import (
 	"sort"
 	"sync"
 	"time"
+
+	"github.com/google/cadvisor/summary"
 )
 
 type NthPercentile uint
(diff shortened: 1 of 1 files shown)


mvdan:
Note: I tried using a `sync.RWMutex` but I got stuck at escalating `RLock()` into `Lock()` when the cache needs updating. A `sync.Mutex` is probably good enough here. The fact that the better option is not feasible probably means that we should use some other strategy in the long run.

mvdan:
Don't merge yet, will update this PR after I've exposed the percentile code in cadvisor so as to not duplicate code.

PR #425, opened 2015-07-21
Store fixes

mvdan:


afein:
LGTM!
Thanks a lot for fixing these, most of them were introduced during debugging and I omitted to clean them up. Can you squash?

mvdan:
I split up the commits on purpose, so as to have one change per commit. This is easier to find and read, and also easier to revert if any of these changes are wrong.

afein:
Fair enough, even though they're definitely correct

PR #434, opened 2015-07-24
Exit on any shell script errors equally on all scripts

mvdan:
Especially on the jenkins one, since we don't want to run the integration
tests if the unit tests already failed.

CC @vishh 

You can see an example of this problem here:

http://104.154.52.74/job/heapster-e2e-gce/1714/console

`make` fails, but `godep` is run anyway.


vishh:
LGTM

PR #437, opened 2015-07-27
[Model 9] Sentinels, housekeeping and improved cpu calculation

afein:
This PR finalizes the initial implementation of the model APIs for metrics, by adding sanity checks for edge cases in API calls and cpu usage calculation.

Also, the housekeeping of the model is now decoupled from that of the cache, which corrects some weird behaviors, such as values changing in both the past and future. The new housekeeping interval ensures that there is at least 1 datapoint per entity for each model resolution interval.

Please include this PR in the next release.

\cc @rjnagal @vishh

### DIFF
--- a/api/v1/model_handlers.go
+++ b/api/v1/model_handlers.go
@@ -15,6 +15,7 @@
 package v1
 
 import (
+	"errors"
 	"net/http"
 	"time"
 
@@ -24,6 +25,10 @@ import (
 	"github.com/GoogleCloudPlatform/heapster/model"
 	"github.com/GoogleCloudPlatform/heapster/store"
 )
+
+// errModelNotActivated is the error that is returned when manager.cluster
+// has not beed initialized.
+var errModelNotActivated = errors.New("the model is not activated")
 
 // RegisterModel registers the Model API endpoints.
 // All endpoints that end with a {metric-name} also receive a start time query parameter.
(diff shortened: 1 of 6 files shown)


api/v1/model_handlers.go
```diff
@@ -318,18 +318,27 @@ func (a *Api) containerPaths(request *restful.Request, response *restful.Respons
 // allNodes returns a list of all the available node names in the cluster.
 func (a *Api) allNodes(request *restful.Request, response *restful.Response) {
 	cluster := a.manager.GetCluster()
+	if cluster == nil {
+		response.WriteErrorString(400, "The model is not activated")
```
mvdan: I can't help but think that these three lines are unnecessarily duplicated across the file. What about a new method/function, something like:

```go
func (a *Api) getNonNilCluster() (model.Cluster, error) {
    cluster := a.manager.GetCluster()
    if cluster == nil {
        return nil, fmt.Errorf("the model is not activated")
    }
    return cluster
}
```

And then:

```go
cluster, err := a.getNonNilCluster()
if err != nil {
    response.WriteError(400, err)
}
[...]
```

mvdan: Granted that this pretty much just deduplicates the error string.

vishh: +1

vishh:
LGTM. I will let @mvdan do a final review before merging this.

vishh:
@cadvisorJenkinsBot: retest this please

api/v1/model_handlers.go
mvdan: In fact, for the sole purpose of deduplicating the error string you could just have the error string defined as a const:

```go
const errModelNotActivated = errors.New("the model is not activated")
[...]
response.WriteError(400, errModelNotActivated)
```

Which is much simpler than adding the new method.

mvdan:
LGTM except that nit, which can be ignored.

api/v1/model_handlers.go
mvdan: Correction - apparently errors can't be constants, so var instead.

vishh:
@afein: Can you fix this PR soon? I intend to make a new release of heapster today.

afein:
Fixed according to mvdan's last suggestion and squashed. Sorry for the delay!

PR #440, opened 2015-07-30
[Model 10] Derived stats implementation and APIs

afein:
This PR includes the initial implementation of derived stats on top of the existing model.
This implementation is highly redundant in memory, as DayStores are used for all measured metrics, including limits. 

The future plan is to only allow derived stats calculations when these are explicitly enabled for specific stats (incoming in [Model 12], most probably).

Test coverage of the model is at 96.9%.

Areas that can be structurally and functionally improved:
- DayStore and StatStore implementation

\cc @rjnagal @vishh @mvdan


mvdan:
Jenkins barfed again. But that doesn't matter, because Travis failed for a good reason. Looks like you're trying to use the code I exported in `cadvisor/summary`, but that code isn't in Godeps yet. You probably want to update cadvisor too, besides adding that subpackage.

store/stat_store.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	acc := uint64(0)
+	for _, u := range inc {
+		acc += u
+	}
+	s.cachedAverage = float64(acc) / float64(len(inc))
+	s.cachedFiftieth = inc.GetPercentile(Fiftieth)
+	s.cachedNinetieth = inc.GetPercentile(Ninetieth)
+	s.cachedNinetyFifth = inc.GetPercentile(NinetyFifth)
+	s.cachedMax = inc[len(inc)-1]
```
mvdan: I know I wrote this, but perhaps we should note that at this stage inc is already sorted in a comment. Or sort it ourselves explicitly before grabbing the max.

store/day_store.go
```diff
@@ -0,0 +1,101 @@
...
+// DayMax returns the maximum value that has occured over the past Day.
+// A Timestore is used to ensure that the maximum reflects only the last day's values.
+// TODO(afein): cache the curMax until a new value is inserted in DayMaxima.
+func (ds *DayStore) DayMax() uint64 {
+	zeroTime := time.Time{}
+	allTPs := ds.DayMaxima.Get(zeroTime, zeroTime)
+	curMax := uint64(0)
+	for _, tp := range allTPs {
+		newVal := tp.Value.(uint64)
+		if newVal > curMax {
```
mvdan: nit: you can do `if newVal := tp.Value.(uint64); newVal > curMax {`

store/day_store.go
```diff
@@ -0,0 +1,101 @@
...
+	dayResolution time.Duration
+}
+
+// DayPassed returns true iff the DayStore is populated with more than 24 hours of data.
+func (ds *DayStore) DayPassed() bool {
+	lastTP := ds.Hour.Last()
+	if lastTP == nil || !ds.created.Add(24*time.Hour).Before(lastTP.Timestamp) {
+		return false
+	}
+	return true
```
mvdan: This is actually a question - do you generally accept `if cond { return false } return true` instead of `return cond`? I know `cond` in this case is rather large, but still wondering.

model/impl.go
```diff
@@ -303,6 +304,81 @@ func (rc *realCluster) GetAvailableMetrics() []string {
...
+// getClusterStats extracts the derived stats and uptime for the Cluster entity.
+func (rc *realCluster) GetClusterStats() (map[string]StatBundle, time.Duration, error) {
+	return getStats(rc.InfoType), rc.InfoType.Uptime, nil
+}
+
+// getNodeStats extracts the derived stats and uptime for a Node entity.
+func (rc *realCluster) GetNodeStats(req NodeRequest) (map[string]StatBundle, time.Duration, error) {
+	node, ok := rc.Nodes[req.NodeName]
+	if !ok {
+		return nil, time.Duration(0), fmt.Errorf("the requested node is not present in the cluster")
```
mvdan: I see more duplicated error strings :)

store/day_store.go
afein: I guess due to the size of the condition both could work. However, I will be deleting this function instead, as the most requested feature seems to be providing up-until-24-hours stats, which do not require the dayPassed check.

afein:
Two more commits:
- Separated all the getter implementations and tests outside of impl.go and impl_test.go
- Updated the cadvisor dependency, as I was building against a version on my local GOPATH apparently

afein:
ping @vishh @mvdan

api/v1/model_types.go
```diff
@@ -27,3 +27,21 @@ type MetricResult struct {
 	Metrics         []MetricPoint `json:"metrics"`
 	LatestTimestamp time.Time     `json:"latestTimestamp"`
 }
+
+type Stats struct {
+	Average    uint64 `json:"average"`
+	Percentile uint64 `json:"percentile"`
```
vishh: What percentile?

manager/manager.go
```diff
@@ -70,14 +70,18 @@ type syncData struct {
...
 	var newCluster model.Cluster = nil
 	if useModel {
-		newCluster = model.NewCluster(tsConstructor, modelRes)
+		// TimeStore and DayStore constructors passed to the model.
+		tsConstructor := func() store.TimeStore {
+			// TODO(afein): determine default analogy of cache duration to Timestore durations.
+			return store.NewGCStore(store.NewCMAStore(), time.Hour)
+		}
+		dayConstructor := func() store.DayStore {
+			return store.NewDayStore(20 * time.Minute) // TODO(afein): fine-tune this value
```
vishh: what is the 20 minutes value for?

model/getters.go
```diff
@@ -0,0 +1,364 @@
...
+	for key := range noderef.FreeContainers {
+		res = append(res, key)
+	}
+	return res
+}
+
+// GetAvailableMetrics returns the names of all metrics that are available on the cluster.
+// Due to metric propagation, all entities of the cluster have the same metrics.
+func (rc *realCluster) GetAvailableMetrics() []string {
+	rc.lock.RLock()
```
vishh: nit: `lock` can be embedded into `rc`.

model/util.go
```diff
@@ -156,3 +158,28 @@ func instantFromCumulativeMetric(value uint64, stamp time.Time, prev *store.Time
...
+
+// getStats extracts derived stats from an InfoType.
+func getStats(info InfoType) map[string]StatBundle {
+	res := make(map[string]StatBundle)
+	for key, ds := range info.Metrics {
+		res[key] = StatBundle{
+			Minute: Stats{
+				Average:    ds.Hour.Last().Value.(uint64),
+				Percentile: ds.Hour.Last().Value.(uint64),
+				Max:        ds.Hour.Last().Value.(uint64),
```
vishh: Why are all these values the same?

model/util.go
```diff
@@ -156,3 +158,28 @@ func instantFromCumulativeMetric(value uint64, stamp time.Time, prev *store.Time
...
+			},
+			Hour: Stats{
+				Average:    ds.Hour.Average(),
+				Percentile: ds.Hour.Percentile(0.95),
```
vishh: Make it explicit that you are providing 95%ile in the field name.

store/stat_store.go
```diff
@@ -0,0 +1,152 @@
...
+	// GetMax gets the max value of the data
+	Max() uint64
+	// Percentile gets the specified Nth percentile of the data
+	Percentile(n float64) uint64
+}
+
+type statStore struct {
+	ts TimeStore
+
+	cacheLock         sync.Mutex
```
vishh: nit: This can be embedded.

store/stat_store.go
```diff
@@ -0,0 +1,152 @@
...
+	return s.ts.Delete(start, end)
+}
+
+func (s *statStore) getAll() []TimePoint {
+	zeroTime := time.Time{}
+	return s.ts.Get(zeroTime, zeroTime)
+}
+
+func (s *statStore) Last() *TimePoint {
+	all := s.getAll()
```
vishh: Why not just the last sample here? Reading all the data back seems wasteful.

vishh:
@afein: Did an initial pass. This PR could have been split into several smaller ones for easier/better code review.

api/v1/model_types.go
afein: It is the 95th percentile. The underlying Statstore only provides one hardcoded percentile value as of now, the 95th percentile, due to a pending integration with CAdvisor's percentile derivations.

I can rename the struct field to NinetyFifth and consider reworking the StatStore implementation

afein:
closing this PR to split it into multiple ones and reimplement the StatStore.

PR #445, opened 2015-07-31
Added ssl certs to kube-config google.

piosz:
As requested in https://github.com/GoogleCloudPlatform/kubernetes/pull/12098

### DIFF
--- a/deploy/kube-config/google/heapster-controller.json
+++ b/deploy/kube-config/google/heapster-controller.json
@@ -24,6 +24,14 @@
 		}
 	    },
 	    "spec": {
+		"volumes": [
+		    {
+			"name": "ssl-certs",
+			"hostPath": {
+			    "path": "/etc/ssl/certs"
+			}
+		    }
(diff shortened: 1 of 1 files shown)


vishh:
ok to test

deploy/kube-config/google/heapster-controller.json
```diff
@@ -24,6 +24,14 @@
 		}
 	    },
 	    "spec": {
+		"volumes": [
+		    {
+			"name": "ssl-certs",
+			"hostPath": {
+			    "path": "/etc/ssl/certs"
```
jimmidyson: Does this work on non-debian based distros?

ScOut3R:
I have tried this on CoreOS and it does not work. I changed the path to "/usr/share/ca-certificates" and that works. I assume the symlinks under /etc/ssl/certs aren't resolvable inside the container.

vishh:
cc @mikedanese

PR #451, opened 2015-08-01
Adding events to cache.

vishh:
This PR will pave way for making the cache the central source of information for the API, model and sinks.

cc @afein

### DIFF
--- a/sinks/cache/cache_impl.go
+++ b/sinks/cache/cache_impl.go
@@ -15,6 +15,7 @@
 package cache
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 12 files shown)


vishh:
Unit tests are yet to be updated.

vishh:
Ping @afein @saad-ali

sinks/cache/cache_impl.go
```diff
@@ -225,10 +230,45 @@ func (rc *realCache) GetFreeContainers(start, end time.Time) []*ContainerElement
...
+	}
+	return result
+}
+
 func NewCache(bufferDuration time.Duration) Cache {
 	return &realCache{
 		pods:           make(map[string]*podElement),
 		nodes:          make(map[string]*nodeElement),
+		events:         store.NewGCStore(store.NewTimeStore(), bufferDuration),
+		eventUIDs:      make(map[string]struct{}),
```
saad-ali: What map to a struct and not a boolean?

mvdan: @saad-ali an empty struct takes up no memory. The only reason why using a boolean would be better is if you wanted to have three states instead of two, which isn't really necessary for a set.

saad-ali: Ah, TIL. Thanks.

afein:
LGTM

vishh:
Thanks for the review @afein

PR #453, opened 2015-08-04
Fix cache garbage collection 

### LINKED ISSUE #446: Pods that no longer exist not being garbage collected from v1 HTTP API
I'll look into the root cause, but I'm seeing this behavior on 0.15.0. Any idea of whether it's been fixed since would be helpful.

@vishh @vmarmol

vishh:
Fixes #446.

cc @a-robinson @afein @mvdan

### DIFF
--- a/heapster.go
+++ b/heapster.go
@@ -37,7 +37,7 @@ var (
 	argPollDuration    = flag.Duration("poll_duration", 10*time.Second, "The frequency at which heapster will poll for stats")
 	argStatsResolution = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats. Acceptable values are in the range [1 second, 'poll_duration')")
 	argAlignStats      = flag.Bool("align_stats", false, "Whether to align timestamps of metrics to multiplicity of 'stats_resolution'")
-	argCacheDuration   = flag.Duration("cache_duration", 10*time.Minute, "The total duration of the historical data that will be cached by heapster.")
+	argCacheDuration   = flag.Duration("cache_duration", 5*time.Minute, "The total duration of the historical data that will be cached by heapster.")
 	argUseModel        = flag.Bool("use_model", false, "When true, the internal model representation will be used")
 	argModelResolution = flag.Duration("model_resolution", 2*time.Minute, "The resolution of the timeseries stored in the model. Applies only if use_model is true")
 	argPort            = flag.Int("port", 8082, "port to listen to")
@@ -86,7 +86,7 @@ func validateFlags() error {
 }
 
 func doWork() ([]source_api.Source, sinks.ExternalSinkManager, manager.Manager, error) {
-	c := cache.NewCache(*argCacheDuration)
+	c := cache.NewCache(*argCacheDuration, time.Minute)
 	sources, err := newSources(c)
 	if err != nil {
 		return nil, nil, nil, err
--- a/model/impl_test.go
+++ b/model/impl_test.go
@@ -23,6 +23,7 @@ import (
 	cadvisor "github.com/google/cadvisor/info/v1"
 	fuzz "github.com/google/gofuzz"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/GoogleCloudPlatform/heapster/sinks/cache"
 	source_api "github.com/GoogleCloudPlatform/heapster/sources/api"
@@ -560,7 +561,7 @@ func TestUpdate(t *testing.T) {
 	mem_work_ts = *(ns.Metrics[memWorking])
 	actual = mem_work_ts.Get(zeroTime, zeroTime)
 	assert.Len(actual, 1)
-	assert.Equal(actual[0].Value.(uint64), uint64(1204))
+	assert.Equal(actual[0].Value.(uint64), uint64(2408))
 
 	// Invocation with no fresh data - expect no change in cluster
 	assert.NoError(cluster.Update(source_cache))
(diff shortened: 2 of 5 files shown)


brendandburns:
LGTM.

sinks/cache/cache_impl.go
```diff
@@ -264,11 +306,13 @@ func (rc *realCache) GetEvents(start, end time.Time) []*Event {
...
 func NewCache(bufferDuration time.Duration) Cache {
-	return &realCache{
+	rc := &realCache{
 		pods:           make(map[string]*podElement),
 		nodes:          make(map[string]*nodeElement),
 		events:         store.NewGCStore(store.NewTimeStore(), bufferDuration),
 		eventUIDs:      make(map[string]struct{}),
 		bufferDuration: bufferDuration,
 	}
+	go util.Until(rc.runGC, time.Second, util.NeverStop)
```
a-robinson: Every second seems way too often to run GC that only garbage collects things on the order of minutes old. Can you increase the period between runs? Every minute seems like it'd be more than enough, since holding data an extra minute shouldn't pose much risk.

If you have this set so low for the sake of testing, then just make it configurable and set it lower in testing.

sinks/cache/cache_impl_test.go
```diff
@@ -37,9 +37,30 @@ func TestFuzz(t *testing.T) {
 	assert := assert.New(t)
 	assert.NoError(cache.StorePods(pods))
 	assert.NoError(cache.StoreContainers(containers))
+	time.Sleep(5 * time.Second)
+	zeroTime := time.Time{}
+	assert.NotEmpty(cache.GetPods(zeroTime, zeroTime))
```
a-robinson: Did you remove the assert.NotEmpty(cache.GetFreeContainers(zeroTime, zeroTime)) line by accident?

sinks/cache/cache_impl.go
a-robinson: Or base it on the buffer duration

sinks/cache/cache_impl.go
```diff
@@ -21,16 +21,19 @@ import (
...
 const rootContainer = "/"
 
 type containerElement struct {
+	lastUpdated time.Time
 	Metadata
 	metrics store.TimeStore
 }
 
 type podElement struct {
+	lastUpdated time.Time
```
a-robinson: I assume you probably already checked, but can you please double check that this is set  appropriately everywhere that the pod/container elements are updated?

a-robinson: Including when it's initially created

heapster.go
```diff
@@ -37,7 +37,7 @@ var (
 	argPollDuration    = flag.Duration("poll_duration", 10*time.Second, "The frequency at which heapster will poll for stats")
 	argStatsResolution = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats. Acceptable values are in the range [1 second, 'poll_duration')")
 	argAlignStats      = flag.Bool("align_stats", false, "Whether to align timestamps of metrics to multiplicity of 'stats_resolution'")
-	argCacheDuration   = flag.Duration("cache_duration", 10*time.Minute, "The total duration of the historical data that will be cached by heapster.")
```
a-robinson: Does this need to be part of the cherry-pick? It'd be ideal to keep unrelated changes to a minimum.

sinks/cache/cache_impl.go
vishh: Good point! Manually controlling gc duration makes sense. Updated code.

sinks/cache/cache_impl_test.go
vishh: No. Fuzz might not be generating free containers. So its flaky as of now.

sinks/cache/cache_impl.go
vishh: The unit tests are supposed to cover the corner cases. I intend to add an e2e test for this. But with the time crunch, that might not happen.

heapster.go
vishh: This will reduce the memory usage. I'd recommend taking this in. WDYT?

a-robinson: Has memory usage been an issue that needs fixing?

sinks/cache/cache_impl.go
a-robinson: But you did manually check in place of a test, then, right?

heapster.go
vishh: Users have reported that memory usage being a concern. This is an easy way to mitigate that. Moreover we don't have to store stats for more than 3 minutes in reality, assuming we scrape it faster than that.

a-robinson:
Are there any potential problems if a container is deleted from the cache while it still exists? Say, for example, heapster fails to reach a node for 5 minutes. Will it continue trying to collect that data by virtue of what it knows from the apiserver, or is this cache considered the source of truth on what needs to be scraped?

heapster.go
a-robinson: Sure, we can keep it in. I just want to minimize the risk running into bugs due to some other component unexpectedly depending on things being around for 10 minutes, or a cache array having some minimum number of data points.

vishh:
It will continue to talk to the api server and recreate pods and containers once the node is online. cache is merely a routing entity.

a-robinson:
Perfect, thanks. This LGTM as long as you're confident that lastUpdatedTime is being tracked everywhere it needs to be.

model/impl_test.go
```diff
@@ -1194,7 +1197,7 @@ func cacheFactory() cache.Cache {
 				Hostname:  "hostname3",
 				Status:    "Running",
 			},
-			Containers: []source_api.Container{},
+			Containers: containers,
```
afein: The changes in this file do not appear to have any impact on unit test success or coverage of either the cache or the model. Was this change made for the sanity of not having an empty pod?

If so, the comments of cacheFactory need to be updated as well. I can do this myself in a follow-up PR, if needed.

PS: the pod was originally empty to test pod-level error-handling branches. However, further unit tests have covered these branches so there is no impact from this change.

afein:
LGTM, just one comment (no need to address it atm)

vishh:
Thanks for clarifying @afein. Will address it in a separate PR.

PR #456, opened 2015-08-07
Add image label to container metrics

mwielgus:


mwielgus:
cc: @piosz

sinks/api/v1/supported_labels.go
```diff
@@ -54,12 +54,17 @@ var (
 		Key:         "host_id",
 		Description: "Identifier specific to a host. Set by cloud provider or user",
 	}
+	LabelImage = LabelDescriptor{
+		Key:         "image",
+		Description: "User-defined image name that is run inside the container",
```
vishh: How about "Container base image" instead?

sinks/api/decoder.go
```diff
@@ -98,6 +98,8 @@ func (self *defaultDecoder) getContainerMetrics(container *source_api.Container,
 		return nil
 	}
 	labels[sinksV1Api.LabelContainerName.Key] = container.Name
+	labels[sinksV1Api.LabelImage.Key] = container.Image
```
vishh: Can you add this in sinks/api/v1/decoder.go as well? The code is in a state of flux. Thats the reason for duplication.

vishh:
Just a few comments. Otherwise LGTM

vishh:
ok to test

sinks/api/v1/supported_labels.go
mwielgus: Done

sinks/api/decoder.go
mwielgus: Added here and to the cache.

vishh:
LGTM. Thanks for the PR @mwielgus

PR #442, opened 2015-07-30
Updating Grafana to 2.0.

vishh:
Tested this manually and it works.

PR #457, opened 2015-08-07
Remove merge artifacts from heapster-controller.json

mwielgus:
Introduced in #442


vishh:
ok to test

vishh:
Thanks for the cleanup @mwielgus :+1:

PR #459, opened 2015-08-09
Go generate fixes

mvdan:


mvdan:
@vishh `hooks/check_boilerplate.sh` complains about the generated file not having the copyright. It's a generated file, so I'm thinking we should make the script ignore that file. Thoughts?

mvdan:
As said above, made `hooks/check_boilerplate.sh` ignore go generated files, which is only `*extpoints/extpoints.go` for now.

vishh:
Excluding generated files SGTM. Why not delete and regenerate extpoints? With this PR will extpoints.go be regerated whenever that package is modified?

mvdan:
See my commit messages - `go generate` is not supposed to be run by the user, but by the developer(s).

When what package is modified, `go-extpoints` or `heapster`? At the moment, neither. It is up to whomever changes the `go-extpoints` input to re-run the tool. And if the tool ever changes, which it shouldn't do often, someone would eventually run the updated tool and notice.

I thought about making travis run `go generate` and check if it is up to date. I could bundle that with a `go-extpoints` in `Godep/` so that it never breaks in the future. What do you think?

Side note: master currently has this problem. If `go-extpoints` breaks, heapster breaks, because we just `go get` it via master.

vishh:
I meant chnages to heapster extpoints/interface.go which should result in
regenerating extpoints/ expoints.go. Vendoring go-extpoints makes sense.

mvdan:
I'll vendor `go-extpoints` and add the thing to travis.

mvdan:
retest this please

mvdan:
retest this please

mvdan:
@vishh any idea why jenkins failed twice with the same error message?

mvdan:
Otherwise this is ready to be merged

vishh:
retest this please

mvdan:
The new jenkins build failed, but an earlier one succeeded. Jenkins is not being very deterministic.

vishh:
retest this please

mvdan:
green again!

PR #461, opened 2015-08-10
Time and value-compressible statStore

afein:
\cc @vishh

This statStore can be configured to make tradeoffs between precision and memory consumption though the epsilon and resolution arguments.

### DIFF
--- a/store/stat_store.go
+++ b/store/stat_store.go
@@ -1,0 +1,442 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package store
+
+import (
+	"container/list"
+	"fmt"
+	"math"
+	"sort"
+	"sync"
+	"time"
+)
+
+func NewStatStore(epsilon uint64, resolution time.Duration, windowDuration uint8, supportedPercentiles []float64) *statStore {
+	return &statStore{
+		buffer:               list.New(),
+		epsilon:              epsilon,
+		resolution:           resolution,
+		windowDuration:       windowDuration,
+		supportedPercentiles: supportedPercentiles,
+	}
+}
+
+// statStore is an in-memory rolling timeseries buffer that allows the extraction of -
+// segments of the stored timeseries and derived stats.
+// The tradeoff between space and precision can be configured through epsilon and resolution.
+// statStore only retains information about the latest TimePoints that are within its
+// specified duration.
+// statStore is a TimeStore-like object that does not implement the TimeStore interface.
+type statStore struct {
+	// start is the start of the time window.
+	start time.Time
+
+	// buffer is a list of tpBucket that is sequenced in a time-descending order, meaning that
+	// Front points to the latest tpBucket and Back to the oldest one.
+	buffer *list.List
+
+	// A RWMutex guards all operations on the underlying buffer.
+	sync.RWMutex
+
+	// epsilon is the acceptable error difference for the storage of TimePoints.
+	// Increasing epsilon decreases the memory usage of the statStore, at the cost of precision.
+	// The precision of max is not affected.
+	epsilon uint64
+
+	// resolution is the standardized duration between points in the statStore.
+	// the Get operation returns TimePoints at every multiple of resolution,
+	// even if TimePoints have not been Put for all such times.
+	resolution time.Duration
+
+	// windowDuration is the maximum number of time resolutions that is stored in the statStore
+	// e.g. windowDuration 60, with a resolution of time.Minute represents an 1-hour window
+	windowDuration uint8
+
+	// tpCount is the number of TimePoints that are represented in the statStore.
+	// if tpCount is equal to windowDuration, then the statStore window is considered full.
+	tpCount uint8
+
+	// lastPut maintains the state of values inserted within the last resolution.
+	// When values of a later resolution are added to the statStore, lastPut is flushed to
+	// the appropriate tpBucket.
+	lastPut putState
+
+	// suportedPercentiles is a slice of values from (0,1) that represents the percentiles
+	// that are calculated by the statStore.
+	supportedPercentiles []float64
+
+	validCache bool
+
+	// cachedAverage, cachedMax and cachedPercentiles are the cached derived stats that -
+	// are exposed by the StatStore.
+	cachedAverage     uint64
+	cachedMax         uint64
+	cachedPercentiles []uint64
+}
+
+// tpBucket is a bucket that represents a set of consecutive TimePoints with different
+// timestamps whose values differ less than epsilon.
+// tpBucket essentially represents a time window with a constant value.
+type tpBucket struct {
+	// count is the number of TimePoints represented in the tpBucket.
+	count uint8
+
+	// value is the approximate value of all in the tpBucket, +- epsilon.
+	value uint64
+
+	// max is the maximum value of all TimePoints that have been used to generate the tpBucket.
+	max uint64
+
+	// maxIdx is the number of resolutions after the start time where the max value is located.
+	maxIdx uint8
+}
+
+// putState is a structure that maintains context of the values in the resolution that is currently
+// being inserted.
+// Assumes that Puts are performed in a time-ascending order.
+type putState struct {
+	actualCount uint16
+	average     float64
+	max         uint64
+	stamp       time.Time
+}
+
+func (ss *statStore) Put(tp TimePoint) error {
+	ss.Lock()
+	defer ss.Unlock()
+
+	ss.validCache = false
+
+	// Flatten timestamp to the last multiple of resolution
+	ts := tp.Timestamp.Truncate(ss.resolution)
+
+	lastPutTime := ss.lastPut.stamp
+
+	// Handle the case where the buffer and lastPut are both empty
+	if lastPutTime.Equal(time.Time{}) {
+		ss.resetLastPut(ts, tp.Value.(uint64))
+		return nil
+	}
+
+	if ts.Before(lastPutTime) {
+		// Ignore TimePoints with Timestamps in the past
+		return fmt.Errorf("the provided timepoint has a timestamp in the past")
+	}
+
+	if ts.Equal(lastPutTime) {
+		// update lastPut with the new TimePoint
+		newVal := tp.Value.(uint64)
+		if newVal > ss.lastPut.max {
+			ss.lastPut.max = newVal
+		}
+		oldAvg := ss.lastPut.average
+		n := float64(ss.lastPut.actualCount)
+		ss.lastPut.average = (float64(newVal) + (n * oldAvg)) / (n + 1)
+		ss.lastPut.actualCount++
+		return nil
+	}
+
+	// new point is in the future, lastPut needs to be flushed to the statStore.
+
+	// Determine how many resolutions in the future the new point is at.
+	// The statStore always represents values up until 1 resolution from lastPut.
+	numRes := uint8(0)
+	curr := ts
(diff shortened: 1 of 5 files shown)


store/stat_store.go
```diff
@@ -0,0 +1,214 @@
...
+// statStore only retains information about the latest TimePoints that are within its
+// specified duration.
+// statStore is a TimeStore-like object that does not implement the TimeStore interface.
+type statStore struct {
+	// start is the start of the time window.
+	start time.Time
+
+	// buffer is a list of tpEntry that is sequenced in a time-descending order, meaning that
+	// Front points to the latest tpEntry and Back to the oldest one.
+	buffer *list.List
```
vishh: Take a look at this library: https://github.com/jbrukh/window

afein: this library assumes that the window size is a constant number of elements. In this case, we want an arbitrary number of tpEntries whose count fields sum up to the window size. For example, for a 1-minute resolution, 1-hour duration statStore, we might have just one tpEntry with a count of 60, if the value of the metric has not changed more than epsilon, on average.

Due to that, our corresponding rewind operation would subtract values from the count of the earliest tpEntry, and it would be removed only if the count is zero.

store/stat_store.go
```diff
@@ -0,0 +1,214 @@
...
+	zeroTime := time.Time{}
+	result := []TimePoint{}
+	skipped := 0
+	for elem := ss.buffer.Front(); elem != nil; elem = elem.Next() {
+		entry := elem.Value.(tpEntry)
+
+		// calculate the start time of the entry
+		entryStart := ss.start.Add((ss.tpCount - skipped - entry.count + 1) * ss.resolution)
+
+		// ignore tpEntries later than the requested end time
```
vishh: Does this mean you are storing in unsorted fashion?

store/stat_store.go
```diff
@@ -0,0 +1,214 @@
...
+
+		// set lastPut to the new time and value
+		ss.lastPut.lastPutTime = ts
+		ss.lastPut.actualCount = 1
+		ss.lastPut.average = tp.Value.(uint64)
+	}
+
+}
+
+// Get generates a []TimePoint from the appropriate tpEntries.
```
vishh: Why is this method needed?

store/stat_store.go
```diff
@@ -0,0 +1,323 @@
...
+// Get receives a start and end time as inputs.
+func (ss *statStore) Get(start, end time.Time) []TimePoint {
+	ss.RLock()
+	defer ss.RUnlock()
+	if ss.buffer.Len() == 0 {
+		return nil
+	}
+
+	zeroTime := time.Time{}
+	result := []TimePoint{}
```
mvdan: nit: use `var result []TimePoint` to avoid allocation if no elements are to be added

store/stat_store.go
```diff
@@ -0,0 +1,323 @@
...
+	// Create a new bucket if the buffer is empty
+	if ss.buffer.Front() == nil {
+		ss.newBucket()
+		ss.resetLastPut(ts, tp.Value.(uint64))
+		return nil
+	}
+
+	// Delete the earliest represented TimePoint if the window is full
+	if ss.tpCount == ss.windowDuration {
+		rewind()
```
mvdan: surely this must be `ss.rewind()`?

store/stat_store.go
```diff
@@ -0,0 +1,323 @@
...
+		ss.resetLastPut(ts, tp.Value.(uint64))
+	}
+
+	if ts.Equal(lastPutTime) {
+		// update lastPut with the new TimePoint
+		newVal := tp.Value.(uint64)
+		if newVal > ss.lastPut.max {
+			ss.lastPut.max = newVal
+		}
+		oldAvg = ss.lastPut.average
```
mvdan: `:=`? (assuming you haven't tried compiling any of this now)

afein:
yes, please don't nit pick at this time, the code does not compile yet. This PR was opened mostly to initiate discussion. I will remove the (WIP) prefix when it's ready for an actual review

mvdan:
Right, didn't know.

mvdan:
retest this please

mvdan:
@vishh I think jenkins and travis have broken since the switch from GoogleCloudPlatform/ to kubernetes/.

CC @mikedanese

store/stat_store.go
```diff
@@ -0,0 +1,378 @@
...
+	// Update the start time of the statStore
+	ss.start = ss.start.Add(ss.resolution)
+}
+
+// Get generates a []TimePoint from the appropriate tpEntries.
+// Get receives a start and end time as inputs.
+func (ss *statStore) Get(start, end time.Time) []TimePoint {
+	ss.RLock()
+	defer ss.RUnlock()
+	result := []TimePoint{}
```
mvdan: nit: `var result []Timepoint`

afein:
Implementation is finished. Feel free to review as normal.

\cc @vishh @mvdan

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+import (
+	"container/list"
+	"fmt"
+	"math"
+	"sort"
+	"sync"
+	"time"
+)
+
+func NewStatStore(e uint64, r time.Duration, w uint8, sp []float64) *statStore {
```
mvdan: Without looking at the code, it's hard to tell what each argument does. And in generated docs, since the fields are not exported, it will be for the user to guess.

I suggest something like `NewStatStore(epsilon uint64, resolution time.Duration, duratoin uint8, percentiles []float64)`

mvdan: An alternative would be to keep the current ones and add a proper comment explaining what each argument is.

afein: totally agree, i'll make the names more explicit and add comments as well.

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+}
+
+// statStore is an in-memory rolling timeseries buffer that allows the extraction of -
+// segments of the stored timeseries and derived stats.
+// The tradeoff between space and precision can be configured through epsilon and resolution.
+// statStore only retains information about the latest TimePoints that are within its
+// specified duration.
+// statStore is a TimeStore-like object that does not implement the TimeStore interface.
+type statStore struct {
```
vishh: Why is this private?

mvdan: +1

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+		count:  numRes,
+		value:  ((uint64(ss.lastPut.average) / ss.epsilon) + 1) * ss.epsilon,
+		max:    ss.lastPut.max,
+		maxIdx: 0,
+	}
+	ss.buffer.PushFront(newEntry)
+	ss.tpCount += numRes
+
+	// If this was the first bucket, update ss.start
+	if ss.start.Equal(time.Time{}) {
```
mvdan: comparing to an empty time seems wrong to me... are you sure you cannot do `if !ss.start {` if that is its zero value?

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+	for i, spc := range ss.supportedPercentiles {
+		if p == spc {
+			found = true
+			idx = i
+			break
+		}
+	}
+
+	if !found {
+		return uint64(0), fmt.Errorf("the requested percentile is not supported")
```
mvdan: Could it be that due to a loss of precision you end up with this error when you shouldn't?

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+	// Front points to the latest tpBucket and Back to the oldest one.
+	buffer *list.List
+
+	// A RWMutex guards all operations on the underlying buffer.
+	sync.RWMutex
+
+	// epsilon is the acceptable error difference for the storage of TimePoints.
+	// Increasing epsilon decreases the memory usage of the statStore, at the cost of precision.
+	// The precision of max is not affected.
+	epsilon uint64
```
mvdan: Why did you choose to make epsilon `uint64` instead of `float64`?

store/stat_store.go
afein: I don't believe that could be the case. Percentiles are declared in the 0-1 range. If someone is requesting the 95.4853th percentile then it might be a problem but for two decimal digits this should be accurate

store/stat_store.go
afein: epsilon represents error margin in value. Since all TimePoint values are assumed to be uint64, then having a floating point epsilon makes no sense. For example, setting epsilon to 50.53412 would effectively be the same as setting it to 50. Also, epsilon cannot be a percentage offset as that would cause bucket overlaps. For example, a bucket with average value of 100 and epsilon of 0.1 would overlap with a bucket with value of 110.

store/stat_store.go
afein: agree, will address.

store/stat_store.go
afein: invalid operation: ! time.Time

afein:
Addressed comments. Another commit with further documentation will be added soon.

store/stat_store.go
mvdan: Actually, for strings you do `if s != ""`. Maybe `if ss.start != time.Time{}`? Although it's essentially the same.

store/stat_store.go
mvdan: :+1:

mvdan:
With @vishh we were just commenting that it would be a good idea to move this into a separate package so that kubelet could use it without having to import all of `heapster/store`. What do you think?

Otherwise LGTM.

store/stat_store.go
afein: Time struct equality is only supported through the .Equal method of time.Time, as the method has to compare each field of the two structs

afein:
moved to a separate package and added more documentation

afein:
rebased, no further changes pending on my part

mvdan:
If you squash and fix Travis, it's fine by me.

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+		return nil
+	}
+
+	lastEntry := ss.buffer.Front().Value.(tpBucket)
+	lastAvg := ss.lastPut.average
+
+	// Place lastPut in the latest bucket if the difference from its average
+	// is less than epsilon
+	if uint64(math.Abs(float64(lastEntry.value)-lastAvg)) < ss.epsilon {
+		lastEntry.count += numRes
```
vishh: Why add numRes here? IIUC, we did not get any data points for the duration between lastPut and current timestamp?

store/stat_store.go
```diff
@@ -0,0 +1,442 @@
...
+		// update in list
+		ss.buffer.Front().Value = lastEntry
+	} else {
+		// Create a new bucket
+		ss.newBucket(numRes)
+	}
+
+	// Delete the earliest represented TimePoints if the window is full
+	for ss.tpCount > ss.windowDuration {
+		ss.rewind()
```
vishh: nit: Add a todo to consider garbage collecting outside of Put as well.

store/stat_store.go
afein: As we discussed offline, absence of data points is assumed as the value remaining constant at the last bucket. If an actual deletion/shutdown has happened, a higher-level event handler should take care of that and insert zero-valued timepoints.

store/statstore/stat_store.go
```diff
@@ -0,0 +1,486 @@
...
+func (ss *StatStore) rewind() {
+	firstElem := ss.buffer.Back()
+	firstEntry := firstElem.Value.(tpBucket)
+	// Decrement number of TimePoints in the earliest tpBucket
+	firstEntry.count--
+	// Decrement total number of TimePoints in the StatStore
+	ss.tpCount--
+	if firstEntry.maxIdx == 0 {
+		// The Max value was just removed, lose precision for other maxes in this bucket
+		firstEntry.max = firstEntry.value
```
vishh: Why not calculate the new max?

store/statstore/stat_store.go
```diff
@@ -0,0 +1,486 @@
...
+		firstEntry.maxIdx = firstEntry.count - 1
+	} else {
+		firstEntry.maxIdx--
```
vishh: Why move the maxIdx here? Can you add a small example as comments here?

store/statstore/stat_store.go
afein: this is the bucket max, and we cannot recalculate it as we have lost
information of the other maxes in that bucket

vishh:
LGTM.

afein:
retest this please

PR #462, opened 2015-08-10
Adding container base image label to metrics API output.

vishh:


vishh:
Merging this to fix tests.

PR #467, opened 2015-08-13
Get aggregated/summed metric for pods

mwielgus:


mwielgus:
cc:  vishh

mwielgus:
cc: @vishh

api/v1/model_handlers.go
```diff
@@ -272,6 +274,20 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(MetricResult{}))
 
+	// The /sum/namespaces/{namespace-name}/pods/{pod-name}/metrics/{metric-name} endpoint exposes
+	// an aggregated metric for a list od pods of the model.
+	ws.Route(ws.GET("/sum/namespaces/{namespace-name}/pods/{pod-list}/metrics/{metric-name}").
```
afein: The existing API structure is meant to resemble the Kubernetes API, by having subpaths that indicate increased specificity in the cluster structure. 

Because of that, something like /namespaces/{namespace-name}/pods/{pod-list}/metrics/{metric-name} would be cleaner. To implement that, we could change the existing handler for that path to attempt to extract a comma-separated list of podnames.

api/v1/model_handlers.go
```diff
@@ -474,6 +490,76 @@ func (a *Api) podMetrics(request *restful.Request, response *restful.Response) {
...
+func (s AggregatedMetricPoints) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+func (s AggregatedMetricPoints) Less(i, j int) bool {
+	// we want the descending order
+	return s[j].Timestamp.Before(s[i].Timestamp)
+}
+
+// podMetrics returns a metric timeseries for a metric of the Pod entity.
+func (a *Api) podMetricsSum(request *restful.Request, response *restful.Response) {
```
afein: This is a much more "fat" API handler than the rest of the model handlers, which does not consider model locking. Due to the consecutive calls to GetPodMetric, it is possible that a model Update operation may be performed between these calls, leading to potential concurrency issues. 
For example, if "end" is specified as zero, it is possible that some of the pods will return a []TimePoint with some timestamps further in the future than the result of other pods.

To avoid this issue, I would recommend following the pattern used in the other handlers, where a single model method is called by the handler. This would involve the creation of the following model method:

GetPodListMetric(PodListRequest) []store.TimePoint

type PodListRequest struct {
NamespaceName string
PodNames []string
MetricName: string
Start: time.Time
End: time.Time
}

afein:
Posted two primary concerns.

For future reference, I am currently in the process of refactoring some portions of the model and writing documentation, which might prove useful for future extensions such as this one.

PS: the Travis test that is failing is a flaky test.

api/v1/model_handlers.go
mwielgus: `/namespaces/{namespace-name}/pods/sum/metrics/{metric-name}` will also match to 
`/namespaces/{namespace-name}/pods/{pod-name}/metrics/{metric-name}` causing path ambiguity.

Reusing /namespaces/{namespace-name}/pods/{pod-name}/metrics/{metric-name} will require a change in the result type in this (an other paths for consistency). Are you OK with that?

api/v1/model_handlers.go
mwielgus: Ok, should I expand TimePoint or create a new type (sample count is needed for the sum to be meaningful)? Other option is to return something like `[][]*store.TimePoint`.

afein: For this case I would recommend creating a new struct in model/types.go that encapsulates a TimePoint and the count. 

For further reference, on the next release of Heapster the model will always return TimePoints at every resolution for all entities that are active in the cluster, so at that point the sample count might not be as meaningful.

In general, feel free to modify any external type you may need at api/v1/model_types.go .
However, please do not edit the TimePoint type or any existing types in model/types.go, as we are trying to minimize the total memory footprint of the model at this point.

api/v1/model_handlers.go
afein: You're right, the `/pods/sum` subpath could cause issues in case there is pod named "sum" in that namespace.
 
A change in external result types is perfectly fine for this change.

vishh: One more option - /namespaces/{namespace-name}/pods can return metrics for all pods in that namespace. The client can them perform the summation. Heapster can then focus only in distributions for all logical entities. WDYT?

vishh: If we go with the option I proposed, we should do the same for `/nodes` as well, but thats for a separate PR.

vishh:
Thanks for the PR @mwielgus!

api/v1/model_handlers.go
afein: At the moment, /namespaces/{namespace-name}/pods/ returns a list of all the pods, along with the latest cpu and memory usage values. If we change this behavior, then the API will not be fully browsable, as there would be no way for a consumer to know which pods are represented in the model.

mwielgus: There is no CPU/MEM information in /namespaces/{namespace-name}/pods/:
```
curl 10.0.98.141:80/api/v1/model/namespaces/kube-system/pods | gunzip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   237  100   237    0     0  69277      0 --:--:-- --:--:-- --:--:-- 79000
[
  "fluentd-cloud-logging-kubernetes-minion-yxir",
  "kube-ui-v1-8b1lb",
  "fluentd-cloud-logging-kubernetes-minion-ihlf",
  "fluentd-cloud-logging-kubernetes-minion-a5u3",
  "kube-controller-manager-kubernetes-master",
  "monitoring-influx-grafana-v1-s97to",
  "kube-scheduler-kubernetes-master",
  "kube-apiserver-kubernetes-master",
  "kube-dns-v8-vdsrr",
  "monitoring-heapster-v7-czp8j",
  "fluentd-cloud-logging-kubernetes-minion-9y8h",
  "etcd-server-kubernetes-master",
  "monitoring-heapster-v7-p9aze",
  "fluentd-cloud-logging-kubernetes-master",
  "monitoring-heapster-v7-iio5r",
  "monitoring-heapster-v7-r1ddh"
 ]
```

api/v1/model_handlers.go
mwielgus: In the end I decided to try with `[][]store.TimePoint`. It seems clearer to get all metrics and sum them elsewhere than summing them up inside the db, under lock. Please take a look.

api/v1/model_handlers.go
mwielgus: What about `/namespaces/{namespace-name}/pod-list/{pod-list}/metrics/{metric-name}`?

vishh: @mwielgus: I chatted with @afein offline and that discussion is captured here #476. The current API format is not well suited for adding more use cases. This PR is a classic example of that. 
Essentially if we had a `/api/v1.1/stats/namespaces/<namespace>/pods` to begin with which returned all the stats for pods under the given namespace, that should satisfy your use case, I hope. In the future, we can add support for label selectors as query params, which should help result in querying only a subset of pods. 
In general, +1 for doing client side summation for now. Once we understand the use cases more, we can consider moving operations to the server side. WDYT?

api/v1/model_handlers.go
vishh: +1 for returning a list of timepoints and performing summation at client side.

api/v1/model_handlers.go
mwielgus: For now we are fine with summation on the client (the PR is updated).

api/v1/model_handlers.go
```diff
@@ -272,6 +273,20 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(MetricResult{}))
 
+	// The /namespaces/{namespace-name}/pods/{pod-name}/metrics/{metric-name} endpoint exposes
+	// an aggregated metric for a list od pods of the model.
```
jszczepkowski: Please update the comment.

jszczepkowski:
CC @jszczepkowski

api/v1/model_handlers.go
mwielgus: Updated.

mwielgus:
Ping :)

vishh:
ok to test

vishh:
LGTM. Merging this for now, until a better the new APIs are added.

PR #468, opened 2015-08-13
Model API documentation

afein:
This PR documents the usage and API endpoints of the Heapster model. I plan to add one more documentation file on Heapster regarding contributions to the model later today.

\cc @vishh @mvdan

### DIFF
--- a/docs/model.md
+++ b/docs/model.md
@@ -28,28 +28,28 @@ then all data later than `start` will be returned.
 ### Cluster-level Metrics and Stats
 `/api/v1/model/`: The root path of the model API, displays all browsable subpaths.
 
-`/api/v1/model/metrics/`: Displays a list of available cluster-level metrics.
+`/api/v1/model/metrics/`: Returns a list of available cluster-level metrics.
 
-`/api/v1/model/metrics/{metric-name}?start=X&end=Y`: Displays a set of (Timestamp, Value) 
+`/api/v1/model/metrics/{metric-name}?start=X&end=Y`: Returns a set of (Timestamp, Value) 
 pairs for the requested cluster-level metric, between the time range specified by `start` and `end`. 
 
 `/api/v1/model/stats/`: Exposes the average, max and 95th percentile over the
 past minute, hour and day for each cluster metric.
 
 ### Node-level Metrics and Stats
-`/api/v1/model/nodes`: Displays a list of all available nodes, along
+`/api/v1/model/nodes`: Returns a list of all available nodes, along
 with their latest CPU and Memory Usage values.
 
-`/api/v1/model/nodes/{node-name}`: Displays all browsable subpaths for a
+`/api/v1/model/nodes/{node-name}`: Returns all browsable subpaths for a
 specific node.
 
-`/api/v1/model/nodes/{node-name}/pods`: Displays a list of all available pods
+`/api/v1/model/nodes/{node-name}/pods`: Returns a list of all available pods
 under a given node, along with their latest CPU and Memory Usage values
 
-`/api/v1/model/nodes/{node-name}/metrics/`: Displays a list of available
+`/api/v1/model/nodes/{node-name}/metrics/`: Returns a list of available
 node-level metrics
 
-`/api/v1/model/metrics/nodes/{node-name}/{metric-name}?start=X&end=Y`: Displays a set of (Timestamp, Value) 
+`/api/v1/model/metrics/nodes/{node-name}/{metric-name}?start=X&end=Y`: Returns a set of (Timestamp, Value) 
 pairs for the requested node-level metric, within the time range specified by `start` and `end`. 
 
 `/api/v1/model/nodes/{node-name}/stats/`: Exposes the average, max and 95th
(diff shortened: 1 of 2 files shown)


docs/model.md
```diff
@@ -0,0 +1,122 @@
+# Heapster Metric Model
+
+The Heapster Model is a structured representation of metrics for Kubernetes clusters, which is exposed through a set of REST API endpoints.
+It allows the extraction of up to 1 hour of historical data for any Container, Pod, Node or Namespace in the cluster, as well as the cluster itself.
```
vishh: Isn't it a day?

docs/model.md
```diff
@@ -0,0 +1,122 @@
...
+that represent the start and end time of the requested timeseries. The result
+will be a list of (Timestamp, Value) pairs in the time range (start, end].
+`start` and `end` are strings formatted according to RFC3339. If `start` is not
+defined, it is assumed as the zero Unix epoch time. If `end` is not defined,
+then all data later than `start` will be returned.
+
+### Cluster-level Metrics and Stats
+`/api/v1/model/`: The root path of the model API, displays all browsable subpaths.
+
+`/api/v1/model/metrics/`: Displays a list of available cluster-level metrics.
```
vishh: nit: Suggest `Returns` in favor of `Displays`

docs/model.md
afein: not for historical data, meaning that you can't extract a timeseries for more than 1 hour ago.
As we discussed offline, the DayStore only provides derived stats using the hourly derived stats

docs/model.md
```diff
@@ -0,0 +1,122 @@
...
+
+`/api/v1/model/metrics/{metric-name}?start=X&end=Y`: Displays a set of (Timestamp, Value) 
+pairs for the requested cluster-level metric, between the time range specified by `start` and `end`. 
+
+`/api/v1/model/stats/`: Exposes the average, max and 95th percentile over the
+past minute, hour and day for each cluster metric.
+
+### Node-level Metrics and Stats
+`/api/v1/model/nodes`: Displays a list of all available nodes, along
```
vishh: Can we define a top level resources supported and then just use that resources keyword everywhere else?

vishh: For now this will be just 'cpu' and 'memory'

docs/model.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+### Namespace-level Metrics and Stats
+`/api/v1/model/namespaces`: Displays a list of all available namespaces, along
+with their latest CPU and Memory Usage values.
+
+`/api/v1/model/namespaces/{namespace-name}`: Displays all browsable subpaths for
+a specific namespace.
+
+`/api/v1/model/namespaces/{namespace-name}/metrics/`: Displays a list of available namespace-level metrics
```
vishh: What if the user wants all the metrics in a namespace?

docs/model.md
```diff
@@ -0,0 +1,122 @@
...
+
+`/api/v1/model/namespaces/{namespace-name}/metrics/{metric-name}?start=X&end=Y`: Displays a set of (Timestamp, Value) 
+pairs for the requested namespace-level metric, within the time range specified by `start` and `end`. 
+
+`/api/v1/model/namespaces/{namespace-name}/stats/`: Exposes the average, max and 95th percentile over the
```
vishh: I assume this is per resource. Mention that.

vishh:
Super useful PR! Thanks @afein! Much appreciated!

vishh:
cc @piosz

docs/model.md
afein: That is a more complicated use case, assuming you mean that they want exactly the same `start` and `end` times for each metric. 
As for now they can call each individual metric's API with the desired start and end.

To enrich the API, we could add a model method that extracts all metrics for an entity given the same `start` and `end` times

docs/model.md
afein: It is mentioned that these stats are exposed for each X metric, where X is {cluster, node, namespace, pod, container}. If that is not clear I can change the phrasing

docs/model.md
afein: could you elaborate? I am trying to be very explicit in the documentation and I'm not avoiding repetition. As latest CPU and Memory usage values are the only two things that are shown in the listing, I'm explicitly stating that.

docs/model.md
vishh: Ah! Got it.

docs/model.md
vishh: Yeah. You are stating that in every API. If we add another resource, it requires updating all the API  docs. Instead if we declare something like 'Supported Resources' and refer to that keyword in the rest of the doc, it might be easy to maintain in the long run. WDYT?

docs/model.md
vishh: Ok.

docs/model.md
vishh: `"/api/v1/model/namespaces/{namespace-name}/stats/": Exposes the average, max and 95th percentile over the past minute, hour and day for each namespace metric.`

There is no reference to the specific metrics like CPU and Memory here.

afein:
retest this please

piosz:
cc @mwielgus @jszczepkowski

PR #470, opened 2015-08-14
Updating base docker image to alpine.

vishh:
cc @mikedanese: I assume this will actually resolve the cert issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ deps:
 	go get github.com/tools/godep
 
 build: clean deps
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a ./...
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a ./... 
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a
 
 sanitize:
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -1,11 +1,12 @@
-FROM progrium/busybox
+FROM alpine:3.2
 MAINTAINER vishnuk@google.com
+
+RUN apk --update add curl ca-certificates tar && \
+    curl -Ls https://circle-artifacts.com/gh/andyshinn/alpine-pkg-glibc/8/artifacts/0/home/ubuntu/alpine-pkg-glibc/packages/x86_64/glibc-2.21-r2.apk > /tmp/glibc-2.21-r2.apk && \
+    apk add --allow-untrusted /tmp/glibc-2.21-r2.apk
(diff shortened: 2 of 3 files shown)


mikedanese:
but does it break dns-search paths? @thockin what should be aware of when switching to alpine?

jimmidyson:
Yes it will. Install glibc & you still get a small image but with expected DNS behaviour. See https://github.com/anapsix/docker-alpine-java/blob/master/Dockerfile#L8 for how to do it.

vishh:
Thanks for the tip @jimmidyson! Updated the PR. PTAL.

deploy/docker/Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM progrium/busybox
+FROM alpine:3.2
 MAINTAINER vishnuk@google.com
 
-# cAdvisor discovery via external files.
-VOLUME /var/run/heapster/hosts
-
-RUN opkg-install ca-certificates
+RUN apk --update add curl ca-certificates tar && \
+    curl -Ls https://circle-artifacts.com/gh/andyshinn/alpine-pkg-glibc/6/artifacts/0/home/ubuntu/alpine-pkg-glibc/packages/x86_64/glibc-2.21-r2.apk > /tmp/glibc-2.21-r2.apk && \
```
jimmidyson: I guess you haven't tried building this? That URL doesn't work IIRC. Try changing the 6 to 8. Relying on build artifacts isn't great but no other option right now if using alpine.

vishh: I did try building and running the image. The e2e test also builds it and runs it in a cluster.

jimmidyson: Interesting - failed for me, sorry. Most have been a local issue.

mikedanese:
LGTM.

deploy/docker/Dockerfile
vishh: No problem :) Updated to `8` as suggested.

vishh:
Merging this. @jimmidyson If there are any other issues, I will address them in a separate PR.

thockin:
I will take a look at this - how big is an Alpine+glibc base?  I am
concerned that people will see this and use alpines package manager which
is just not going to work.

Makefile
```diff
@@ -8,7 +8,7 @@ deps:
 	go get github.com/tools/godep
 
 build: clean deps
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a ./...
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a ./...
```
thockin: trailing whitespace....

PR #471, opened 2015-08-14
DayStore implementation

afein:
Please let me know what the appropriate policy is on adding modified third-party code. In this case, I had to modify https://github.com/jbrukh/window to support hourEntry structs instead of float64.

\cc @vishh @mvdan

### DIFF
--- a/store/daystore/day_store.go
+++ b/store/daystore/day_store.go
@@ -1,0 +1,234 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package store
+
+import (
+	"fmt"
+	"math"
+	"sort"
+	"sync"
+	"time"
+
+	statstore "k8s.io/heapster/store/statstore"
+
+	"k8s.io/heapster/third_party/window"
+)
+
+// DayStore is an in-memory window of derived stats for each hour of a day.
+// DayStore holds 24 hours of derived stats, plus an hour-long StatStore that holds -
+// historical data on the current hour.
+// DayStore can calculate the Average, Max and 95th Percentile over the past 24 hours.
+
+// Note on how derived stats are extracted:
+// If the DayStore holds less than 1 hour of data, then the average, max and 95th are -
+// calculated only from the the Hour store.
+//
+// If the DayStore holds more than 1 hour of data, then the average and 95th are calculated -
+// only from the 24 past hourly stats that have been collected.
+// Assuming a resolution of 1 minute, this behavior may ignore up to 59 minutes of the latest data.
+// This behavior is required to avoid calculating less than a full day's data.
+// In order to correctly capture spikes, the Max takes into account the latest data, causing the -
+// max to reflect values in the past [24 hours, 25 hours)
+type DayStore struct {
+	// a RWMutex guards all operations on the underlying window and cached values.
+	sync.RWMutex
+
+	// Hour is a StatStore with data from the last one hour
+	Hour *statstore.StatStore
+
+	window *window.MovingWindow
+
+	// size is the number of items currently stored in the window
+	size int
+
+	// lastFlush is the time when the previous hourly stats were flushed into the DayStore
+	lastFlush time.Time
+
+	// validAvgPct marks whether the cached average and percentiles are correct.
+	validAvgPct bool
+	// validMax marks whether the cached max value is correct.
+	validMax bool
+
+	cachedAverage     uint64
+	cachedMax         uint64
+	cachedNinetyFifth uint64
+}
+
+// hourEntry is the set of derived stats that are maintained per hour.
+type hourEntry struct {
+	average     uint64
+	max         uint64
+	ninetyFifth uint64
+}
+
+// NewDayStore is a DayStore constructor.
+// The recommended minimum resolution is at least one minute.
+func NewDayStore(epsilon uint64, resolution time.Duration) *DayStore {
+	// Calculate how many resolutions correspond to an hour
+	hourNS := time.Hour.Nanoseconds()
+	resNS := resolution.Nanoseconds()
+	intervals := uint8(hourNS / resNS)
+	if hourNS%resNS != 0 {
+		intervals++
+	}
+
+	return &DayStore{
+		window: window.New(24, 1),
+		Hour:   statstore.NewStatStore(epsilon, resolution, intervals, []float64{0.95}),
+	}
+}
+
+// Put stores a TimePoint into the Hour StatStore, while checking whether it -
+// is time to flush the last hour's stats in the window.
+func (ds *DayStore) Put(tp statstore.TimePoint) error {
+	ds.Lock()
+	defer ds.Unlock()
+
+	err := ds.Hour.Put(tp)
+	if err != nil {
+		return err
+	}
+
+	ds.validMax = false
+
+	// Check if this is the first TimePoint ever, in which case flush in one hour.
+	if ds.lastFlush.Equal(time.Time{}) {
+		ds.lastFlush = tp.Timestamp
+		return nil
+	}
+
+	// The new TimePoint is not newer by at least one hour since the last flush
+	if tp.Timestamp.Add(-time.Hour).Before(ds.lastFlush) {
+		return nil
+	}
+
+	// flush the current hour to the window.
+	ds.validAvgPct = false
+	avg, _ := ds.Hour.Average()
+	max, _ := ds.Hour.Max()
+	nf, _ := ds.Hour.Percentile(0.95)
+	newEntry := hourEntry{
+		average:     avg,
+		max:         max,
+		ninetyFifth: nf,
+	}
+	if ds.size < 24 {
+		ds.size += 1
+	}
+	ds.window.PushBack(newEntry)
+	ds.lastFlush = tp.Timestamp
+	return nil
+}
+
+// fillMax caches the max of the DayStore.
+func (ds *DayStore) fillMax() {
+	// generate a slice of the window
+	day := ds.window.Slice()
+
+	// calculate th max of the hourly maxes
+	curMax, _ := ds.Hour.Max()
+	for _, elem := range day {
+		he := elem.(hourEntry)
+		if he.max > curMax {
+			curMax = he.max
+		}
+	}
+	ds.cachedMax = curMax
+	ds.validMax = true
+}
+
+// fillAvgPct caches the average, 95th percentile of the DayStore.
+func (ds *DayStore) fillAvgPct() {
+	ds.validAvgPct = true
+
+	// If no past Hourly data has been flushed to the window,
+	// return the average and 95th percentile of the past hour.
+	if ds.size == 0 {
+		ds.cachedAverage, _ = ds.Hour.Average()
+		ds.cachedNinetyFifth, _ = ds.Hour.Percentile(0.95)
+		return
+	}
+	// Otherwise, ignore the past one hour and use the window values
+
+	// generate a slice of the window
+	day := ds.window.Slice()
+
+	// calculate the average value of the hourly averages
+	// also create a sortable slice of float64
+	var sum uint64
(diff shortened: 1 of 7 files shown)


afein:
I am unable to reproduce the same test failure as Travis on my machine. Any clues?

mvdan:
Careful there. You are relicensing code written by someone else.

mvdan:
Why are the tests passing now? Are the travis gods aware of license violations? :)

afein:
it was a flaky test: https://github.com/kubernetes/heapster/pull/315/files#diff-e4dfa1a125bfbc4332db2c3b31a2b471L150

store/daystore/day_store.go
```diff
@@ -0,0 +1,212 @@
...
+package store
+
+import (
+	"fmt"
+	"math"
+	"sort"
+	"sync"
+	"time"
+
+	ss "k8s.io/heapster/store/statstore"
```
mvdan: I'd probably leave the import as `statstore`

store/daystore/day_store.go
```diff
@@ -0,0 +1,212 @@
...
+
+	window "k8s.io/heapster/third_party/window"
```
mvdan: You don't need to override the import name here, right?

mvdan: Correction - you do because you named that package `store`. Please rename it `window` and remove this name override here.

store/daystore/day_store.go
```diff
@@ -0,0 +1,212 @@
...
+// DayStore can calculate the Average, Max and 95th Percentile over the past 24 hours.
+// The latest one hour is only considered for the calculation of the daily Max.
+type DayStore struct {
+	// a RWMutex guards all operations on the underlying window and cached values.
+	sync.RWMutex
+
+	// Hour is a StatStore with data from the last one hour
+	Hour *ss.StatStore
+
+	window *window.MovingWindow
```
mvdan: Why use a pointer to a window? The structure is light.

store/daystore/day_store.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Check if this is the first TimePoint ever, in which case flush in one hour.
+	if ds.lastFlush.Equal(time.Time{}) {
+		ds.lastFlush = tp.Timestamp
+		return nil
+	}
+
+	// The new TimePoint is newer by at least one hour since the last flush
+	// flush the current hour to the window.
+	if !tp.Timestamp.Add(-time.Hour).Before(ds.lastFlush) {
```
mvdan: Looks like this is the normal flow of the code. How about inverting the `if`, making it return and then moving the rest of the code out of the block?

store/daystore/day_store.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// fillAvgPct caches the average, 95th percentile of the DayStore.
+func (ds *DayStore) fillAvgPct() {
+	// generate a slice of the window
+	day := ds.window.Slice()
+
+	// calculate the average value of the hourly averages
+	var sum uint64
+	nf := []float64{}
```
mvdan: nit: var to avoid allocation

third_party/window/window.go
```diff
@@ -0,0 +1,87 @@
...
+// popped.
+//
+// You can optimize the amount of slice copying
+// that the MovingWindow will be doing by trading
+// off with space complexity. Namely, the underlying
+// array is allocated with a size that is the
+// multiple of the intended capacity of the queue
+// so that copying is less frequent.
+type MovingWindow struct {
+	arr  []interface{}
```
mvdan: If we're only going to be using this for the StatStore, you might want to use the specific type that we need. The only reason why would be so that we can use `[]hourEntry`, which is contiguous in memory, instead of `[]interface{}` which uses pointers under the covers.

In general, unless each element is very large and thus expensive to swap/move/copy, it's better to store its value directly.

mvdan: Building on that - this code is rather simple, we could probably reimplement it for our needs under our own license in the `store/daystore` package with the usage of `[]hourEntry` instead of `[]interface{}`.

store/daystore/day_store.go
afein: fixed

store/daystore/day_store.go
afein: changed

third_party/window/window.go
afein: using []hourEntry would cause the daystore implementation to be fragmented to the third party package, for the definition of hourEntry. interface{} is cleaner

mvdan: That's why I said that we could write an implementation of our own to suit our needs :)

Anyway, I could do this if needed if I encounter performance issues in the future.

store/daystore/day_store_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+package store
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	ss "k8s.io/heapster/store/statstore"
```
mvdan: Could you change the import here too?

third_party/window/window.go
afein: that was the initial plan but apparently this library performs really well. If the interface{} causes it to slow down we might as well scrap it for a ring buffer. \cc @vishh

vishh: @mvdan: The use of interfaces in our use case, will result in copying pointers only. If we store structures, we will end up copying structures between regions of heap. WDYT?

vishh: +1 for switching over to our own impl if we have issues with this.

mvdan: Copying structures is only a problem if they are noticeably heavier than a pointer. In this case, `hourEntry` is three `uint64`s. Even if assuming the worst - a 32bit system - that is far from a large struct.

The data being compact and local in memory is very important. You should only give those up if the structure is very heavy.

But as said, I'm fine with using this implementation for now and rewriting it later myself if I find that this is a problem.

store/daystore/day_store_test.go
afein: ETA: 3 hours

store/daystore/day_store.go
mvdan: Bump! This comment got hidden in an outdated diff.

store/daystore/day_store_test.go
mvdan: Wait, so should this be closed in favour of #478?

afein: No, #478 uses this daystore.

afein:
ping @mvdan @vishh
unless there are any nits, this PR is ready to merge

store/daystore/day_store.go
```diff
@@ -0,0 +1,224 @@
...
+	statstore "k8s.io/heapster/store/statstore"
+
+	"k8s.io/heapster/third_party/window"
+)
+
+// DayStore is an in-memory window of derived stats for each hour of a day.
+// DayStore holds 24 hours of derived stats, plus an hour-long StatStore that holds -
+// historical data on the current hour.
+// DayStore can calculate the Average, Max and 95th Percentile over the past 24 hours.
+// The latest one hour is only considered for the calculation of the daily Max.
```
vishh: Why is that so? The max should be inclusive of the maximum during the entire day right?

store/daystore/day_store.go
```diff
@@ -0,0 +1,224 @@
...
+	ds.validAvgPct = true
+
+	// If no past Hourly data has been flushed to the window,
+	// return the average and 95th percentile of the past hour.
+	if ds.size == 0 {
+		ds.cachedAverage, _ = ds.Hour.Average()
+		ds.cachedNinetyFifth, _ = ds.Hour.Percentile(0.95)
+		return
+	}
+	// Otherwise, ignore the past one hour and use the window values
```
vishh: Why?

store/daystore/day_store.go
```diff
@@ -0,0 +1,224 @@
...
+	// calculate the average value of the hourly averages
+	// also create a sortable slice of float64
+	var sum uint64
+	var nf []float64
+	for _, elem := range day {
+		he := elem.(hourEntry)
+		sum += he.average
+		nf = append(nf, float64(he.ninetyFifth))
+	}
+	ds.cachedAverage = sum / uint64(ds.size)
```
vishh: where do you set validAvgPct?

store/daystore/day_store_test.go
```diff
@@ -0,0 +1,437 @@
...
+	assert.Equal(max, maxHr)
+
+	// The Ninetyfifth percentile does not take the last hour into account
+	nf, err = ds.NinetyFifth()
+	assert.NoError(err)
+	nfHr, err := ds.Hour.Percentile(0.95)
+	assert.NoError(err)
+	assert.Equal(nf, nfHr)
+
+	// Put in 11 more hours of data.
```
vishh: nit: Can we do this in a loop?

third_party/window/window.go
vishh: +1 for rewriting if there is a performance issue.

vishh: SUggestion: How about storing pointers as floats? Will that help us avoid having to fork this package?

vishh:
@afein: Just a few comments. Overall LGTM! Good work!

third_party/window/window.go
mvdan: @vishh what do you mean? Storing `float64` instead of `hourEntry`?

vishh: Storing a pointer to hourEntry as a float64 :)

mvdan: I don't see why that would help. I don't think Go would allow it either, unless you use the `unsafe` package. You can't cast stuff around in Go like you would with `void*` in C.

vishh: Ah! Good point @mvdan.

store/daystore/day_store.go
afein: I guess the comment is misleading. What I meant to say is the following (will be edited in a new commit):

if the DayStore has data within the past 0-1 hour:
- The Average, Max and 95th pecentile are the corresponding values of the underlying StatStore

if the DayStore has more than 1 hour of data then hourly stats have been flushed to the window buffer. Then:
- The Max is calculated as the max of all entries in the window, and the max of the underlying StatStore
- The Average and 95th percentiles are calculated from the entries in the window. the values in the underlying StatStore are NOT considered as some datapoints will be accounted for twice. For example, if a flush happened 10 minutes ago, the Hour StatStore will hold 50 minutes of data that have been accounted for in the latest hourEntry of the window. 

If this is a concern on the accuracy of daily stats, I could Get all the TimePoints from the Hour StatStore that are later than the last hourly flush, and weigh these points appropriately for the calculation of the average. For the 95th percentile, we could calculate the 95th percentile of the remainder of the past one hour and use that to calculate the daily 95th percentile. WDYT?

store/daystore/day_store.go
afein: please view my answer on the previous line comment

store/daystore/day_store.go
afein: I edited the above comment, please view on Github

vishh: Getting points from the hour store will be an expensive operation since the store does not return stats for custom time durations. I am concerned about spikes that we will miss if we don't do it though. I don't see why we will calculate percentile for more than a day. My understanding if that the day store will hold 23 hourly stats and the 24th sample will be derived from the hourly stats store.

afein: We can still capture spikes through the Max, which considers the current hour, and their effect on the overall daily percentile/average will be reflected when that specific hour is flushed, as of now.

If we hold 23 hourly stats and use the StatStore for the past hour then we can only guarantee that the derived stats correspond to the range of (-24,-23) hours ago. The reason is that if the last flush has happened 1 minute ago, then the StatStore holds one hour of information, 59 minutes of which we do not need. Therefore, we would calculate the average,max and 95th percentile of the past 23 hours and 1 minute, in that case. This would lead to the daily Max not reflecting a spike that has occured 23 hours and 10 minutes ago.

If we hold 24 points to avoid that and we still utilize values from the StatStore, then these values are over the 24 hour range that is stored. As we cannot account for partial data in the earliest hour, we would essentially be calculating the max, average and 95th for the past 24 hours and 1 minute instead.

Because of these reasons, I believe that the current implementation is a fair compromise of precision over time length. We use exactly 24 hours of data for the average and 95th, even though these values are at most 59 minutes in the past. For the max, we consider up to 24 hours and 59 minutes in the past, which leads to increased flexibility for spike detection. WDYT

vishh: Thanks for the explanation @afein. Existing implementation is fine for now. Let's document the limitation though. In the future, if it affects users, we can improve it.

afein:
addressed the two issues on behavior documentation and test readability

vishh:
LGTM

afein:
squashing and self-merging

PR #478, opened 2015-08-14
Model Leftovers

afein:
just dropping this here for review while I'm cleaning it up.

\cc @vishh @mvdan 

Please take a look before EOD today.
The tests are not expected to work as of this point, this would involve about ~1.5k LOC changes as well.

### DIFF
--- a/api/v1/model_handlers.go
+++ b/api/v1/model_handlers.go
@@ -24,12 +24,12 @@ import (
 	"github.com/golang/glog"
 
 	"k8s.io/heapster/api/v1/types"
-	"k8s.io/heapster/model"
-	"k8s.io/heapster/store"
+	model_api "k8s.io/heapster/model"
+	"k8s.io/heapster/store/statstore"
 )
 
-// errModelNotActivated is the error that is returned when manager.cluster
-// has not beed initialized.
+// errModelNotActivated is the error that is returned by the API handlers
+// when manager.model has not been initialized.
 var errModelNotActivated = errors.New("the model is not activated")
 
 // RegisterModel registers the Model API endpoints.
@@ -43,7 +43,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Consumes("*/*").
 		Produces(restful.MIME_JSON)
 
-	// The / endpoint returns a list of all the entities that are available in the cluster
+	// The / endpoint returns a list of all the entities that are available in the model
 	ws.Route(ws.GET("/").
 		To(a.allEntities).
 		Filter(compressionFilter).
@@ -56,6 +56,13 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Filter(compressionFilter).
 		Doc("Get a list of all available metrics for the Cluster entity").
 		Operation("availableMetrics"))
+
+	// The /stats/ endpoint returns a list of all available stats for the Cluster entity of the model.
+	ws.Route(ws.GET("/stats/").
+		To(a.clusterStats).
+		Filter(compressionFilter).
+		Doc("Get all available stats for the Cluster entity").
+		Operation("clusterStats"))
 
 	// The /metrics/{metric-name} endpoint exposes an aggregated metric for the Cluster entity of the model.
 	ws.Route(ws.GET("/metrics/{metric-name}").
@@ -84,6 +91,14 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Operation("nodePaths").
 		Param(ws.PathParameter("node-name", "The name of the node to lookup").DataType("string")))
 
+	// The /nodes/{node-name}/stats endpoint returns all available derived stats for a Node entity.
+	ws.Route(ws.GET("/nodes/{node-name}/stats/").
+		To(a.nodeStats).
+		Filter(compressionFilter).
+		Doc("Get all available stats for a Node entity.").
+		Operation("nodeStats").
+		Param(ws.PathParameter("node-name", "The name of the node to lookup").DataType("string")))
+
 	// The /nodes/{node-name}/metrics endpoint returns a list of all available metrics for a Node entity.
 	ws.Route(ws.GET("/nodes/{node-name}/metrics/").
 		To(a.availableMetrics).
@@ -105,7 +120,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(types.MetricResult{}))
 
-	// The /namespaces/ endpoint returns a list of all Namespace entities in the cluster.
+	// The /namespaces/ endpoint returns a list of all Namespace entities in the model.
 	ws.Route(ws.GET("/namespaces/").
 		To(a.allNamespaces).
 		Filter(compressionFilter).
@@ -118,6 +133,14 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Filter(compressionFilter).
 		Doc("Get a list of all available API paths for a namespace entity").
 		Operation("namespacePaths").
+		Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")))
+
+	// The /namespaces/{namespace-name}/stats endpoint returns all available derived stats for a Namespace entity.
+	ws.Route(ws.GET("/namespaces/{namespace-name}/stats/").
+		To(a.namespaceStats).
+		Filter(compressionFilter).
+		Doc("Get all available stats for a Namespace entity.").
+		Operation("namespaceStats").
 		Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")))
 
 	// The /namespaces/{namespace-name}/metrics endpoint returns a list of all available metrics for a Namespace entity.
@@ -141,7 +164,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(types.MetricResult{}))
 
-	// The /namespaces/{namespace-name}/pods endpoint returns a list of all Pod entities in the cluster,
+	// The /namespaces/{namespace-name}/pods endpoint returns a list of all Pod entities in the model,
 	// under a specified namespace.
 	ws.Route(ws.GET("/namespaces/{namespace-name}/pods").
 		To(a.allPods).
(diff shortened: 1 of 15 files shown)


api/v1/model_handlers.go
```diff
@@ -371,7 +449,17 @@ func (a *Api) allFreeContainers(request *restful.Request, response *restful.Resp
 		return
 	}
 	node := request.PathParameter("node-name")
-	response.WriteEntity(cluster.GetFreeContainers(node))
+	response.WriteEntity(makeExternalEntityList(cluster.GetFreeContainers(node)))
+}
+
+// nodePods returns a list of all the available API paths that are available for a node.
+func (a *Api) nodePods(request *restful.Request, response *restful.Response) {
+	cluster := a.manager.GetCluster()
```
vishh: Does this Getter copy the entire model?

deploy/kube-config/kubedash/heapster-controller.json
```diff
@@ -0,0 +1,40 @@
+{
```
vishh: Why is this required if the model is turned on by default?

model/aggregation.go
```diff
@@ -182,33 +183,44 @@ func (rc *realCluster) aggregateMetrics(target *InfoType, sources []*InfoType) e
...
-	// Put all the new values in the TimeStores under target
+	// Put all the new values in the DayStores under target
 	for key, tpSlice := range newMetrics {
 		_, ok := target.Metrics[key]
 		if !ok {
 			// Metric does not exist on target InfoType, create TimeStore
-			newTS := rc.tsConstructor()
-			target.Metrics[key] = &newTS
+			// TODO(afein): configure epsilon
+			newDS := daystore.NewDayStore(100, rc.resolution)
```
vishh: Do you have a follow up PR for epsilon? `100` defeats the whole effort.

afein: please do not review this PR yet, there are 650 failing assertions and a
set of refactoring changes that need to be put in first

model/types.go
```diff
@@ -19,43 +19,52 @@ import (
...
 )
 
 type Cluster interface {
 	// The Update operation populates the Cluster from a cache.
 	Update(cache.Cache) error
 
+	// The simple Get operations extract structural information from the Cluster.
+	GetAvailableMetrics() []string
+	GetNodes() []EntityListEntry
+	GetNamespaces() []EntityListEntry
```
vishh: Ideally, only nodes and namespaces must be accessible from cluster. Those objects should then expose the pods that are under them. I am ok with this flat interface for now, but add a TODO.

vishh:
Why add the new cadvisor deps?

model/aggregation.go
vishh: ok. Ping me once you have updated.

model/types.go
afein: the interface can be renamed to Model to reflect its holistic scope of the cluster's structure. Having sub-objects would require a redesign of the current locking mechanisms, as we have discussed in the past

api/v1/model_handlers.go
afein: this should have been a pointer on the manager. Thanks for catching this!

afein:
An update on this PR:

At the moment, I'm trying to bring this to a condition where tests are passing.
After that, most of the points that @vishh has mentioned will be also addressed, along with other leftover issues.

The majority of this PR includes Model10, which was submitted independently 4-5 weeks ago.
Now that the underlying stores have changed, a large set of past tests are failing, mostly due to assumptions taken on the behavior of Get and Put with the old stores.

I invited an initial review on this PR mostly for structural comments on the implementation of derived stats. However, there are still a few things that are leftover:
- make tests passing
- compare accuracy/memory footprint with that of the the older stores
- cleanup of CAdvisor dependencies
- breakdown of this PR into readable commits.
- definition of epsilon per metric (initially hardcoded in types.go, with a TODO for a configuration object argument on the model constructor)

I will ping @vishh and @mvdan when these tasks are completed and I will remove the (WIP) prefix from this PR then. At that point, please review the PR commit-by-commit for readability and reduced mental contexts.
ETA: Tuesday 08/25

jszczepkowski:
CC @jszczepkowski

afein:
Quick Update:
This PR is almost fully operational. I've had to change the behavior of a few components to comply with the new stores.
As soon as this PR is mergeable, I will start submitting separate smaller PRs that link to this one to facilitate the code review process.

The changes that will be included in this set of PRs are:
- usage of the new StatStore and DayStore
- static epsilon configuration
- Uptime calculation bugfix
- Unit test refactoring for compliance
- renaming of the top-level model interface to "Model" instead of "Cluster" to avoid confusion with the ClusterInfo struct

Furthermore, based on preliminary testing, the memory usage of the Heapster Pod is now consistently between 30-40MB for a not-too-noisy cluster. Using the previous stores, the usage was 200MB instead, leading to an improvement of 5x (rough estimate).

vishh:
Thanks for the update Alex! Looking forward to the PRs!

afein:
retest this please

afein:
@mvdan these are the travis and jenkins errors I was getting after my initial godep cleanup, which is why I closed that original PR.

I am able to build and test locally, so I'm guessing there is some mismatch between what the expected testing environment is and what it should be

mvdan:
Makes sense, it's a dependency of fleet, which isn't removed by this cleanup. Then the question should be, why is it the local build working fine?

afein:
fleet is being used in sources/nodes/ and clusters/, therefore it is reasonable that the build should fail.
The reason that the local build was passing is that the googleapi dependency was located in my gopath.

The main concern around this issue is the following:
- During the `godep save ./...` operation, the googleapi dependency is not included in Godeps, even though it is dependent by fleet

I will run a secondary check on a clean gopath to see if reinstating the googleapi dependency will resolve this issue alone

mvdan:
Sounds like a bug in godep. `godep build` et al falling back to `GOPATH` is also very confusing and sounds like a misfeature to me.

mvdan:
Hang on, are you running the latest godep? I already did the cleanup here: https://github.com/kubernetes/heapster/commit/8523d38ab030e9cc65864d7b0638f12d6a36639d

It cleaned up some libraries, but not that one. So your godep is doing extra cleanups compared to mine, which are wrong.

afein:
You're right, seems to be a bug on my Godep version. I checked out all the Godeps from master and it seems to be working

deploy/kube-config/kubedash/heapster-controller.json
```diff
@@ -0,0 +1,40 @@
...
+	"template": {
+	    "metadata": {
+		"labels": {
+		    "name": "heapster"
+		}
+	    },
+	    "spec": {
+		"containers": [
+		    {
+			"image": "afein/heapster-model:latest",
```
piosz: How about making this image more official?

afein: That's a fair point. I no longer have access to a gcr.io bucket so someone else would have to actually deploy this.

These deployment files can be eventually deprecated, as soon as the next milestone is deployed in all kubernetes clusters. In that case, the model should be accessible by default in the heapster pod running in the kube-system namespace.

api/v1/model_handlers.go
```diff
@@ -332,88 +395,124 @@ func (a *Api) podPaths(request *restful.Request, response *restful.Response) {
...
 		"metrics/",
+		"stats/",
 	}
 	response.WriteEntity(entities)
 }
 
-// allNodes returns a list of all the available node names in the cluster.
+// makeExternalEntityList converts the result of model's getX methods to the external type.
+func makeExternalEntityList(list []model_api.EntityListEntry) []types.ExternalEntityListEntry {
+	res := make([]types.ExternalEntityListEntry, 0)
```
mvdan: `0, len(list)`

api/v1/types/model_types.go
```diff
@@ -31,3 +31,27 @@ type MetricResult struct {
 type MetricResultList struct {
 	Items []MetricResult `json:"items"`
 }
+
+type Stats struct {
+	Average    uint64 `json:"average"`
+	Percentile uint64 `json:"percentile"`
```
mvdan: What percentile?

api/v1/types/model_types.go
```diff
@@ -31,3 +31,27 @@ type MetricResult struct {
...
+
+type ExternalStatBundle struct {
+	Minute Stats `json:"minute"`
+	Hour   Stats `json:"hour"`
+	Day    Stats `json:"day"`
+}
+
+type StatsResponse struct {
+	// Uptime is in seconds
+	Uptime uint64                        `json:"uptime"`
```
mvdan: Why not use time.Time or util.Time?

afein: time.Time represents a time format, however the Uptime is a time.Duration field.
Since there is no standardized format for durations (unlike RFC3339 for time.Time), I opted to output the total number of seconds for that duration

api/v1/types/model_types.go
afein: 95th. I'll change that name, as well as the Stats struct in model/types.go
However, I will not change the json representation as of now to avoid conflicts in Kubedash

api/v1/model_handlers.go
afein: done

api/v1/types/model_types.go
mvdan: I was thinking of Duration, yes :) Sounds fair.

model/getters.go
```diff
@@ -0,0 +1,514 @@
...
+	res := (*ts).Hour.Get(req.Start, req.End)
+	return res, rc.timestamp, nil
+}
+
+// makeEntityList creates an EntityListEntry from a map of metrics.
+func makeEntityListEntry(name string, entities map[string]*daystore.DayStore) EntityListEntry {
+	newListEntry := EntityListEntry{}
+	cpu, ok := entities[cpuUsage]
+	if !ok {
+		newListEntry.CPUUsage = uint64(0)
```
mvdan: All the `uint64(0)` assignments are unnecessary since that's the zero value.

api/v1/model_handlers.go
```diff
@@ -57,6 +57,13 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Doc("Get a list of all available metrics for the Cluster entity").
 		Operation("availableMetrics"))
 
+	// The /stats/ endpoint returns a list of all available stats for the Cluster entity of the model.
+	ws.Route(ws.GET("/stats/").
+		To(a.clusterStats).
+		Filter(compressionFilter).
+		Doc("Get all available stats for the Cluster entity").
+		Operation("clusterStats"))
```
mvdan: Aren't all of these new endpoints missing `Writes()`?

Also, many other endpoints say `Writes(types.MetricResult{})` but that's just wrong, e.g. `GET("/nodes/")`

afein:
Closing this PR. The feature set that was included has been broken down into #508 and #525

Issue #490, opened 2015-08-21
Extract core api types so they can be used by Kubernetes. 


mwielgus:
K8s needs model_types.go (and maybe types.go) from api/v1.

mwielgus:
Fixed with #494

PR #494, opened 2015-08-24
Movie API types to a separate package

mwielgus:
So that they can be imported in k8s without handlers etc.

### DIFF
--- a/integration/heapster_api_test.go
+++ b/integration/heapster_api_test.go
@@ -29,7 +29,7 @@ import (
 	apiErrors "github.com/GoogleCloudPlatform/kubernetes/pkg/api/errors"
 	"github.com/golang/glog"
 	"github.com/stretchr/testify/require"
-	heapster_api "k8s.io/heapster/api/v1"
+	api_v1 "k8s.io/heapster/api/v1/types"
 	sink_api "k8s.io/heapster/sinks/api"
 	"k8s.io/heapster/sinks/cache"
 )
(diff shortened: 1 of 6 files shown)


vishh:
@cadvisorJenkinsBot: add to whitelist

sinks/external_test.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	kube_api "github.com/GoogleCloudPlatform/kubernetes/pkg/api"
 	fuzz "github.com/google/gofuzz"
+	_ "k8s.io/heapster/api/v1/types"
```
vishh: Why is this necessary?

integration/heapster_api_test.go
```diff
@@ -29,7 +29,7 @@ import (
 	apiErrors "github.com/GoogleCloudPlatform/kubernetes/pkg/api/errors"
 	"github.com/golang/glog"
 	"github.com/stretchr/testify/require"
-	heapster_api "k8s.io/heapster/api/v1"
+	heapster_api_types "k8s.io/heapster/api/v1/types"
```
vishh: nit: rename this to `api_v1`.

sinks/external_test.go
mwielgus: Not sure but without this the test fails. Probably a fuzzer checks what packages are initialized or so.

integration/heapster_api_test.go
mwielgus: Done.

sinks/external_test.go
vishh: Hmm. Thats wierd!

vishh:
Thanks for the PR @mwielgus! LGTM.

mwielgus:
#490

sinks/external_test.go
afein: I have observed this test being fuzzy even at the current master. Not sure what the cause is

PR #497, opened 2015-08-25
Launching the influxdb service in kube-system namespace

sanjana-bhat:
monitoring-influxdb service was being launched in the default namespace and heapster was not able to connect to it. This commit moves the service to kube-system where the rest of the services and rc are being launched.


sanjana-bhat:
I signed it!

sanjana-bhat:
I will raise a PR again with commit including my email id associated with the CLA

PR #498, opened 2015-08-25
Launching the influxdb service in kube-system namespace

sanjana-bhat:
This commit includes launching the influxdb service in kube system namespace so that heapster can push to influxdb. There was an error in the influxdb service spec file where default namespace was used.


vishh:
ok to test

vishh:

mvdan:
LGTM

PR #499, opened 2015-08-26
godep cleanup

afein:
removing unused godeps

\cc @vishh @mvdan


afein:
Travis complained on the googleapi dependency at #478 so I'm guessing this cleanup is not actually required

PR #503, opened 2015-08-28
Fix Go 1.5 build

mvdan:
@afein already tried to do this godep cleanup that I did unintentionally in #499. I don't fully understand why he closed that PR, so leaving it in for now.

I can manually revert that cleanup that godep did when I rearranged go-extpoints, if he can confirm that godep is doing something wrong.


afein:
Jenkins was complaining after I removed the googleapi Godeps. I closed the PR to avoid confusion as I wasn't sure at the time if they are somehow used in integration tests

jimmidyson:
Might be worth asking the godep command to a Makefile target? `make godeps` sounds good to me.

mvdan:
@jimmidyson updating the godeps is not as simple as a single command, it usually involves `godep restore`, updating the library you want, and then `godep save`. Or a separate `GOPATH` if you don't want to modify your main one. So I wouldn't have it as a make target.

@afein jenkins isn't complaining, and I can't see why it should be :)

vishh:
LGTM. Thanks for the update @mvdan!

PR #504, opened 2015-08-28
Makefile tweaks

mvdan:


Makefile
```diff
@@ -8,8 +8,8 @@ deps:
 	go get github.com/tools/godep
 
 build: clean deps
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a ./...
```
vishh: This is useful when compiling on OSX. If we switch to docker based builds, we can get rid of this safely.

mvdan: Oh. Fair enough.

mvdan:
@vishh this makes two flaky tests fail way more often, which I think is a good thing. They are `TestSetSinksStore` and `TestRealCacheData`.

mvdan:
Just fixed `TestRealCacheData`. Left it running in a forever loop for nearly an hour, no test failures.

mvdan:
@vishh the GC bug is a serious regression in the latest stable, perhaps a release is due?

vishh:
@mvdan: Thanks for fixing the race. Which GC bug are you referring to?

mvdan:
There was no race, it was a bug in the garbage collection that I just fixed. That is what I meant with "GC bug". It's a serious regression because it removes all nodes at every `runGC()`.

vishh:
Have you posted a PR?

vishh:
@mvdan: Can you include just the makefile changes in this PR?

mvdan:
Sure.

BTW, the `TestSetSinksStore` flake was introduced in 6d4d6e2251fb42e355. It was definitely not flaky when I made my "changing sinks at runtime" changes.

mvdan:
Mentioning the flake because it now happens almost always thanks to `-race`. In the long run we want `-race` enabled and the flake fixed. In the short run, if we can't fix the flake, we can hold off `-race` for now.

vishh:
We can fix the flake. I can take a look at the failing test.

afein:
regarding the failing test: https://github.com/kubernetes/heapster/pull/478#issuecomment-136442475

mvdan:
All green now thanks to #510, merging

PR #506, opened 2015-08-29
Changed behavior of StatStore Get and Last

afein:
Part 1 of #478
 
This PR changes the behavior of the StatStore Get operation to also include one TimePoint that contains the Max of values stored in the current minute.
This behavior is needed to avoid metrics lagging behind after aggregation for abstract entities.

For example, if the latest timestamp is N minutes for all containers, then with the old behavior of Get the latest timestamp for pods would be N-1 and the latest timestamp for namespaces would be at N-2.

The behavior of Last has also changed to return values from LastPut.
- Last(true) now returns the Max value of the items represented in lastPut.
- Last(false) returns the average value of the items represented in lastPut

These two changes guarantee that the derived stats exposed by the heapster model are always sane, meaning that the last minute's Max cannot be larger that the hourly or daily max.

Bonuses: 
- changed the StatStore window size field to uint16, to support stores with more than 256 stored resolutions.
- allowed the DayStore to support Put operations at multiple hours in the future. Similarly to the StatStore, the missing resolutions are assumed to remain constant in value.

\cc @vishh @mvdan

### DIFF
--- a/store/daystore/day_store.go
+++ b/store/daystore/day_store.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package store
+package daystore
 
 import (
 	"fmt"
@@ -30,6 +30,7 @@ import (
 // DayStore holds 24 hours of derived stats, plus an hour-long StatStore that holds -
 // historical data on the current hour.
 // DayStore can calculate the Average, Max and 95th Percentile over the past 24 hours.
+// The DayStore needs to be populated in a chronological order.
 
 // Note on how derived stats are extracted:
 // If the DayStore holds less than 1 hour of data, then the average, max and 95th are -
@@ -79,7 +80,8 @@ func NewDayStore(epsilon uint64, resolution time.Duration) *DayStore {
 	// Calculate how many resolutions correspond to an hour
 	hourNS := time.Hour.Nanoseconds()
 	resNS := resolution.Nanoseconds()
-	intervals := uint8(hourNS / resNS)
+	intervals := uint(hourNS / resNS)
+
 	if hourNS%resNS != 0 {
 		intervals++
 	}
@@ -92,6 +94,7 @@ func NewDayStore(epsilon uint64, resolution time.Duration) *DayStore {
 
 // Put stores a TimePoint into the Hour StatStore, while checking whether it -
 // is time to flush the last hour's stats in the window.
+// Put operations need to be performed in a chronological (time-ascending) order
 func (ds *DayStore) Put(tp statstore.TimePoint) error {
 	ds.Lock()
 	defer ds.Unlock()
@@ -101,9 +104,11 @@ func (ds *DayStore) Put(tp statstore.TimePoint) error {
 		return err
 	}
 
-	ds.validMax = false
+	if tp.Value > ds.cachedMax {
+		ds.validMax = false
+	}
 
-	// Check if this is the first TimePoint ever, in which case flush in one hour.
+	// Check if this is the first TimePoint ever, in which case flush in one hour from now.
 	if ds.lastFlush.Equal(time.Time{}) {
 		ds.lastFlush = tp.Timestamp
 		return nil
@@ -114,7 +119,7 @@ func (ds *DayStore) Put(tp statstore.TimePoint) error {
 		return nil
 	}
 
-	// flush the current hour to the window.
+	// create an hourEntry for the existing hour
 	ds.validAvgPct = false
 	avg, _ := ds.Hour.Average()
 	max, _ := ds.Hour.Max()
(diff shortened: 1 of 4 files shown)


afein:
retest this please

afein:
retest this please

store/daystore/day_store.go
```diff
@@ -79,7 +79,8 @@ func NewDayStore(epsilon uint64, resolution time.Duration) *DayStore {
 	// Calculate how many resolutions correspond to an hour
 	hourNS := time.Hour.Nanoseconds()
 	resNS := resolution.Nanoseconds()
-	intervals := uint8(hourNS / resNS)
+	intervals := uint16(hourNS / resNS)
```
vishh: by cast it to uint16 instead of uint32?

store/daystore/day_store.go
```diff
@@ -124,11 +127,20 @@ func (ds *DayStore) Put(tp statstore.TimePoint) error {
...
 		ninetyFifth: nf,
 	}
-	if ds.size < 24 {
-		ds.size += 1
+
+	// check if the TimePoint is multiple hours in the future
+	// insert the hourEntry the appropriate amount of hours
+	distance := tp.Timestamp.Sub(ds.lastFlush)
+	nextflush := tp.Timestamp
+	for distance.Nanoseconds() >= time.Hour.Nanoseconds() {
```
vishh: I assume this logic still requires input in a chronological order. Make sure you document that.

store/daystore/day_store.go
afein: I opted to use the next largest possible type than uint8. I am not very familiar with Go's struct padding so I am not sure if that will result in any space optimization .

If you believe that uint32 is more appropriate, I'd gladly change it

store/statstore/stat_store.go
```diff
@@ -327,10 +337,26 @@ func (ss *StatStore) Get(start, end time.Time) []TimePoint {
...
 
-	if ss.IsEmpty() || (start.After(end) && end.After(time.Time{})) {
+	if start.After(end) && end.After(time.Time{}) {
+		return result
+	}
+
+	// Generate a TimePoint for the lastPut, if within range
+	low := start.Equal(time.Time{}) || start.Before(ss.lastPut.stamp)
+	hi := end.Equal(time.Time{}) || !end.Before(ss.lastPut.stamp)
+	if ss.lastPut.stamp.After(time.Time{}) && low && hi {
```
vishh: Is the initial check required?

store/statstore/stat_store.go
```diff
@@ -374,32 +400,31 @@ func (ss *StatStore) Get(start, end time.Time) []TimePoint {
...
 
-// Last returns the latest TimePoint represented in the StatStore,
-// or an error if the StatStore is empty.
-// Note: the lastPut field is not used, as we are not confident of its final value.
-func (ss *StatStore) Last() (TimePoint, error) {
+// Last returns the latest TimePoint as held by lastPut,
+// Last returns an error if no Put operations have been performed on the StatStore.
+// The max argument, if true, causes Last to return the max value of the lastPut, -
+// instead of the average
+func (ss *StatStore) Last(max bool) (TimePoint, error) {
```
vishh: Why not return both timepoint and max?

store/statstore/stat_store.go
afein: If the initial check is not performed, then an empty StatStore would return an array with one empty TimePoint. Instead, we would want to skip the lastPut field as well and return an empty array, which would represent that no Put operations have been performed in the StatStore as of yet

store/statstore/stat_store.go
afein: fair point, will do so

store/daystore/day_store.go
afein: will do

afein:
addressed all comments except the change from uint16 -> uint32 .
@vishh please let me know if you believe this change will be beneficial

store/daystore/day_store.go
mvdan: uint8/16 will be padded if the struct is accompanied by a uint32/64. So if your only reason to use uint8/16 here is space optimization, I would just use uint32.

mvdan: As a side note - if your goal is to optimize memory at all costs, what you can do to avoid padding is to use slices or arrays instead of multiple fields of different sizes. But I don't think we need that level of premature optimization - at least not until it is proved necessary.

vishh: uint should be OK here.

store/statstore/stat_store.go
vishh: If we maintain a count, we would not have this issue right?

vishh:
@afein: Replace uint16 to uint if possible. Otherwise just a few nits. Overall LGTM! Thanks!

mvdan:
Agreed with @vishh, uint is better than uint32. I keep on assuming 32-bit... :)

LGTM too

store/statstore/stat_store.go
afein: we are actually maintaining a count within lastPut to be able to calculate the cumulative average. I will change the implementation to use that counter instead.

store/daystore/day_store.go
afein: roger that

afein:
addressed comments and squashed

vishh:
LGTM. Will merge once jenkins is happy!

Issue #501, opened 2015-08-26
/api/v1/model/metrics/ response is empty 


sanjana-bhat:
I am running heapster and I can see the metrics flowing to influxdb. But when I try to access the metrics using the following endpoint http://{master-ip}/api/v1/proxy/namespaces/kube-system/services/heapster/api/v1/model/metrics/ I get empty response. Same goes with /api/v1/model/namespaces/ and /api/v1/model/nodes/{node-name}/metrics

I use the following parameters  "--use_model=true","--model_resolution=1m". Any help here is really appreciated.

vishh:
cc @afein

afein:
Do you get any response when accessing `http://{master-ip}/api/v1/proxy/namespaces/kube-system/services/heapster/api/v1/model/`? 
Also, could you tell us the heapster version you're using through `kubectl logs <heapster-pod>`?

sanjana-bhat:
Yes, I do get response for `http://{master-ip}/api/v1/proxy/namespaces/kube-system/services/heapster/api/v1/model/`
`[
  "metrics/",
  "namespaces/",
  "nodes/"
]`

Heapster version `0.17.0`

I do see these errors in the log
`W0827 03:52:04.018268       1 impl.go:515] Internal Error: reached unreachable CPU Usage parsing flow
W0827 03:52:04.018483       1 impl.go:515] Internal Error: reached unreachable CPU Usage parsing flow
W0827 03:52:04.018612       1 impl.go:515] Internal Error: reached unreachable CPU Usage parsing flow
E0827 03:52:04.021118       1 util.go:69] Recovered from panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
/usr/local/google/home/vishnuk/go/src/github.com/GoogleCloudPlatform/heapster/Godeps/_workspace/src/github.com/GoogleCloudPlatform/kubernetes/pkg/util/util.go:63
/usr/local/google/home/vishnuk/go/src/github.com/GoogleCloudPlatform/heapster/Godeps/_workspace/src/github.com/GoogleCloudPlatform/kubernetes/pkg/util/util.go:54
/usr/lib/google-golang/src/runtime/asm_amd64.s:401
/usr/lib/google-golang/src/runtime/panic.go:387
/usr/lib/google-golang/src/runtime/panic.go:42`

mvdan:
Edited the last comment to fix formatting.

[...]

Issue #484, opened 2015-08-20
Integrate DayStore and StatsStore 


vishh:
The model uses in-memory timestore for storing metrics as of now. This results in increased memory usage.
The newly added Day Store and Stats Store provide a memory efficient way for storing metrics. 
The model should start using these new storage objects.

vishh:
#478

mwielgus:
#508

PR #508, opened 2015-09-01
Change model to use the new stores & Cleanup

alexmavr:
Part 2 of #478 

Changeset:
- Moved Getter methods to getters.go and tests to getters_test.go
- renamed top-level entity to Model from Cluster
- added a /nodes/{node-name}/pods endpoint
- Listing endpoints now also return the latest memory and CPU usage value through the new ExternalEntityListEntry struct
- ported the underlying stores to utilize DayStore structs instead. Epsilon is configured statically for each metric at model/types.go

Please do not be discouraged by the PR size, the majority of changes are on unit tests.

\cc @vishh @mvdan @piosz

### DIFF
--- a/api/v1/model_handlers.go
+++ b/api/v1/model_handlers.go
@@ -24,12 +24,12 @@ import (
 	"github.com/golang/glog"
 
 	"k8s.io/heapster/api/v1/types"
-	"k8s.io/heapster/model"
-	"k8s.io/heapster/store"
+	model_api "k8s.io/heapster/model"
+	"k8s.io/heapster/store/statstore"
 )
 
-// errModelNotActivated is the error that is returned when manager.cluster
-// has not beed initialized.
+// errModelNotActivated is the error that is returned by the API handlers
+// when manager.model has not been initialized.
 var errModelNotActivated = errors.New("the model is not activated")
 
 // RegisterModel registers the Model API endpoints.
@@ -43,7 +43,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Consumes("*/*").
 		Produces(restful.MIME_JSON)
 
-	// The / endpoint returns a list of all the entities that are available in the cluster
+	// The / endpoint returns a list of all the entities that are available in the model
 	ws.Route(ws.GET("/").
 		To(a.allEntities).
 		Filter(compressionFilter).
@@ -105,7 +105,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(types.MetricResult{}))
 
-	// The /namespaces/ endpoint returns a list of all Namespace entities in the cluster.
+	// The /namespaces/ endpoint returns a list of all Namespace entities in the model.
 	ws.Route(ws.GET("/namespaces/").
 		To(a.allNamespaces).
 		Filter(compressionFilter).
@@ -141,7 +141,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(types.MetricResult{}))
 
-	// The /namespaces/{namespace-name}/pods endpoint returns a list of all Pod entities in the cluster,
+	// The /namespaces/{namespace-name}/pods endpoint returns a list of all Pod entities in the model,
 	// under a specified namespace.
 	ws.Route(ws.GET("/namespaces/{namespace-name}/pods").
 		To(a.allPods).
@@ -192,7 +192,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")).
 		Param(ws.PathParameter("pod-name", "The name of the pod to lookup").DataType("string")))
 
-	// The /namespaces/{namespace-name}/pods/{pod-name}/containers/metrics/{container-name} endpoint
+	// The /namespaces/{namespace-name}/pods/{pod-name}/containers/{container-name} endpoint
 	// returns a list of all API paths available for a Pod Container
 	ws.Route(ws.GET("/namespaces/{namespace-name}/pods/{pod-name}/containers/{container-name}").
 		To(a.containerPaths).
@@ -229,6 +229,14 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
 		Writes(types.MetricResult{}))
 
+	// The /nodes/{node-name}/pods/ endpoint returns a list of all Pods entities under a specified node.
+	ws.Route(ws.GET("/nodes/{node-name}/pods/").
+		To(a.nodePods).
+		Filter(compressionFilter).
+		Doc("Get a list of all Pods belonging to a specified Node in the model").
+		Operation("nodePods").
+		Param(ws.PathParameter("node-name", "The name of the namespace to lookup").DataType("string")))
+
 	// The /nodes/{node-name}/freecontainers/ endpoint returns a list of all free Container entities,
 	// under a specified node.
 	ws.Route(ws.GET("/nodes/{node-name}/freecontainers/").
@@ -256,8 +264,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Filter(compressionFilter).
 		Doc("Get a list of all available metrics for a free Container entity").
 		Operation("availableMetrics").
-		Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")).
-		Param(ws.PathParameter("pod-name", "The name of the pod to lookup").DataType("string")).
+		Param(ws.PathParameter("node-name", "The name of the namespace to lookup").DataType("string")).
 		Param(ws.PathParameter("container-name", "The name of the namespace to use").DataType("string")))
 
 	// The /nodes/{node-name}/freecontainers/{container-name}/metrics/{metric-name} endpoint exposes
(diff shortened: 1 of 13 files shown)


mvdan:
Could you split up the PR in two commits, one for the prod code and another for the unit tests? This way code reviewing the prod code first should be a lot easier, especially since the unit test changes are so large.

piosz:
cc @mwielgus @jszczepkowski

afein:
As per @mvdan's suggestion, I added a commit to roll back tests to their respective versions under master. The old tests are, of course, not building or passing.

mvdan:
I actually meant just splitting the changes in two commits, which would mean travis wouldn't fail :) But this works too

afein:
rebased to split implementation and test, as per @mvdan's request.

mvdan:
Added `-race` to go test, so reopening to trigger the tests again.

mwielgus:
All checks are green - can we merge this PR?

afein:
as soon as we get an LGTM from @vishh

afein:
Added a fix for the panic error mentioned in #501

api/v1/types/model_types.go
```diff
@@ -31,3 +31,9 @@ type MetricResult struct {
 type MetricResultList struct {
 	Items []MetricResult `json:"items"`
 }
+
+type ExternalEntityListEntry struct {
```
vishh: This name is very hard to interpret. What is this meant for?

heapster.go
```diff
@@ -40,7 +40,7 @@ var (
 	argSinkFrequency   = flag.Duration("sink_frequency", 10*time.Second, "Frequency at which data will be pushed to sinks")
 	argCacheDuration   = flag.Duration("cache_duration", 5*time.Minute, "The total duration of the historical data that will be cached by heapster.")
 	argUseModel        = flag.Bool("use_model", false, "When true, the internal model representation will be used")
-	argModelResolution = flag.Duration("model_resolution", 2*time.Minute, "The resolution of the timeseries stored in the model. Applies only if use_model is true")
+	argModelResolution = flag.Duration("model_resolution", 1*time.Minute, "The resolution of the timeseries stored in the model. Applies only if use_model is true")
```
vishh: Is it ok to update the model more often? Can we derive model updates based on polling durations instead?

api/v1/model_handlers.go
```diff
@@ -336,84 +344,106 @@ func (a *Api) containerPaths(request *restful.Request, response *restful.Respons
 	response.WriteEntity(entities)
 }
 
-// allNodes returns a list of all the available node names in the cluster.
+// makeExternalEntityList converts the result of model's getX methods to the external type.
+func makeExternalEntityList(list []model_api.EntityListEntry) []types.ExternalEntityListEntry {
```
vishh: What is an external entity?

model/aggregation.go
```diff
@@ -181,34 +185,70 @@ func (rc *realCluster) aggregateMetrics(target *InfoType, sources []*InfoType) e
...
+		err := target.Metrics[key].Put(newTP)
+		if err != nil {
+			return fmt.Errorf("error while performing aggregation: %s", err)
 		}
-		for _, tp := range tpSlice {
-			(*target.Metrics[key]).Put(tp)
+
+	}
+
+	// Set the creation time of the entity to the earliest one that we have found data for.
```
vishh: What is the purpose for tracking earliest timestamp?

api/v1/model_handlers.go
afein: it's actually an external (entity list), rather than an (external entity) list

This function converts a slice of internal entityListEntry to a slice of ExternalEntityListEntry.
The internal and external structs are identical, as of now, but they are distinct to avoid outputting internal types through the API

model/types.go
```diff
@@ -65,34 +66,34 @@ const memWorking = "memory-working"
 const fsLimit = "fs-limit"
 const fsUsage = "fs-usage"
 
-// Request Types.
-// Used as parameters to all the Get methods of the model.
-type ClusterRequest struct {
+// epsilon values for the underlying in-memory stores
+const cpuLimitEpsilon = 10
```
vishh: What are the units?

model/types.go
```diff
@@ -65,34 +66,34 @@ const memWorking = "memory-working"
...
+const cpuUsageEpsilon = 10
+const memLimitEpsilon = 4194304   // 4 MB
+const memUsageEpsilon = 4194304   // 4 MB
+const memWorkingEpsilon = 4194304 // 4 MB
+const fsLimitEpsilon = 10485760   // 10 MB
+const fsUsageEpsilon = 10485760   // 10 MB
+const defaultEpsilon = 100        // used for testing
```
vishh: Why declare the test one here?

api/v1/model_handlers.go
vishh: What is an `entity`?

vishh:
Just a few questions / comments. Overall LGTM! Thanks @afein!

api/v1/model_handlers.go
afein: An entity is something that holds metrics in the model and can be provided as part of a listing. It 
could be a namespace, a pod, a node, or a container.

An entity could also be the cluster itself, but since there is only one cluster entity it cannot be part of a listing

api/v1/types/model_types.go
afein: it is the external type for a single entry of a list of entities.

heapster.go
afein: Model updates should ideally occur every 1 - 3 resolutions. We cannot update more often than the model resolution with the current implementation.

Moreover, the update operation should not be an exact multiple of resolution, to be able to serve more accurate values at the latest minute. For that reason, I set the update frequency to min(cacheDuration, 2 * ModelResolution + 25 sec). This minimizes the number of Update operations that occur at the beginning of the latest resolution, which could decrease accuracy for metrics that are extracted on the very last minute.

I can open an issue with the required steps, if more often updates are required.

model/aggregation.go
afein: Tracking the earliest creation timestamp during aggregation helps identify the uptime for abstract entities.
For example, the creation time of a namespace is the creation time of the first pod that was created in that namespace

model/types.go
afein: The same units as the values for that specific metric.
It's millicores for all CPU metrics and bytes for all other metric. Should I document that?

model/types.go
afein: As of now, filesystem metrics default to defaultEpsilon. I will address this in an upcoming PR

heapster.go
vishh: How did you arrive at the 25 seconds delta? The model resolution of 1 minute is OK. I am more interested in making the updates happen more frequently to ensure that we can serve more recent data to users.

model/aggregation.go
vishh: Got it.

model/types.go
vishh: Yes please :)

model/types.go
vishh: ok.

api/v1/model_handlers.go
vishh: Ok. Can you add a comment about this? Not necessarily in this PR.

api/v1/types/model_types.go
vishh: If this type is exposed over the wire, can we choose an alternative name? `Entity` is hard to interpret. How about `Point` or `MetricPoint` or something thats more descriptive?

vishh:
Just a couple of comments to be addressed @afein

heapster.go
afein: A delta is required due to the existing implementation of the StatStore, where values of the past 1 minute are being averaged together before being placed into buckets. The Get and Last methods of the StatStore expose that last one minute and we are, therefore, interested in its accuracy.

By having the 25 second delta, the values of the last one minute represent a larger set of samples than when not having an offset.

For example, if Updates were happening at every resolution, then the very last resolution of every metric might represent only one actual sample that was used to generate it, depending on the alignment of metrics flowing in.

If no delta is used, then a mechanism would have to be added to force each metric to lag behind in time for the appropriate duration.
For example, a container would be lagging by 1 resolution, its parent pod would be lagging by 2 resolutions and its parent namespace would be lagging behind by 3 resolutions.

api/v1/types/model_types.go
afein: I'm not sure that `Point` is descriptive of what this struct represents, as `Point` sounds like a single point in time of a timeseries.

Essentially this struct could be represent any of the following sets of values:
(Node_name, latest_CPU, latest_Mem) 
(Pod_name, latest_CPU, latest_Mem) 
(Namespace_name, latest_CPU, latest_Mem)
(Container_name, latest_CPU, latest_Mem)

I was using the word `Entity` to describe `Node | Pod | Namespace | Container`, which are essentially the entities of the cluster that are represented in the model. Do you think we should look for a better word to describe these "entities"?

afein:
Addressed comments in the latest commit.

Pending work on this feature set:
- Filesystem Epsilon
- Finalize the period of the Update method

afein:
This PR effectively addressed #484

heapster.go
vishh: I guess the underlying issue here is that of combining metrics storage and aggregation. We can store container and node metrics at a much higher granularity. But we do not want to aggregate at higher granularities. Update() needs to be broken down. Can you file an issue?

api/v1/types/model_types.go
vishh: How about 
```go
type LatestMetric struct {
     Kind string // oneof: Node, Pod, Namespace, Container
     Name string
     CPU uint64 // millicores
     Memory uint64 // bytes
}
```

heapster.go
afein: Totally agree, opened #528

PR #512, opened 2015-09-03
Removed poll_duration flag

piosz:
The flag is redundant with sink_frequency


piosz:
I can add a follow up PR which simplifies interface Source.GetInfo() method by removing resolution field there, since it's now longer needed if this PR is merged.

cc @vishh @mvdan

heapster.go
```diff
@@ -35,8 +35,7 @@ import (
 )
 
 var (
-	argPollDuration    = flag.Duration("poll_duration", 10*time.Second, "The frequency at which heapster will poll for stats")
-	argStatsResolution = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats. Acceptable values are in the range [1 second, 'poll_duration')")
+	argStatsResolution = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats.")
```
vishh: We currently use stats_resolution to down sample stats. Have you looked through the rest of the code to remove the downsampling, since we are choosing to poll more frequently?

piosz: Do you mean this logic in [sources/datasource/utils.go](https://github.com/kubernetes/heapster/blob/master/sources/datasource/utils.go#L24)? That's what I'm proposing in https://github.com/kubernetes/heapster/pull/512#issuecomment-137437419. I think it worth to do it in this particular PR.

If you are ok with those changes I'll do it tomorrow.

vishh: Without resolution, we will end up handling metrics of 1 second granularity. Is this something we want? It will lead to additional memory usage?

piosz: I don't mean removing sampling logic here at all but rather simplifying since we need to take exactly one samples per time interval as for now. It might make sense in the future to support such call in cadvisor (give me the first sample from specified time range) which will significantly reduce the amount of data being sent between heapster and cadvisor. WDYT?

vishh: Got it. I agree with everything you mentioned! I intend to make the new stats APIs in kubelet support resolution. One option we have is to query more often, but ask for only the most recent stat. WDYT of that option?
With this change, my understanding is that heapster will drop all but one stat entry from every query. Have you tested that?

piosz:
cc @mwielgus

manager/manager.go
```diff
@@ -112,12 +112,21 @@ func (rm *realManager) HousekeepModel() {
 }
 
 func (rm *realManager) Housekeep() {
+	for {
+		start := rm.lastSync
+		end := start.Add(rm.resolution)
+		timeToNextSync := end.Sub(time.Now())
+		// TODO: consider adding some delay here
```
mwielgus: What delay?

sources/datasource/utils.go
```diff
@@ -15,31 +15,12 @@
...
-	"github.com/golang/glog"
 	cadvisor "github.com/google/cadvisor/info/v1"
 )
 
-func sampleContainerStats(stats []*cadvisor.ContainerStats, resolution time.Duration) []*cadvisor.ContainerStats {
+func sampleContainerStats(stats []*cadvisor.ContainerStats) []*cadvisor.ContainerStats {
 	if len(stats) == 0 {
 		return stats
 	}
-	filteredStats := []*cadvisor.ContainerStats{}
```
mwielgus: Why this fragment is redundant?

sources/datasource/utils.go
```diff
@@ -15,31 +15,12 @@
...
 
 import (
-	"time"
-
-	"github.com/golang/glog"
 	cadvisor "github.com/google/cadvisor/info/v1"
 )
 
-func sampleContainerStats(stats []*cadvisor.ContainerStats, resolution time.Duration) []*cadvisor.ContainerStats {
+func sampleContainerStats(stats []*cadvisor.ContainerStats) []*cadvisor.ContainerStats {
```
mwielgus: i cannot see any sampling here.

manager/manager.go
piosz: As for now scrapping data from sources takes place right after the end time passed. For example scrapping data for [10:01, 10:02] takes place a few milliseconds after 10:02. If the resolution is high some data may be missed so that we may want add some delay here in the future. I'm ok with removing this comment if you think it is useless.

sources/datasource/utils.go
piosz: After poll_duration flag is removed we want to get exactly one sample for each time interval. There is also a possibility to add support of querying for 1 sample from time interval in cadvisor which will reduce significantly network load (see https://github.com/kubernetes/heapster/pull/512#discussion_r38791456).

mwielgus:
LGTM

vishh:
LGTM. @piosz kindly test both on kubernetes and using a local cadvisor instance to validate this PR. The integration tests don't capture the cadvisor use case. Thanks!

piosz:
I've tested it on Kubernetes already. How can I test the cadvisor case?

vishh:
Run a cadvisor locally and make heapster use an [external source](
https://github.com/kubernetes/heapster/blob/master/docs/source-configuration.md#external
)

piosz:
Tested manually on local setup with cadvisor and Heapster. Seems to work.

PR #518, opened 2015-09-04
Listener for cache (evictions only for now)

mwielgus:
A temporary way of synchronizing evictions in Cache and Model storage. Long term target is to refactor pod/nodeApi (and the framework) so that it provides a convenient stream of updates to pod and node lists.


mwielgus:
cc: @piosz  @jszczepkowski

vishh:
LGTM. The long-term plan also sounds good. We can have a mux on top of the node and pod API objects, Thanks @mwielgus!

PR #522, opened 2015-09-07
Delete functions for model storage

mwielgus:
and some additional fields in Pod/Node/NamespaceInfo

### DIFF
--- a/model/impl.go
+++ b/model/impl.go
@@ -57,6 +57,10 @@ func (rc *realModel) updateTime(new_time time.Time) {
 	rc.timestamp = new_time
 }
 
+func getPodKey(namespace, name string) string {
+	return namespace + "/" + name
+}
+
 // addNode creates or finds a NodeInfo element for the provided (internal) hostname.
 // addNode returns a pointer to the NodeInfo element that was created or found.
 // addNode assumes an appropriate lock is already taken by the caller.
@@ -70,6 +74,7 @@ func (rc *realModel) addNode(hostname string) *NodeInfo {
 		// Node does not exist in map, create a new NodeInfo object
 		node_ptr = &NodeInfo{
 			InfoType:       newInfoType(nil, nil, nil),
+			Name:           hostname,
 			Pods:           make(map[string]*PodInfo),
 			FreeContainers: make(map[string]*ContainerInfo),
 		}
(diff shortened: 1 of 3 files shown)


mwielgus:
cc: @vishh @piosz @jszczepkowski

model/impl.go
```diff
@@ -57,6 +57,10 @@ func (rc *realModel) updateTime(new_time time.Time) {
 	rc.timestamp = new_time
 }
 
+func getPodKey(namespace, name string) string {
+	return namespace + "|" + name
```
piosz: How about using "/" here? I believe I've seen somewhere convention <namespace>/<name>

model/impl.go
```diff
@@ -101,6 +122,22 @@ func (rc *realModel) addNamespace(name string) *NamespaceInfo {
...
 }
 
+// Delets the given namespace with all of the pods.
+func (rc *realModel) deleteNamespace(namespace string) {
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+
+	if namespaceInfo, ok := rc.Namespaces[namespace]; ok {
+		for _, podInfo := range namespaceInfo.Pods {
+
```
piosz: remove this empty line

model/impl.go
```diff
@@ -80,6 +85,21 @@ func (rc *realModel) addNode(hostname string) *NodeInfo {
...
 }
 
+// Deletes the given node with all of the pods and containers on it.
+func (rc *realModel) deleteNode(hostname string) {
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+
+	if node, ok := rc.Nodes[hostname]; ok {
+		for _, podInfo := range node.Pods {
+			if namespaceInfo, ok := rc.Namespaces[podInfo.Namespace]; ok {
```
piosz: How about reusing deletePod here?

piosz:
LGTM, just a few nits

model/impl.go
mwielgus: Done

model/impl.go
mwielgus: Done

model/impl.go
mwielgus: Both deleteNode and pod operate on the same lock. Here the lock is already taken and deletePod would deadlock (at least i think so - mutex doc doesn't say anything about possibility to acquire a lock multiple times - "Lock locks m. If the lock is already in use, the calling goroutine blocks until the mutex is available.")

piosz: That's fine. The problem may be solved by delegating the responsibilty of locking to the caller, but maybe this is won't simplify code here.

piosz:
LGTM

mwielgus:
One of the tests is flaky - see: #523

vishh:
cc @afein

afein:
LGTM

PR #525, opened 2015-09-07
Derived Stats implementation and API endpoints

### LINKED ISSUE #485: Enable Derived metrics APIs in the model
The model can provide mean, average and 95%ile for containers, pods, nodes and cluster. This data needs to be exposed via REST endpoints.

afein:
Part 3/3 of #478 

This PR exposes the derived stats that are accessible by the underlying DayStores/StatStores through a set of API endpoints.

Resolves #485 

\cc @vishh @mvdan @mwielgus @piosz


deploy/kube-config/kubedash/heapster-controller.json
```diff
@@ -0,0 +1,40 @@
+{
```
vishh: Why is this required? Why don't we make the default config work for kubedash?

deploy/kube-config/kubedash/heapster-service.json
```diff
@@ -0,0 +1,22 @@
+{
```
vishh: Ditto as above. Why is this needed?

deploy/kube-config/kubedash/heapster-controller.json
afein: You're right, I included these configs accidentally in this PR.

heapster.go
```diff
@@ -112,6 +112,7 @@ func doWork() ([]source_api.Source, sinks.ExternalSinkManager, manager.Manager,
 	// Spawn the Model Housekeeping goroutine even if the model is not enabled.
 	// This will allow the model to be activated/deactivated in runtime.
 	// Set the housekeeping period to 2 * argModelResolution + 25 sec
+	// TODO(afein): select a more well-defined housekeeping interval
```
vishh: Restating my comment from your previous PR: If we split update to have a separate container and node level metrics storage section and aggregation section, it might be easier to deal with these durations.

deploy/kube-config/kubedash/heapster-controller.json
vishh: is it possible to post incremental changes as separate commits and then squash them at the end? Github makes it hard to track comments if commits are rebased :(

vishh:
LGTM. Thanks @afein for handling all these features! Much appreciated!

heapster.go
afein: Agree, #528

Issue #530, opened 2015-09-09
LastSync value is sometimes broken


mwielgus:
External_test.go became flaky another time. From time to time last sync is set to incorrect value.

```
W0909 14:29:21.049479   12715 external_test.go:188] Event 0 2015-09-09 14:29:20.788056342 +0200 CEST 2015-09-09 14:29:20.788056342 +0200 CEST
W0909 14:29:21.049646   12715 external_test.go:188] Event 1 2015-09-09 14:30:20.788056342 +0200 CEST 2015-09-09 14:30:20.788056342 +0200 CEST
W0909 14:29:21.049667   12715 external_test.go:188] Event 2 2015-09-09 14:31:20.788056342 +0200 CEST 2015-09-09 14:31:20.788056342 +0200 CEST
W0909 14:29:21.049684   12715 external_test.go:188] Event 3 2015-09-09 14:32:20.788056342 +0200 CEST 2015-09-09 14:32:20.788056342 +0200 CEST
W0909 14:29:21.049702   12715 external_test.go:188] Event 4 2015-09-09 14:33:20.788056342 +0200 CEST 2015-09-09 14:33:20.788056342 +0200 CEST
W0909 14:29:21.049720   12715 external_test.go:188] Event 5 2015-09-09 14:34:20.788056342 +0200 CEST 2015-09-09 14:34:20.788056342 +0200 CEST
W0909 14:29:21.049738   12715 external_test.go:188] Event 6 2015-09-09 14:35:20.788056342 +0200 CEST 2015-09-09 14:35:20.788056342 +0200 CEST
W0909 14:29:21.049768   12715 external_test.go:188] Event 7 2015-09-09 14:36:20.788056342 +0200 CEST 2015-09-09 14:36:20.788056342 +0200 CEST
W0909 14:29:21.049790   12715 external_test.go:188] Event 8 2015-09-09 14:37:20.788056342 +0200 CEST 2015-09-09 14:37:20.788056342 +0200 CEST
W0909 14:29:21.049808   12715 external_test.go:188] Event 9 2015-09-09 14:38:20.788056342 +0200 CEST 2015-09-09 14:38:20.788056342 +0200 CEST
--- FAIL: TestSyncLastUpdated (3.35s)
        Location:       external_test.go:194
	Error:		Not equal: time.Time{sec:63577398680, nsec:788056342, loc:(*time.Location)(0xd90ee0)} (expected)
[...]

PR #531, opened 2015-09-09
Fix lastsync test falkiness

mwielgus:
Fix for #530


piosz:
LGTM. Thanks for the fix.

mwielgus:
cc: @vishh

vishh:
LGTM. Thanks for the fix!!

Issue #523, opened 2015-09-07
sinks/external_test.go is flaky


mwielgus:
```
--- FAIL: TestSetSinksStore-2 (1.00s)
	Location:	external_test.go:169
	Error:		Not equal: 1 (expected)
			        != 0 (actual)
		
	Location:	external_test.go:170
	Error:		Not equal: 1 (expected)
			        != 0 (actual)
		
	Location:	external_test.go:174
	Error:		Not equal: 1 (expected)
			        != 0 (actual)
		
	Location:	external_test.go:175
	Error:		Not equal: 1 (expected)
			        != 0 (actual)
		
	Location:	external_test.go:178
	Error:		Not equal: 1 (expected)
			        != 0 (actual)
		
	Location:	external_test.go:179
	Error:		Not equal: 1 (expected)
			        != 0 (actual)
		
	Location:	external_test.go:189
	Error:		Not equal: 2 (expected)
			        != 1 (actual)
		
	Location:	external_test.go:190
	Error:		Not equal: 2 (expected)
			        != 1 (actual)
		
FAIL
coverage: 73.1% of statements
FAIL	k8s.io/heapster/sinks	1.014s
```

mwielgus:
Fixed in #532

PR #532, opened 2015-09-09
Fix flakiness in external_test.go/TestSetSinksStore

mwielgus:
Fix for #523


mwielgus:
cc: @vishh @piosz

piosz:
LGTM

vishh:
Thanks for the fix!!!

mvdan:
Adding sleeps to fix flaky tests seems wrong - shouldn't there be a blocking call somewhere?

mwielgus:
Yep, you are correct. I blindly copied sleep from the below where it is used just after sync. All sleeps (new and the old one) can be removed as there is a sync after starting goroutines.

mwielgus:
#534 removes the sleeps.

mvdan:
Thanks @mwielgus

PR #533, opened 2015-09-10
Fix flakiness in sinks/cache/impl_test.go

mwielgus:
Previously it was possible that 2 pods received the same name ("") resulting in the lower number of pods that were actually stored in the cache.


vishh:
LGTM

PR #534, opened 2015-09-10
Remove sleeps from external_test.go

mwielgus:
They are not needed at all.


mvdan:
Thanks for the cleanup! Will merge once CI passes.

PR #539, opened 2015-09-11
Added exporting cpu/mem request metrics and computing node_reservation metric based on them

piosz:


piosz:
cc @vishh

vishh:
add to whitelist

mvdan:
LGTM

sources/api/types.go
```diff
@@ -57,19 +57,27 @@ func (a *AggregateData) Merge(b *AggregateData) {
...
+type ContainerStats struct {
+	cadvisor.ContainerStats
+}
+
 type Container struct {
 	Hostname   string
 	ExternalID string
 	Name       string
 	Image      string
-	// TODO(vishh): Consider defining an internal Spec and Stats API to guard against
```
mwielgus: This TODO should not be removed.

sources/api/types.go
```diff
@@ -57,19 +57,27 @@ func (a *AggregateData) Merge(b *AggregateData) {
...
 }
 
+type ContainerSpec struct {
+	cadvisor.ContainerSpec
+	CpuRequest    int64
+	MemoryRequest int64
+}
+
+type ContainerStats struct {
+	cadvisor.ContainerStats
```
mwielgus: What is the purpose of this struct?

sources/api/types.go
vishh: +1. We can get rid of cadvisor's Spec and Stats completely. The model does that.

sources/api/types.go
```diff
@@ -57,19 +57,27 @@ func (a *AggregateData) Merge(b *AggregateData) {
 	a.Events = append(a.Events, b.Events...)
 }
 
+type ContainerSpec struct {
+	cadvisor.ContainerSpec
+	CpuRequest    int64
```
mwielgus: Can we have these 2 grouped in a struct?
Would it be good to have CpuLimit as well?

sources/pods.go
```diff
@@ -83,6 +83,12 @@ func (self *realPodsApi) parsePod(podNodePair *podNodePair) *api.Pod {
 		localContainer := api.Container{}
 		localContainer.Name = container.Name
 		localContainer.Image = container.Image
+		if cpu, ok := container.Resources.Requests[kapi.ResourceCPU]; ok {
+			localContainer.Spec.CpuRequest = cpu.MilliValue()
```
vishh: We will eventually need both request and limits I guess. Why not follow kube api semantics and make requests and limits a map[resourceType]Quantity?

sinks/api/supported_metrics.go
```diff
@@ -109,26 +124,41 @@ var statMetrics = []SupportedStatMetric{
...
-		GetValue: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []InternalPoint {
+		GetValue: func(spec *source_api.ContainerSpec, stat *source_api.ContainerStats) []InternalPoint {
 			return []InternalPoint{{Value: int64(spec.Memory.Limit)}}
 		},
 		OnlyExportIfChanged: true,
 	},
 	{
 		MetricDescriptor: MetricDescriptor{
+			Name:        "memory/request",
+			Description: "Memory request",
```
vishh: Should we clarify request vs limit? If we were to run with cadvisor in standalone mode, what will this metric be?

sources/api/types.go
vishh: Clarify what request means since it is currently specific to just kubernetes.

mwielgus: Actually I've found cpu limit inside of cadvisor cpu spec. Adding the new cpu-related fields will create a mess. I would opt for removing cadvisor dependency in sources api (and completing the below TODO).

vishh: That makes sense. We should handle both request and limit though. Values from cAdvisor will translate to limits, but that can be overridden in the case of Kubernetes.

sources/pods.go
mwielgus: Limits are inside of cadvisor structs :/

sources/api/types.go
piosz: This is a good idea and a good candidate for 1.2 cleanup.

sources/api/types.go
piosz: To make it consistent with ContainerSpec above.

sources/api/types.go
piosz: Having these 2 grouped in a struct makes sense but probably the better option is what @vishh described in https://github.com/kubernetes/heapster/pull/539#discussion_r39319547.

As @mwielgus noticed cpu limit is already computed by cadvisor. Let's remove cadvisor dep in a new PR (probably after 1.1). AFAIK there is no requests computed by cadvisor so we need to workaround it by obtaining the data from Kubernetes.

sources/pods.go
piosz: Good idea.

sinks/api/supported_metrics.go
piosz: Extended description. In the standalone mode no values will be exported here.

piosz:
PTAL @mwielgus @vishh 

How about defining an internal Spec&Stats API in a follow up PR?

mwielgus:
Please squash the commits.

piosz:
I think the commits are meaningful so let's leave them.

mwielgus:
LGTM

vishh:
@piosz: SGTM.

sources/pods.go
vishh: Introducing internal abstractions for stats and spec should solve this issue.

vishh:
LGTM. Thanks @piosz

PR #541, opened 2015-09-14
Add Start and Stop methods to Manager

mwielgus:
- remove `Housekeep` and `HouskeepModel` from the interface as they are impl specific.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -31,13 +31,6 @@ import (
 // Manager provides an interface to control the core of heapster.
 // Implementations are not required to be thread safe.
 type Manager interface {
-	// Housekeep collects data from all the configured sources and
-	// stores the data to all the configured sinks.
-	Housekeep()
-
-	// HousekeepModel performs housekeeping for the Model entity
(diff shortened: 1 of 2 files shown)


manager/manager.go
```diff
@@ -49,18 +42,27 @@ type Manager interface {
...
+	model         model.Model
+	sinkManager   sinks.ExternalSinkManager
+	sinkUris      Uris
+	lastSync      time.Time
+	resolution    time.Duration
+	decoder       sink_api.Decoder
+	mainStopChan  chan struct{}
+	sinkStopChan  chan<- struct{}
+	modelStopChan chan struct{}
+	modelDuration time.Duration
```
piosz: how about moving "modelDuration" somewhere above to somehow group all fields?

piosz:
LGTM just a nit

manager/manager.go
mwielgus: done

vishh:
LGTM. Thanks for the cleanup @mwielgus

PR #548, opened 2015-09-14
Support for InfluxDB v0.9+

### LINKED ISSUE #302: Support Influxdb v0.9+
There are bunch of new features _and_ API changes.

vishh:
Fixes #302 

cc @mvdan @piosz @mwielgus @dchen1107 

I needed all the commits to test this feature as a whole. I'm not sure if I can break it up much since it will either break existing code or the integration tests.


vishh:
@mvdan: Can you make @googlebot happy?

mvdan:
@vishh what's the exact syntax to make it happy? Let me have a try...

@googlebot cofirming I am okay with this

mvdan:
Sometimes errors are just logged and ignored, sometimes they are returned and logged, and sometimes they are just returned. Perhaps we should be more consistent about that?

mvdan:
@vishh note that I had made some comments on your pre-rebase commits and they got lost. I can't make comments on the diff directly because it's too large since the godep update and github refuses to let me see the diffs.

Perhaps remove the godep update, take comments, and then re-add the godep update commit?

vishh:
@mvdan: Removed godeps for now. Sorry about losing comments. I don't see them in github history as well. Sigh :(

mvdan:
No worries, I remember them and can re-do them.

sinks/influxdb/driver.go
```diff
@@ -28,56 +28,65 @@ import (
...
-	// Create tags once influxDB v0.9.0 is released.
+var (
+	eventPodID   = sink_api.LabelPodId.Key
+	eventPodName = sink_api.LabelPodName.Key
+	eventHost    = sink_api.LabelHostname.Key
+)
+
+func (sink *influxdbSink) Register(metrics []sink_api.MetricDescriptor) error {
+	if err := sink.createDatabase(); err != nil {
+		glog.Error(err)
```
mvdan: return error?

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,84 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
+	}
+	bp := influxdb.BatchPoints{
+		Points:   points,
+		Database: sink.c.dbName,
+	}
+	if _, err := sink.client.Write(bp); err != nil {
+		if strings.Contains(err.Error(), dbNotFoundError) {
+			sink.dbExists = false
+		}
 		glog.Errorf("failed to write events to influxDB - %s", err)
```
mvdan: return error?

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,84 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
-		points[i][0] = event.LastTimestamp.Time.UTC().Round(time.Millisecond).Unix() // Column 0 - time
-		points[i][1] = hashUID(string(event.UID))                                    // Column 1 - sequence_number
-		if event.InvolvedObject.Kind == "Pod" {
-			points[i][2] = event.InvolvedObject.UID  // Column 2 - pod_id
-			points[i][3] = event.InvolvedObject.Name // Column 3 - pod_name
-		} else {
-			points[i][2] = "" // Column 2 - pod_id
-			points[i][3] = "" // Column 3 - pod_name
-		}
+	points := []influxdb.Point{}
```
mvdan: You can use `make([]influxdb.Point, 0, len(events)` since you already know how much to allocate

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,84 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
+	return point
 }
 
 // Stores events into the backend.
 func (sink *influxdbSink) StoreEvents(events []kube_api.Event) error {
-	dataPoints := []*influxdb.Series{}
+	if err := sink.createDatabase(); err != nil {
+		return err
+	}
+	points := []influxdb.Point{}
```
mvdan: This line can be removed, I think

sinks/influxdb/driver.go
```diff
@@ -188,55 +180,30 @@ func hashUID(s string) uint64 {
...
-		Name:    seriesName,
-		Columns: eventColumns,
-		Points:  points,
-	}, nil
-
-}
-
-func (self *influxdbSink) StoreTimeseries(timeseries []sink_api.Timeseries) error {
-	dataPoints := []*influxdb.Series{}
+	dataPoints := []influxdb.Point{}
```
mvdan: same as above, `make()` with `len(timeseries)`

mvdan:
Comments re-done. Also see https://github.com/kubernetes/heapster/pull/548#issuecomment-140200401

sinks/influxdb/driver.go
vishh: That doesn't work. Since we append and not replace entries, we end up with an array of twice the size. One option is to pass in a reference to individual elements and overwrite them. WDYT? Is it worth it?

sinks/influxdb/driver.go
vishh: Commented the logic for not doing that below.

sinks/influxdb/driver.go
mvdan: You're confusing length and capacity. See https://golang.org/ref/spec#Slice_types - in my comment above, the length is `0` and the capacity is the `len(...)`.

sinks/influxdb/driver.go
vishh: This can fail if the DB is not up yet. I am trying to create the DB here just as an optimization.

sinks/influxdb/driver.go
vishh: Good catch. Done

sinks/influxdb/driver.go
vishh: Done

sinks/influxdb/driver.go
vishh: Yes. Removed

sinks/influxdb/driver.go
vishh: Done

vishh:
@mvdan: All your previous comments have been addressed.

mvdan:
Note that 0.9.4 was just released, so we might as well switch to that.

vishh:
Updated to v0.9.4. @piosz @mwielgus PTAL when you get a chance. If there are no comments, I'd like to get this merged soon.

mvdan:
Could you rebase to fix the conflicts?

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,83 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
 		seriesName = fmt.Sprintf("%s_%s", seriesName, timeseries.MetricDescriptor.Type.String())
 	}
-
-	// Add the real metric value.
-	columns = append(columns, "value")
-	values = append(values, timeseries.Point.Value)
+	point := influxdb.Point{
+		Measurement: seriesName,
+		Tags:        make(map[string]string, 0, len(timeseries.Point.Labels)),
+		Fields: map[string]interface{}{
```
mvdan: This looks like it's missing `gofmt`

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,83 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
-	sink_api.LabelPodName.Key,  // Column 3
-	sink_api.LabelHostname.Key, // Column 4
-	"value",                    // Column 5
+	return point
 }
 
 // Stores events into the backend.
 func (sink *influxdbSink) StoreEvents(events []kube_api.Event) error {
-	dataPoints := []*influxdb.Series{}
+	if err := sink.createDatabase(); err != nil {
```
mvdan: Do we want to create a database at each call?

Also, since we return nil if there are no events, perhaps this could be moved down.

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,83 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
+		}
 		glog.Errorf("failed to write events to influxDB - %s", err)
 		sink.recordWriteFailure()
 	} else {
-		glog.V(1).Info("Successfully flushed events to influxDB")
+		glog.V(4).Info("Successfully flushed events to influxDB")
 	}
-	return err
 
+	return err
```
mvdan: It would be more idiomatic go to do `return err` after `recordWriteFailure()`, drop the `else` and return `nil` here.

sinks/influxdb/driver.go
```diff
@@ -85,101 +94,83 @@ func (self *influxdbSink) metricToSeries(timeseries *sink_api.Timeseries) *influ
...
-		points[i][0] = event.LastTimestamp.Time.UTC().Round(time.Millisecond).Unix() // Column 0 - time
-		points[i][1] = hashUID(string(event.UID))                                    // Column 1 - sequence_number
-		if event.InvolvedObject.Kind == "Pod" {
-			points[i][2] = event.InvolvedObject.UID  // Column 2 - pod_id
-			points[i][3] = event.InvolvedObject.Name // Column 3 - pod_name
-		} else {
-			points[i][2] = "" // Column 2 - pod_id
-			points[i][3] = "" // Column 3 - pod_name
-		}
+	points := make(influxdb.Point, 0, len(events))
```
mvdan: Does this compile? It sounds like you want `make([]influxdb.Point, ...`

sinks/influxdb/driver.go
```diff
@@ -188,54 +179,29 @@ func hashUID(s string) uint64 {
...
-		Name:    seriesName,
-		Columns: eventColumns,
-		Points:  points,
-	}, nil
-
-}
-
-func (self *influxdbSink) StoreTimeseries(timeseries []sink_api.Timeseries) error {
-	dataPoints := []*influxdb.Series{}
+	dataPoints := make(influxdb.Point, 0, len(timeseries))
```
mvdan: Same as above.

sinks/influxdb/driver.go
```diff
@@ -188,54 +179,29 @@ func hashUID(s string) uint64 {
...
-	if err != nil {
-		glog.Errorf("failed to write stats to influxDB - %s", err)
-		self.recordWriteFailure()
+	if _, err = sink.client.Write(bp); err != nil {
+		if strings.Contains(err.Error(), dbNotFoundError) {
+			sink.dbExists = false
+		}
+		glog.Errorf("failed to write stats to influxDB - %v", err)
+		sink.recordWriteFailure()
+	} else {
```
mvdan: Like the other `err` comment, it would be more idiomatic to not have an else and do `return nil` at the end.

sinks/influxdb/driver.go
```diff
@@ -28,56 +28,65 @@ import (
...
+	eventPodName = sink_api.LabelPodName.Key
+	eventHost    = sink_api.LabelHostname.Key
+)
+
+func (sink *influxdbSink) Register(metrics []sink_api.MetricDescriptor) error {
+	if err := sink.createDatabase(); err != nil {
+		glog.Error(err)
+	}
+
 	return nil
```
mvdan: Now that Register does create a database, shouldn't Unregister do something too?

sinks/influxdb/driver.go
mvdan: Also - Register should be called before any StoreEvents, so creating the database here seems redundant.

sinks/influxdb/driver.go
vishh: Hmm. gofmt doesn't seem to change it. The unit tests will fail in case it fails.

sinks/influxdb/driver.go
vishh: If InfluxDB is not up before heapster starts, db creation will fail in Register. To avoid the race condition and to avoid db being accidentally deleted while heapster is running, heapster tries to re-create db is required.

sinks/influxdb/driver.go
vishh: Sure.

sinks/influxdb/driver.go
vishh: I could not compile any of the code because of missing godeps. If this PR looks good in general I can add godeps again.

sinks/influxdb/driver.go
mvdan: I would expect heapster to refuse to use a sink if its registration fails, or to retry a few times.

sinks/influxdb/driver.go
vishh: Good catch.

sinks/influxdb/driver.go
vishh: Registration is done synchronously as part of startup. Even if we
successfully create db as part of startup, what if influxDB crashed and
came back up? The previously created DB would no longer exist. These
changes make heapster resilient to DB failures.

sinks/influxdb/driver.go
vishh: Done.

sinks/influxdb/driver.go
mvdan: You're right: https://play.golang.org/ (press the format button)

It looks like gofmt treats composite literals differently.

sinks/influxdb/driver.go
vishh: Do what? I don't think users want to remove DBs ever. They will lose the data.

sinks/influxdb/driver.go
mvdan: Fair enough.

sinks/influxdb/driver.go
mvdan: Right. It just caught my eye.

sinks/influxdb/driver.go
mvdan: Good point.

vishh:
rebased and fixed all comments.

vishh:
Waiting for @piosz and @mwielgus to take a look before adding godeps commit.

influxdb/Dockerfile
```diff
@@ -1,24 +1,21 @@
...
 
-# Install InfluxDB
-ENV INFLUXDB_VERSION 0.8.8
+MAINTAINER Vishnu kannan "<vishnuk@google.com>"
 
-RUN apt-get update && apt-get install -y curl && curl -s -o /tmp/influxdb_latest_amd64.deb https://s3.amazonaws.com/influxdb/influxdb_${INFLUXDB_VERSION}_amd64.deb && \
-  dpkg -i /tmp/influxdb_latest_amd64.deb && \
-  rm /tmp/influxdb_latest_amd64.deb
+# Install InfluxDB
+ENV INFLUXDB_VERSION 0.9.4
```
rvrignaud: Quick note: 0.9.4.1 has been released (see https://github.com/influxdb/influxdb/releases/tag/v0.9.4.1)

vishh: Thanks for the tip! Updated!

vishh:
I'm adding the godeps back in, to get this merged. cc @piosz @mwielgus

piosz:
@vishh I took a look briefly and I have no obvious concerns. Please let me know what is a timeline with bumping the version of Heapster.

vishh:
@piosz: Upgrading to v0.9 requires an intermediate upgrade to v0.8.9. The
next release of heapster (v0.19.0) will support InfluxDB v0.9.

vishh:
@mvdan: Godeps have been added back. PTAL.
Thanks for reviewing @piosz

mvdan:
The build seems to still be failing

vishh:
Oops. Fixed the build.

mvdan:
LGTM

mvdan:
@vishh could we make the heapster jenkins run multiple builds at a time? if some build fails, which is what is happening now, it takes a long time retrying until it gives up. That blocks all the other builds for hours at a time.

vishh:
@mvdan: It is possible:
1. If one of the PRs is changing the cluster version to test against, then that test run has to be exclusive.
2. The runs should synchronize via a file lock (or something else) to avoid managing cluster lifetimes.
3. Jenkins has to be reconfigured.

Got some spare cycles ? :)

vishh:
Merging this change.

piosz:
@vishh I wouldn't say it was a review;)

PR #549, opened 2015-09-15
Add nodeMetrics expapi/v2 endpoint

mvdan:
@vishh PTAL

### DIFF
--- a/expapi/v2/metrics_handlers.go
+++ b/expapi/v2/metrics_handlers.go
@@ -1,0 +1,185 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v2
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"net/http"
+	"time"
+
+	restful "github.com/emicklei/go-restful"
+	"github.com/golang/glog"
+
+	"k8s.io/heapster/expapi/v2/types"
+	model_api "k8s.io/heapster/model"
+	"k8s.io/kubernetes/pkg/api/resource"
+	"k8s.io/kubernetes/pkg/api/unversioned"
+)
+
+// errModelNotActivated is the error that is returned when manager.cluster
+// has not beed initialized.
+var errModelNotActivated = errors.New("the model is not activated")
+
+// RegisterMetrics registers the Metrics API endpoints.
+// All endpoints that end with a {metric-name} also receive a start time query parameter.
+// The start and end times should be specified as a string, formatted according to RFC 3339.
+// These apis are experimental, so they may change or disappear in the future.
+func (a *Api) RegisterMetrics(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/experimental/v2").
+		Doc("Root endpoint of the stats model").
+		Consumes("*/*").
+		Produces(restful.MIME_JSON)
+
+	ws.Route(ws.GET("/nodeMetrics/derived/").
+		To(a.derivedNodeMetricsList).
+		Filter(compressionFilter).
+		Doc("Get a list of all available metrics for all nodes").
+		Writes([]types.DerivedNodeMetrics{}).
+		Operation("derivedNodeMetricsList"))
+
+	ws.Route(ws.GET("/nodeMetrics/derived/{node-name}").
+		To(a.derivedNodeMetrics).
+		Filter(compressionFilter).
+		Doc("Get a list of all raw metrics for a Node entity").
+		Writes(types.DerivedNodeMetrics{}).
+		Operation("derivedNodeMetrics").
+		Param(ws.PathParameter("node-name", "The name of the node to look up").DataType("string")))
+
+	container.Add(ws)
+}
+
+func portStats(w *types.MetricsWindows, byName map[string]model_api.StatBundle) error {
+	byDuration := make(map[time.Duration]*types.MetricsWindow)
+	for _, d := range [...]time.Duration{time.Minute, time.Hour, 24 * time.Hour} {
+		w.Windows = append(w.Windows, types.MetricsWindow{
+			Duration: unversioned.Duration{Duration: d},
+			Mean:     make(types.ResourceUsage),
+			Max:      make(types.ResourceUsage),
+			NinetyFifthPercentile: make(types.ResourceUsage),
+		})
+		byDuration[d] = &w.Windows[len(w.Windows)-1]
+	}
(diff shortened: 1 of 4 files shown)


expapi/v2/metrics_handlers.go
```diff
@@ -0,0 +1,190 @@
...
+// The start and end times should be specified as a string, formatted according to RFC 3339.
+func (a *Api) RegisterMetrics(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/experimental/v2").
+		Doc("Root endpoint of the stats model").
+		Consumes("*/*").
+		Produces(restful.MIME_JSON)
+
+	ws.Route(ws.GET("/nodeMetrics/available/").
```
vishh: Why do we need a separate endpoint for available? Why not display the available sub resources for `/nodeMetrics/`?

vishh: May be I misunderstood the purpose. Is this API returning all the available metrics like `cpu`, `memory`, etc? If yes, why do we need it?

mvdan: It was just copied over from the model. As a means to check if the model was running. I'll remove it.

expapi/v2/metrics_handlers.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func getNodeMetrics(model model_api.Model, name string) (*types.DerivedNodeMetrics, error) {
+	res, _, err := model.GetNodeStats(model_api.NodeRequest{
+		NodeName: name,
+	})
+	if err != nil {
+		return nil, err
+	}
+	const minute, hour, day = "1m", "1h", "1d"
```
vishh: I assume a big chunk of this function will be common to other type of metrics as well. Why make it local to node metrics?

mvdan: Sure.

BTW, this is a WIP, so don't review yet as this is still being written.

expapi/v2/types/metrics_types.go
```diff
@@ -0,0 +1,73 @@
...
+
+type MetricResultList struct {
+	Items []MetricResult `json:"items"`
+}
+
+type DerivedNodeMetrics struct {
+	v1.TypeMeta   `json:",inline"`
+	v1.ObjectMeta `json:"metadata,omitempty"`
+
+	NodeMetrics      MetricsWindows
```
vishh: json tag for this?

expapi/v2/types/metrics_types.go
```diff
@@ -0,0 +1,73 @@
...
+	SystemContainers []struct {
+		ContainerName string         `json:"containername"`
```
vishh: nit: Let's stick to camelCase starting with lower case for json tags.

expapi/v2/types/metrics_types.go
```diff
@@ -0,0 +1,73 @@
...
+		Windows       MetricsWindows `json:"windows"`
+	}
+}
+
+type DerivedNodeMetricsList struct {
+	v1.TypeMeta `json:",inline"`
```
vishh: Are the `meta` objects required from heapster perspective? Since it provides read-only APIs I don't see a reason. I could be wrong.

expapi/v2/metrics_handlers.go
```diff
@@ -0,0 +1,190 @@
...
+// has not beed initialized.
+var errModelNotActivated = errors.New("the model is not activated")
+
+// RegisterMetrics registers the Metrics API endpoints.
+// All endpoints that end with a {metric-name} also receive a start time query parameter.
+// The start and end times should be specified as a string, formatted according to RFC 3339.
+func (a *Api) RegisterMetrics(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/experimental/v2").
```
vishh: We should specify that `experimental` means it will be deleted anytime. Also why not use `experimental/v0.1` ?

vishh:
Just some nits. I will make another pass once the PR is ready for review. Overall LGTM!

expapi/v2/metrics_handlers.go
mvdan: It's `experimental/v2` since this may become `api/v2`.

expapi/v2/types/metrics_types.go
mvdan: We can remove them.

expapi/v2/metrics_handlers.go
vishh: ok

mvdan:
@vishh comments amended and WIP removed. Checked that the endpoint works. Note the two TODOs.

model/types.go
```diff
@@ -48,12 +48,12 @@ type Model interface {
 	GetFreeContainerMetric(FreeContainerMetricRequest) ([]statstore.TimePoint, time.Time, error)
 
 	// The GetXStats operations extract all derived stats for a single entity of the cluster.
-	GetClusterStats() (map[string]StatBundle, time.Duration, error)
-	GetNodeStats(NodeRequest) (map[string]StatBundle, time.Duration, error)
-	GetNamespaceStats(NamespaceRequest) (map[string]StatBundle, time.Duration, error)
-	GetPodStats(PodRequest) (map[string]StatBundle, time.Duration, error)
-	GetPodContainerStats(PodContainerRequest) (map[string]StatBundle, time.Duration, error)
-	GetFreeContainerStats(FreeContainerRequest) (map[string]StatBundle, time.Duration, error)
+	GetClusterStats() (map[string]StatBundle, time.Time, time.Duration, error)
```
vishh: nit: We should have explicit names for the return arguments or at-least document their meaning.

vishh: Also the returns types are essentially the same. Is it worth unifying them into a single struct and re-use in all methods?

mvdan: Yes.

mvdan:
@vishh just applied your stats result struct suggestion, makes everything a lot simpler :)

There's only one TODO left, the system containers bit.

mvdan:
Re-adding WIP until the TODO is fixed.

mvdan:
@vishh progress update:

* Rebased on master
* Updated the experimental api types to match the ones in nodeMetrics (replace map with list et al)
* Updated kubernetes in godeps to my nodeMetrics PR
* Add kube-heapster-bridge

vishh:
@mvdan: is this still a WIP?

mvdan:
FWIW, I'll keep WIP since I'm pushing my progress to this PR.

mvdan:
@vishh rebased on master and removed the bridge changes - they'll come in a separate PR. Ready to merge.

expapi/v2/metrics_handlers.go
```diff
@@ -0,0 +1,185 @@
...
+var errModelNotActivated = errors.New("the model is not activated")
+
+// RegisterMetrics registers the Metrics API endpoints.
+// All endpoints that end with a {metric-name} also receive a start time query parameter.
+// The start and end times should be specified as a string, formatted according to RFC 3339.
+// These apis are experimental, so they may change or disappear in the future.
+func (a *Api) RegisterMetrics(container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.
+		Path("/experimental/v2").
```
vishh: What is the output of this endpoint?

mvdan: This isn't an endpoint - it's the path for the whole WebService. Basically, the prefix and other common parameters for the endpoints under it. So `/nodeMetrics` below becomes `/experimental/v2/nodeMetrics`, etcetera.

vishh:
LGTM

PR #550, opened 2015-09-15
Do error on go vet failure

mvdan:


mvdan:
`go vet` prints errors to stderr, so the output caught by the shell script was always empty.

vishh:
Awesome. Can you update `integration/.jenkins.sh` as well?

mvdan:
Why would I? It uses `make unit-test` just like travis. The `Makefile` itself uses the script.

vishh:
Ah! Never mind. Thanks for reminding the existence of Makefiles.

vishh:
LGTM

PR #553, opened 2015-09-15
Integration test for Model

mwielgus:
It is not a full E2E test (like in `integration/*`) but it checks whether the data is correctly passed from `Source` through `Cache` to `Model` and if `Model` returns the expected metrics properly. It doesn't bring up any instance on Kuberentes.

More tests will be added in follow-up PRs.

### DIFF
--- a/integration/model_test.go
+++ b/integration/model_test.go
@@ -1,0 +1,154 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	cadvisor "github.com/google/cadvisor/info/v1"
+	"github.com/stretchr/testify/assert"
+
+	"k8s.io/heapster/manager"
+	model_api "k8s.io/heapster/model"
+	"k8s.io/heapster/sinks"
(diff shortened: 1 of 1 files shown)


integration/model_test.go
```diff
@@ -0,0 +1,156 @@
...
+	loadAverageMilli = 300
+)
+
+func (t testSource) buildPods(start time.Time) []source_api.Pod {
+
+	timeElapsed := start.Sub(t.createTimestamp)
+
+	result := []source_api.Pod{}
+	for i := 0; i < podCount; i++ {
+
```
piosz: nit:removing this empty line

integration/model_test.go
```diff
@@ -0,0 +1,156 @@
...
+	createTimestamp time.Time
+}
+
+const (
+	podCount         = 10
+	testNamespace    = "testnamespace"
+	loadAverageMilli = 300
+)
+
+func (t testSource) buildPods(start time.Time) []source_api.Pod {
```
piosz: Does it make sense to use fuzzer here and then overwrite some fields required for the test?

integration/model_test.go
```diff
@@ -0,0 +1,156 @@
...
+		PodName:       "pod-0",
+		MetricRequest: model_api.MetricRequest{
+			Start:      start,
+			End:        time.Time{},
+			MetricName: "cpu-usage",
+		},
+	})
+	assert.NoError(err)
+	//TODO: Expect more than 1 metric once #551 is fixed
+	assert.True(len(metrics) >= 1)
```
piosz: I believe there is a better function to test this condition. Same below.

integration/model_test.go
mwielgus: Fuzzed structs are already used in unit tests to check corner cases. Here I would like to check if nothing is lost/dropped on the way if the values are correct. The values set below are (mostly) needed and the rest is irrelevant from the point of view of this test.

integration/model_test.go
mwielgus: Done

integration/model_test.go
mwielgus: Not a big difference but OK. Changed to `NotEmpty` and `InEpsilon`.

piosz:
LGTM feel free to merge on green

mwielgus:
All green, merging.

vishh:
LGTM

PR #536, opened 2015-09-11
complete the doc of running Heapster standalone

huangyuqi:
complete the doc of running Heapster standalone

PR #554, opened 2015-09-15
Enable Model by default

mwielgus:
All of the required PRs are in. Let's enable the model by default.


mwielgus:
cc: @vishh @afein @piosz @jszczepkowski

afein:
Perhaps we should leave the flag as false by default, to avoid conflicts with users of non-Kubernetes clusters. The heapster deployment files that are used by Kubernetes can set the flag to true instead.

mwielgus:
It seems there is no big conflict if Kubernetes is not used. #536 shows Heapster in a standalone mode using model API.

vishh:
Yay! Finally! Thanks for all the amazing work @afein @mwielgus @piosz and @mvdan! LGTM!

PR #556, opened 2015-09-15
Refactored GCE auth utils.

### LINKED ISSUE #552: GCL sink doesn't work once the GCE token expires
The token is not refreshed periodically and after some time it causes the following error:
```
E0915 09:00:25.457853       1 external.go:87] failed to sync data to sinks - encountered the following errors: request to "https:/v1beta3/projects/marine-bison-90112/logs/kubernetes.io%2Fevents/entries:write" failed with status "401 Unauthorized" and response: "{\n  \"error\": {\n    \"code\": 401,\n    \"message\": \"Request had invalid authentication credentials.\",\n    \"status\": \"UNAUTHENTICATED\"\n  }\n}\n"
```

piosz:
fixes #552

### DIFF
--- a/util/gce/auth.go
+++ b/util/gce/auth.go
@@ -19,11 +19,23 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
+	"github.com/golang/glog"
 	"google.golang.org/cloud/compute/metadata"
 )
 
-type AuthToken struct {
+const (
+	metadataAuthScopes       = "instance/service-accounts/default/scopes"
+	metadataAuthToken        = "instance/service-accounts/default/token"
+	earlyRefreshTokenSeconds = 60
+	waitForTokenInterval     = 1 * time.Second
+	waitForTokenTimeout      = 30 * time.Second
+)
+
+// GCE specific representation of authorization token. For parsing purposes.
+type authToken struct {
 	// The actual token.
 	AccessToken string `json:"access_token"`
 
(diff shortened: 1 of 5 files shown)


util/gce/auth.go
```diff
@@ -34,9 +46,104 @@ type AuthToken struct {
...
+	WaitForToken() (string, error)
+}
+
+// AuthTokenProvider is thread-safe.
+type realAuthTokenProvider struct {
+	// Token to use for authentication.
+	token string
+
+	// Protects <token>.
+	mutex sync.RWMutex
```
vishh: nit: We can embed the mutex into the parent struct.

util/gce/auth.go
```diff
@@ -50,18 +157,20 @@ func VerifyAuthScope(expectedScope string) error {
...
+		token: token,
 	}
+}
 
-	var token AuthToken
-	err = json.Unmarshal([]byte(rawToken), &token)
-	if err != nil {
-		return AuthToken{}, fmt.Errorf("failed to unmarshal service account token with output %q: %v", rawToken, err)
-	}
+type fakeAuthTokenProvider struct {
```
vishh: How about placing this in a separate file?

vishh:
Thanks for the fix @piosz! LGTM

util/gce/auth.go
piosz: I think it is implementation specific. There is no mutex in fakeAuthTokenProvider.

util/gce/auth.go
piosz: done

util/gce/auth.go
vishh: I meant 
```go
type realAuthTokenProvider struct {
    sync.RWMutex
    ...
}
```

piosz: Got it! Sorry that I haven't understood before. Done.

Issue #487, opened 2015-08-20
Add integration tests for the model APIs


vishh:
The model API semantics are not tested anywhere as of now. The existing heapster API e2e tests can be extended to test model APIs as well.

vishh:
@afein @piosz @mwielgus Is anyone planning on working on this?

piosz:
@jszczepkowski?

afein:
I am currently busy for the remainder of this week. If noone else picks this up I could work on it over the weekend

mwielgus:
I can take a look.

mwielgus:
First test: #553

PR #558, opened 2015-09-15
Model api integration test + more debugs

mwielgus:
Depends on #554.

### DIFF
--- a/integration/heapster_api_test.go
+++ b/integration/heapster_api_test.go
@@ -52,34 +52,50 @@ var (
 )
 
 func deleteAll(fm kubeFramework, ns string, service *kube_api.Service, rc *kube_api.ReplicationController) error {
+	glog.V(2).Infof("Deleting rc %s/%s...", ns, rc.Name)
 	if err := fm.DeleteRC(ns, rc); err != nil {
+		glog.V(2).Infof("Failed to delete rc: %v", err)
 		return err
 	}
+	glog.V(2).Infof("Deleted rc %s/%s.", ns, rc.Name)
 
+	glog.V(2).Infof("Deleting service %s/%s.", ns, service.Name)
 	if err := fm.DeleteService(ns, service); err != nil {
+		glog.V(2).Infof("Failed to delete service: %v", err)
 		return err
 	}
+	glog.V(2).Infof("Deleted service %s/%s.", ns, service.Name)
 	return nil
 }
 
 func createAll(fm kubeFramework, ns string, service **kube_api.Service, rc **kube_api.ReplicationController) error {
+	glog.V(2).Infof("Creating rc %s/%s...", ns, (*rc).Name)
 	if newRc, err := fm.CreateRC(ns, *rc); err != nil {
+		glog.V(2).Infof("Failed to create rc: %v", err)
 		return err
 	} else {
 		*rc = newRc
 	}
+	glog.V(2).Infof("Created rc %s/%s.", ns, (*rc).Name)
 
+	glog.V(2).Infof("Creating service %s/%s...", ns, (*service).Name)
 	if newSvc, err := fm.CreateService(ns, *service); err != nil {
+		glog.V(2).Infof("Failed to create service: %v", err)
 		return err
 	} else {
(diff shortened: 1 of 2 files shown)


mwielgus:
cc: @vishh @piosz

mwielgus:
Fixes #487

mvdan:
Note that this collides with #554

integration/heapster_api_test.go
```diff
@@ -190,6 +207,31 @@ func getSchema(fm kubeFramework, svc *kube_api.Service) (*api_v1.TimeseriesSchem
...
+		Prefix("proxy").
+		Resource("services").
+		Name(svc.Name).
+		Suffix(url).
+		Do().Raw()
+	if err != nil {
+		return nil, err
+	}
+	var metrics api_v1.MetricResultList
+	if err := json.Unmarshal(body, &metrics); err != nil {
```
vishh: Do we want to verify that all the expected resources are present in the response?

integration/heapster_api_test.go
```diff
@@ -190,6 +207,31 @@ func getSchema(fm kubeFramework, svc *kube_api.Service) (*api_v1.TimeseriesSchem
 	return &timeseriesSchema, nil
 }
 
+func getModelMetrics(fm kubeFramework, svc *kube_api.Service, pod *kube_api.Pod) (*api_v1.MetricResultList, error) {
+	url := fmt.Sprintf("/api/v1/model/namespaces/%s/pod-list/%s/metrics/%s",
```
vishh: What about all the other model APIs?

mvdan: Maybe the issue should be closed once all endpoints are tested. Is that what you mean?

vishh: Yeah. That makes sense.

integration/heapster_api_test.go
mwielgus: Added more checks.

integration/heapster_api_test.go
mwielgus: I would rather check the other endpoints in a more mocked and unit-like test for `api/v1/model_handler.go` where I can verify if the returned values are correct.

vishh: Would unit-like tests validate the restful layer as well? If not, at-least
poking those endpoints to make sure they work will be useful.

mwielgus: Well, I hope so. I can still poke all of the endpoints and check if they don't crash (but probably in the next PR).

mwielgus: With the E2E approach as here i can only check if the output is returned at all and if it is sane, not if the endpoints are correctly bound to model funcs.

vishh: Fair point! I see value for all of them and in terms of priority, the
bounding of methods might be the final thing we want to test, with the
assumption that code reviews would have caught that.

vishh: I am perfectly fine with adding checks for the other endpoints in a future
PR. I just wanted to bring that up.

integration/heapster_api_test.go
```diff
@@ -431,13 +530,16 @@ func apiTest(kubeVersion string) error {
...
 			break
 		}
 		if attempts == 0 {
+			glog.V(2).Info("Too many attempts.")
 			return err
 		}
+		glog.V(2).Infof("Some tests failed. Retrying.")
 		attempts--
-		time.Sleep(time.Second)
+		time.Sleep(time.Second * 10)
```
vishh: This change might cause timeouts in the `go test` framework. We will have to update the timeout in the makefiles.

vishh:
@mwielgus: Just one more comment. Otherwise LGTM. This is super useful!

integration/heapster_api_test.go
mwielgus: If this test doesn't complete in <10min then something is terribly broken (or your connection is dead slow). Increasing the timeout beyond 30m will not make any good.

vishh: Fair point. Lets go with this for now.

Issue #559, opened 2015-09-15
Startup problems in integration tests


mwielgus:
Sometimes integration tests fail because of:
```
I0915 23:37:51.645033   31536 heapster_api_test.go:57] Failed to delete rc: unable to modify replica count for rc heapster: replicationControllers "heapster" not found
--- FAIL: TestHeapster (70.27s)
        Location:       heapster_api_test.go:571
	Error:  	No error is expected but got unable to modify replica count for rc heapster: replicationControllers "heapster" not found
```
or 
```
 System container "docker-daemon" not found on host: "kubernetes-minion-9q5q" - map[kube-proxy:map[kubernetes-minion-lntk:{} kubernetes-minion-nfb8:{} kubernetes-minion-46ao:{}] docker-daemon:map[kubernetes-minion-lntk:{} kubernetes-minion-nfb8:{} kubernetes-minion-46ao:{}] kubelet:map[kubernetes-minion-lntk:{} kubernetes-minion-nfb8:{} kubernetes-minion-46ao:{}] machine:map[kubernetes-minion-lntk:{} kubernetes-minion-nfb8:{} kubernetes-minion-46ao:{}] system:map[kubernetes-minion-46ao:{} kubernetes-minion-nfb8:{} kubernetes-minion-lntk:{}]]
```

vishh:
I wonder if running the tests in a separate namespace would resolve the
first failure.
The second failure may be because some nodes are not `Ready` yet?

mwielgus:
#563

mwielgus:
Probably fixed.

PR #562, opened 2015-09-16
More integration tests for Model + some fixes in the framework

mwielgus:
Fix for #487

### DIFF
--- a/integration/heapster_api_test.go
+++ b/integration/heapster_api_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/stretchr/testify/require"
 	api_v1 "k8s.io/heapster/api/v1/types"
+	"k8s.io/heapster/model"
 	sink_api "k8s.io/heapster/sinks/api"
 	"k8s.io/heapster/sinks/cache"
 	kube_api "k8s.io/kubernetes/pkg/api"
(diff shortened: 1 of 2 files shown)


integration/heapster_api_test.go
```diff
@@ -53,16 +54,32 @@ var (
 
 func deleteAll(fm kubeFramework, ns string, service *kube_api.Service, rc *kube_api.ReplicationController) error {
 	glog.V(2).Infof("Deleting rc %s/%s...", ns, rc.Name)
-	if err := fm.DeleteRC(ns, rc); err != nil {
-		glog.V(2).Infof("Failed to delete rc: %v", err)
-		return err
+	retries := 5
+	for {
```
piosz: How about using here time wait pattern here (see link below)?

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/util.go#L394

integration/heapster_api_test.go
```diff
@@ -53,16 +54,32 @@ var (
...
+		retries--
 	}
 	glog.V(2).Infof("Deleted rc %s/%s.", ns, rc.Name)
 
 	glog.V(2).Infof("Deleting service %s/%s.", ns, service.Name)
-	if err := fm.DeleteService(ns, service); err != nil {
-		glog.V(2).Infof("Failed to delete service: %v", err)
-		return err
+	retries = 5
+	for {
```
piosz: Same here.

integration/heapster_api_test.go
```diff
@@ -419,34 +417,195 @@ func runSinksTest(fm kubeFramework, svc *kube_api.Service) error {
...
+}
+
 func runModelTest(fm kubeFramework, svc *kube_api.Service) error {
-	podList, err := fm.GetPodList()
+	podList, err := fm.GetRunningPods()
 	if err != nil {
 		return err
 	}
-	if len(podList.Items) == 0 {
+	if len(podList) == 0 {
```
piosz: I'd create a dummy pod here rather than assume that at least one is running in the cluster.

mwielgus: There is Heapster running ...

integration/heapster_api_test.go
mwielgus: This is a temporary fix for unreliable delete - see #559. Delete should mean delete and there should be no reason for retries. I will refactor this completely in the next PR.

integration/heapster_api_test.go
mwielgus: As above.

integration/heapster_api_test.go
piosz: Ok, let's assume it's enough.

integration/heapster_api_test.go
piosz: Ack

piosz:
LGTM

vishh:
LGTM. Thanks @mwielgus !

PR #563, opened 2015-09-16
Run integration tests in a custom namespace (and cleanup via ns deletion)

mwielgus:


vishh:
LGTM.

mwielgus:
In Jenkins, Heapster service has repeating problems with finding a matching pod (in the namespace) what is quite odd as it works perfectly fine on my computer/cluster.

vishh:
That's weird. Shall we try bumping up the Kubernetes release ? It is currently set to v1.0.1.

mwielgus:
I bumped the version to 1.0.5. Integration tests pass (travis failed because of flaky test - filled an issue for that #565). I will start yet another test run to confirm that we are good now.

integration/heapster_api_test.go
```diff
@@ -630,10 +625,10 @@ func apiTest(kubeVersion string) error {
 	if err := createAll(fm, ns, &svc, &rc); err != nil {
 		return err
 	}
-	if err := fm.WaitUntilPodRunning(ns, rc.Spec.Template.Labels, time.Minute); err != nil {
+	if err := fm.WaitUntilPodRunning(ns, rc.Spec.Template.Labels, 5*time.Minute); err != nil {
```
mvdan: Why are we bumping the timeout here and below?

mwielgus: Just to check if this helps. There were multiple runs (only in Jenkins) where service could not find his endpoints within the previous timeouts.

mwielgus:
Getting rid of system-like stuff from the configs and changing the labels a bit helped. All tests pass. Merging

integration/heapster_api_test.go
mwielgus: Reverted

PR #574, opened 2015-09-17
Upgrading to InfluxDB v0.8.9

vishh:
Cherrypicking into v0.18 release


vishh:
cc @dchen1107

dchen1107:
Thanks!

PR #589, opened 2015-09-21
RunningInKuberenes flag (used in API handler registration)

### LINKED ISSUE #575: Do not export Kube specific model endpoints in non-kube clusters
@afein: Can we avoid exposing endpoints specific to kubernetes clusters like `namespaces` and `pods` from non-kubernetes clusters? AFAIK, the other endpoints like containers, nodes and cluster should still continue to work.

mwielgus:
Fixes #575

### DIFF
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -26,17 +26,19 @@ import (
 )
 
 type Api struct {
-	manager manager.Manager
+	manager             manager.Manager
+	runningInKubernetes bool
 }
 
 // Create a new Api to serve from the specified cache.
-func NewApi(m manager.Manager) *Api {
+func NewApi(m manager.Manager, runningInKuberentes bool) *Api {
 	return &Api{
-		manager: m,
+		manager:             m,
+		runningInKubernetes: runningInKuberentes,
 	}
 }
 
-// Register the Api on the specified endpoint.
+// Register the mainApi on the specified endpoint.
 func (a *Api) Register(container *restful.Container) {
 	ws := new(restful.WebService)
 	ws.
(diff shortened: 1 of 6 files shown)


heapster.go
```diff
@@ -34,16 +34,17 @@ import (
...
-	argModelResolution = flag.Duration("model_resolution", 1*time.Minute, "The resolution of the timeseries stored in the model. Applies only if use_model is true")
-	argPort            = flag.Int("port", 8082, "port to listen to")
-	argIp              = flag.String("listen_ip", "", "IP to listen on, defaults to all IPs")
-	argMaxProcs        = flag.Int("max_procs", 0, "max number of CPUs that can be used simultaneously. Less than 1 for default (number of cores)")
-	argSources         manager.Uris
-	argSinks           manager.Uris
+	argStatsResolution     = flag.Duration("stats_resolution", 5*time.Second, "The resolution at which heapster will retain stats.")
+	argSinkFrequency       = flag.Duration("sink_frequency", 10*time.Second, "Frequency at which data will be pushed to sinks")
+	argCacheDuration       = flag.Duration("cache_duration", 4*time.Minute, "The total duration of the historical data that will be cached by heapster.")
+	argRunningInKuberentes = flag.Bool("running_in_kubernetes", true, "When true, the kubernetes-specific api will be enabled")
```
vishh: Can we somehow autodetect this? For example, if one of the sources is kubernetes, we can use that as an hint. Default values for flags like this affect usability. WDYT?

vishh:
@mwielgus: Thank you for sending out this PR :)
Just one comment. Otherwise LGTM.

heapster.go
mwielgus: Added detection.

vishh:
Awesome! LGTM!

PR #590, opened 2015-09-21
unify the format of restful api & fix some typo error 

huangyuqi:
Some heapster's URLs of same type are end with different character. Some with "/", and another with not. Like this:
```
/api/v1/model/nodes/{node-name}/pods
```
different from:
```
/api/v1/model/nodes/{node-name}/metrics/
```
So , i want to unify it


vishh:
ok to test

vishh:
@huangyuqi: Thanks for the cleanup! LGTM

huangyuqi:
@vishh , thank you so much.

PR #593, opened 2015-09-21
Updating InfluxDB deployment to work with v0.8.9.

vishh:
cc @piosz


influxdb/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # influxdb
-FROM ubuntu:trusty
+FROM ubuntu
```
mvdan: I would still specify what version to use.

mvdan:
One nit, otherwise LGTM.

influxdb/Dockerfile
vishh: Done.

PR #594, opened 2015-09-21
Improve go vet output by reusing kubernetes code

mvdan:
This is a simplified version of the code in kubernetes that runs go vet.
Advantages over ours:
- Prints the errors instead of telling you to find them out for yourself
- Doesn't tell you to run `go fix` (fix only fixes usage of old apis)
- Doesn't hard-code k8s.io/heapster unnecessarily

### DIFF
--- a/hooks/run_vet.sh
+++ b/hooks/run_vet.sh
@@ -14,8 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


mvdan:
Correction - the current behaviour does print the errors, but in a messy way.

vishh:
LGTM.

PR #598, opened 2015-09-22
Docs fixes

mvdan:
CC @vishh


docs/sink-configuration.md
```diff
@@ -29,13 +26,11 @@ The following options are available:
...
 
-
 ### Google Cloud Monitoring
 This sink supports monitoring metrics only.
 To use the GCM sink add the following flag:
-```
---sink=gcm
-```
+
+	--sink=gcm
```
vishh: What is the markdown syntax here?

vishh:
LGTM. Thanks for the cleanup @mvdan. I can imagine how the existing docs should have annoyed you :)

docs/sink-configuration.md
mvdan: You can use four spaces or a tab to do code blocks just like you can with surroinding text with three backticks.

This is especially useful for very few blocks, since it requires two less lines and it's more readable in plaintext.

mvdan:
Yeah, that's pretty much why :)

PR #600, opened 2015-09-24
Enable TLS option & client certificate authentication

jimmidyson:
Heapster exposes unencrypted endpoints without authorization. In secure environments like OpenShift we'd like to secure the endpoints, both via TLS & by client certificate so that the master can use a certificate to authenticate for autoscaler stuff, etc.

This PR allows the configuration of TLS cert/key, client certificate checks & restricting users based on the CN of client cert. All optional of course.

Note that the godeps that have been updated are from a `godep restore` & `godep save ./... github.com/progrium/go-extpoints` to add in additional packages - I'm guessing some of the godep json had been updated manually prior to this.

/cc @mwringe @jcantrill @liggitt @pweil-

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -1,0 +1,115 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"k8s.io/kubernetes/pkg/auth/authenticator"
+	"k8s.io/kubernetes/pkg/auth/user"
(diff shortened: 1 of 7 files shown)


jimmidyson:
Let me sort out the godeps in a separate PR before even looking at this.

jimmidyson:
See #601 for clean up of godeps.

jimmidyson:
Rebased, ready for review.

auth.go
```diff
@@ -0,0 +1,115 @@
...
+package main
+
+import (
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/openshift/origin/pkg/auth/authenticator/request/x509request"
```
vishh: Just curious: Does kubernetes also import this package?

jimmidyson: No. The reason for using this package is to enable more configurable auth handlers in future. Right now we only see a need for client cert auth but there may be call for other auth methods in future or more granular authzn policy. We've learnt the hard way that it's better to put the auth behind interfaces now in case we want to enhance the auth policy later.

liggitt: Yes: https://github.com/kubernetes/kubernetes/blob/master/pkg/apiserver/authn.go#L31

liggitt: oops, misread... you should be importing the kubernetes package

jimmidyson: Ah OK thanks @liggitt my mistake.

jimmidyson: Sorted.

liggitt:
much better... still shouldn't need to pull in `bearertoken/bearertoken.go`

jimmidyson:
That's just a godep special... no real harm done ;)

vishh:
LGTM.

PR #601, opened 2015-09-24
Sanitized godeps

jimmidyson:
Cleaned up redundant files in godep workspace.

Ran:

```
godep restore
rm -rf Godeps
godep save ./... github.com/progrium/go-extpoints
```

to recreate vendored godeps.

Without this I was getting loads of issues adding in new dependencies - could have done by hand but would rather rely on tooling to do it.


vishh:
add to whitelist

mvdan:
LGTM, thanks @jimmidyson. I tried doing this after semi-manually updating the kubernetes dep, but it kept on removing necessary dependencies.

jimmidyson:
Yeah godep seems to have some problem with transitive side effect imports but doing it from a clean Godeps dir seemed to work nicely.

mvdan:
Oh, nice find.

Jenkins is failing in a weird way, but since all code and tests compile this is good to merge.

PR #607, opened 2015-09-24
Updating docs.

vishh:
cc @mvdan

### DIFF
--- a/docs/standalone.md
+++ b/docs/standalone.md
@@ -96,7 +96,7 @@ I0911 13:42:03.662594   21041 heapster.go:68] Starting heapster on port 8082
 If you got debug information as before, means that the heapster is working normally.You can test it by calling its [RESTful API](https://github.com/kubernetes/heapster/blob/master/docs/model.md).  
 You can use some restful tools to test, or just run a `curl` command (_ heapster implements gzip encoding, so we need add gunzip in command "curl"_).  
 
-	curl http://0.0.0.0:8082/api/v1/model/stats/ | gunzip
+	curl http://0.0.0.0:8082/api/v1/model/stats/
 
 and you will get the response like this:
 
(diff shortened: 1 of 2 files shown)


docs/standalone.md
```diff
@@ -104,17 +104,16 @@ I0911 13:42:03.657918   21041 heapster.go:58] Heapster version 0.17.0
...
-###Test it out
+
+### Test it out
 If you got debug information as before, means that the heapster is working normally.You can test it by calling its [RESTful API](https://github.com/kubernetes/heapster/blob/master/docs/model.md).  
 You can use some restful tools to test, or just run a `curl` command (_ heapster implements gzip encoding, so we need add gunzip in command "curl"_).  
 
-```shell
-curl http://0.0.0.0:8082/api/v1/model/stats/ | gunzip
-```
+	curl http://0.0.0.0:8082/api/v1/model/stats/ | gunzip
```
jimmidyson: So does heapster compress every response, even if the client hasn't requested compression?

vishh: Yes. That is the current state.

jimmidyson: Is that just due to the potential response size? We could just let go-restful handle the content encoding negotiation & let the client choose?

vishh: Yeah. The response size was the concern. It makes sense to let clients
handshake though.

jimmidyson: OK let me put in a PR for that.

vishh:
cc @mvdan

docs/standalone.md
jimmidyson: PR merged - can get rid of the `| gunzip` or add `-H "Accept-Encoding: gzip"` to the curl command.

mvdan:
After my docs fixes PR, this PR needs rebasing.

docs/standalone.md
vishh: Thanks @jimmidyson :100: 
Updated docs.

mvdan:
LGTM. Flaky test unrelated to this PR - investigating atm.

PR #608, opened 2015-09-24
Updating kubernetes source configs to include naemspace scope in DNS name

### LINKED ISSUE #586: heapster deployment with influxdb/grafana
As a beginner, I tried deploying heapster with influxdb/grafana following [guide](https://github.com/kubernetes/heapster/blob/master/docs/influxdb.md) and [manifests](https://github.com/kubernetes/heapster/tree/master/deploy/kube-config/influxdb). I think here might be a mistake:

`"--source=kubernetes:https://kubernetes"`
https://github.com/kubernetes/heapster/blob/master/deploy/kube-config/influxdb/heapster-controller.json#L33

If not change, there will be errors from heapster pod:
```
kubectl logs <heapster-podname>
...
E0919 09:17:33.649401       1 reflector.go:136] Failed to list *api.Node: Get https://kubernetes/api/v1/nodes: dial tcp: lookup kubernetes: no such host
```

I wonder the reason might be the "kubernetes" service only exists under "default" namespace, so the "heapster" under "kube-system" namespace can't reach to "kubernetes" through DNS/Kube-proxy because there is no route under "kube-system". Refer to heapster standalone deployment, I changed the `--source` to :
```
"--source=kubernetes"
```
Following: https://github.com/kubernetes/heapster/blob/master/deploy/kube-config/standalone/heapster-controller.json#L41
Then everything is fine till now.

vishh:
Fixes #586

### DIFF
--- a/deploy/kube-config/google/heapster-controller.json
+++ b/deploy/kube-config/google/heapster-controller.json
@@ -38,7 +38,7 @@
 			"name": "heapster",
 			"command": [
 			    "/heapster",
-			    "--source=kubernetes",
+			    "--source=kubernetes:https://kubernetes.default",
 			    "--sink=gcm",
 			    "--sink=gcl",
 			    "--poll_duration=2m",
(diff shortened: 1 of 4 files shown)


mvdan:
LGTM

PR #609, opened 2015-09-24
Enable content encoding negotation

jimmidyson:
Current approach of forcing gzip encoding isn't very friendly to clients. With this approach clients can use standard `Accept-Encoding` header to request compress, either gzip or deflate, as provided by go-restful.


vishh:
cc @mvdan

vishh:
Thanks for the fix @jimmidyson

mvdan:
LGTM, thanks for the fix!

PR #611, opened 2015-09-25
Make the model use a result struct to return stats

mvdan:
This makes the interface and functions cleaner, and allows us to add the
timestamp needed by features like the nodeMetrics experimental endpoint.

Extracted from #549.

CC @afein


vishh:
Thanks for the cleanup @mvdan! LGTM.

PR #612, opened 2015-09-26
Code cleanup and simplification

mvdan:
Ran multiple linters and checkers to see what they had to say.


vishh:
Once again much appreciated changed :+1: Should we add the linter to our pre-commit hooks?

vishh:
LGTM!

mvdan:
The linters have way too many false positives. They do much more than `go vet`, but the advantage of `go vet` is that it has no false positives.

So what I do is run them manually from time to time and apply some of their suggestions. They are definitely too heavy for pre-commit. As for CI, they would be a nightmare due to the many false positives.

mvdan:
As proof, one of the "code simplifications" that golint suggested broke the build :) undone.

vishh:
Ah! Got it :)

vishh:
Feel free to self merge once tests pass.

PR #614, opened 2015-09-27
Use yaml for config files

mvdan:
Note that the json->yaml conversion was automated via python. I only checked that the order of the fields was kept and that the files made sense, no more. Hoping the tests will still pass.


mvdan:
In fact, some of the fields weren't kept in the right order. I can manually fix that if it's a problem.

vishh:
A huge +1. This change helps keep configs across k8s and heapster repo consistent.

deploy/kube-config/google/heapster-service.json
```diff
@@ -1,23 +0,0 @@
-{
```
vishh: While you are here, should we symlink the service definitions instead of duplicating them?

vishh:
The order of the fields does not matter IMO. From a brief glance, the PR LGTM.

deploy/kube-config/google/heapster-service.json
mvdan: What about having it as `deploy/kube-config/heapster-service.json`, if it applies to all different configs?

vishh: +1

vishh:
LGTM

PR #615, opened 2015-09-28
Simplify Travis GOPATH handling

mvdan:


mvdan:
@vishh PTAL. Also, by looking at `integration/.jenkins.sh`, it looks like jenkins is cloning/getting the old `GoogleCloudPlatform/heapster` repo. If it's fixed there too, those lines could be removed.

vishh:
Didn't you clean up `.jenkins.sh`?

vishh:
This PR LGTM

mvdan:
I cleaned up `.jenkins.sh` regarding the make commands, but not the GOPATH handling as it requires changing how jenkins gets the source code. Which AFAIK is not handled by any shell script in this repo.

vishh:
Oh  yeah. Jenkins github plugins handle that. We can update the plugin to
use the vanity domain.

PR #627, opened 2015-10-02
Adding a dockerfile to release canary images on every commit.

vishh:
cc @mvdan

### DIFF
--- a/deploy/docker/canary/Dockerfile
+++ b/deploy/docker/canary/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.4.3
+MAINTAINER vishnuk@google.com
+
+
+RUN apt-get install -y git
+RUN git clone https://github.com/kubernetes/heapster.git /go/src/k8s.io/heapster
+RUN cd /go/src/k8s.io/heapster && make
+
+ENTRYPOINT ["/go/src/k8s.io/heapster/heapster"]


mvdan:
Now, why on earth is this jenkins failing...

Will investigate.

vishh:
@mvdan: I fixed it.

vishh:
I'm adding another jenkins project to push a docker image with canary tag after every merge to master.

deploy/docker/canary/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:latest
```
mvdan: Perhaps specify a version, namely 1.4.3?

deploy/docker/canary/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+MAINTAINER vishnuk@google.com
+
+
+RUN apt-get install -y git
+RUN git clone https://github.com/kubernetes/heapster.git /go/src/k8s.io/heapster
+RUN cd /go/src/k8s.io/heapster && make
+
+ENTRYPOINT ["/go/src/k8s.io/heapster/heapster"]
```
mvdan: You can probably simplify building and running by doing instead:

```
RUN apt-get install -y git
RUN go get k8s.io/heapster

ENTRYPOINT ["/go/bin/heapster"]
```

mvdan:
How did you fix jenkins?

Two nits, otherwise LGTM

vishh:
You can specify a custom checkout path in the git plugin.

deploy/docker/canary/Dockerfile
vishh: go get will not work because of `godep`

deploy/docker/canary/Dockerfile
vishh: fixed it.

vishh:
@mvdan: addressed comments.

vishh:
Merging this for now. We can fix the image later if required. I want to test the jenkins project.

deploy/docker/canary/Dockerfile
mvdan: You're right. In that case, instead of git clone you can do `go get -u -d k8s.io/heapster` instead of the git clone, but essentially it's the same.

PR #631, opened 2015-10-06
Update Hawkular-Sink to support filtering

burmanm:
This PR updates the Hawkular-Sink to support the following features:

* Password and username authentication
* If podId is missing (system metrics), use hostId in the naming
* Add filtering capabilities to avoid storing certain metrics

I think the filtering capabilities would be nice in the Heapster itself, however this would require a bit more planning and perhaps a query language designed with yacc or equal (this is very simple filtering language)

### DIFF
--- a/sinks/hawkular/driver.go
+++ b/sinks/hawkular/driver.go
@@ -19,7 +19,9 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -44,6 +46,30 @@ const (
 	defaultServiceAccountFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 )
 
+// Filter which Timeseries are stored. Store if returns true, filter if returns false
+type Filter func(*sink_api.Timeseries) bool
+type FilterType int
+
+const (
+	// Filter by label's value
+	Label FilterType = iota
+	// Filter by metric name
+	Name
+	// Unknown filter type
+	Unknown
+)
+
+func (f FilterType) From(s string) FilterType {
+	switch s {
+	case "label":
+		return Label
+	case "name":
+		return Name
+	default:
+		return Unknown
+	}
+}
+
 type hawkularSink struct {
 	client  *metrics.Client
 	models  map[string]*metrics.MetricDefinition // Model definitions
@@ -53,6 +79,8 @@ type hawkularSink struct {
 	uri *url.URL
 
 	labelTenant string
+	modifiers   []metrics.Modifier
+	filters     []Filter
 }
 
 // START: ExternalSink interface implementations
@@ -78,7 +106,11 @@ func (self *hawkularSink) Register(mds []sink_api.MetricDescriptor) error {
 
 // Fetches definitions from the server and checks that they're matching the descriptors
 func (self *hawkularSink) updateDefinitions(mt metrics.MetricType) error {
-	mds, err := self.client.Definitions(metrics.Filters(metrics.TypeFilter(mt)))
+	m := make([]metrics.Modifier, len(self.modifiers), len(self.modifiers)+1)
+	copy(m, self.modifiers)
+	m = append(m, metrics.Filters(metrics.TypeFilter(mt)))
+
+	mds, err := self.client.Definitions(m...)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


mvdan:
ok to test

sinks/hawkular/driver.go
```diff
@@ -44,6 +46,29 @@ const (
...
+	// Filter by label's value
+	Label FilterType = iota
+	// Filter by metric name
+	Name
+	// Unknown filter type
+	Unknown
+)
+
+func (f FilterType) From(s string) FilterType {
+	if s == "label" {
```
mvdan: switch?

sinks/hawkular/driver.go
```diff
@@ -153,7 +182,14 @@ func (self *hawkularSink) groupName(p *sink_api.Point) string {
 }
 
 func (self *hawkularSink) idName(p *sink_api.Point) string {
-	n := []string{p.Labels[sink_api.LabelContainerName.Key], p.Labels[sink_api.LabelPodId.Key], p.Name}
+	n := make([]string, 0, 3)
+	if p.Labels[sink_api.LabelPodId.Key] != "" {
+		n = append(n, p.Labels[sink_api.LabelContainerName.Key], p.Labels[sink_api.LabelPodId.Key], p.Name)
+	} else {
+		n = append(n, p.Labels[sink_api.LabelContainerName.Key], p.Labels[sink_api.LabelHostID.Key], p.Name)
+	}
```
mvdan: You could split up the append in three parts since two are exactly the same, as in:

```
n = append(n, p.Labels[sink_api.LabelContainerName.Key])
if p.Labels[sink_api.LabelPodId.Key] != "" {
    n = append(n, p.Labels[sink_api.LabelPodId.Key])
} else {
    n = append(n, p.Labels[sink_api.LabelHostID.Key])
}
n = append(n, p.Name)
```

sinks/hawkular/driver.go
```diff
@@ -153,7 +182,14 @@ func (self *hawkularSink) groupName(p *sink_api.Point) string {
...
+
+	// n := []string{p.Labels[sink_api.LabelContainerName.Key], p.Labels[sink_api.LabelPodId.Key], p.Name}
```
mvdan: remove this?

sinks/hawkular/driver.go
```diff
@@ -78,7 +105,9 @@ func (self *hawkularSink) Register(mds []sink_api.MetricDescriptor) error {
 
 // Fetches definitions from the server and checks that they're matching the descriptors
 func (self *hawkularSink) updateDefinitions(mt metrics.MetricType) error {
-	mds, err := self.client.Definitions(metrics.Filters(metrics.TypeFilter(mt)))
+	m := self.modifiers
```
mvdan: This will not make a full copy of the `[]metrics.Modifier`, is this on purpose?

sinks/hawkular/driver.go
```diff
@@ -78,7 +106,11 @@ func (self *hawkularSink) Register(mds []sink_api.MetricDescriptor) error {
 
 // Fetches definitions from the server and checks that they're matching the descriptors
 func (self *hawkularSink) updateDefinitions(mt metrics.MetricType) error {
-	mds, err := self.client.Definitions(metrics.Filters(metrics.TypeFilter(mt)))
+	m := make([]metrics.Modifier, len(self.modifiers), len(self.modifiers)+1)
```
jimmidyson: Couldn't this just be `m := append(self.modifiers, metrics.Tenant(k))` ?

mvdan: That doesn't copy the underlying data, fwiw.

burmanm: Yep, the underlying self.modifiers should not be modified in this case.

jimmidyson: Not that it'll make too much difference given the probable size of the slice, but do you actually want to copy the underlying data?

jimmidyson: `append` doesn't modify the underlying slice. Of course if stuff operates on the elements then that will be reflected in both slices which isn't what you want, but copying for the sake of it is a bit of a waste - but as it's a negligible optimisation I'm not going to argue the point.

jimmidyson:
LGTM - @mvdan any more comments before I merge?

mvdan:
LGTM once commits are squashed.

jimmidyson:
@mvdan Thanks.

@burmanm Could you please squash the commits then I'll merge?

jimmidyson:
Thanks for squashing - merging.

mvdan:
Thank you both!

PR #635, opened 2015-10-09
Introduced --model_frequency flag to define frequency of model updates

jszczepkowski:
Introduced --model_frequency flag to define frequency at which model is updated.

### DIFF
--- a/heapster.go
+++ b/heapster.go
@@ -40,6 +40,7 @@ var (
 	argCacheDuration   = flag.Duration("cache_duration", 4*time.Minute, "The total duration of the historical data that will be cached by heapster.")
 	argUseModel        = flag.Bool("use_model", true, "When true, the internal model representation will be used")
 	argModelResolution = flag.Duration("model_resolution", 1*time.Minute, "The resolution of the timeseries stored in the model. Applies only if use_model is true")
+	argModelFrequency  = flag.Duration("model_frequency", 145*time.Second, "Frequency at which model will be updated. Applies only if use_model is true")
 	argPort            = flag.Int("port", 8082, "port to listen to")
 	argIp              = flag.String("listen_ip", "", "IP to listen on, defaults to all IPs")
 	argMaxProcs        = flag.Int("max_procs", 0, "max number of CPUs that can be used simultaneously. Less than 1 for default (number of cores)")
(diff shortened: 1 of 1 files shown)


jszczepkowski:
CC @vishh @fgrzadkowski @piosz

piosz:
LGTM

piosz:
ok to test

mwielgus:
ok to test

vishh:
Should we validate this be flag? @afein mentioned that the updates to model are dependent on model resolution.

jszczepkowski:
> Should we validate this be flag? @afein mentioned that the updates to model are dependent on model resolution.

You are right, it should be higher. I've added validation.

jszczepkowski:
@vishh Can you order running integration tests for this PR? It seems that folks in Warsaw have no permissions to do it.

heapster.go
```diff
@@ -40,6 +40,7 @@ var (
 	argCacheDuration   = flag.Duration("cache_duration", 4*time.Minute, "The total duration of the historical data that will be cached by heapster.")
 	argUseModel        = flag.Bool("use_model", true, "When true, the internal model representation will be used")
 	argModelResolution = flag.Duration("model_resolution", 1*time.Minute, "The resolution of the timeseries stored in the model. Applies only if use_model is true")
+	argModelFrequency  = flag.Duration("model_frequency", 145*time.Second, "Frequency at which model will be updated. Applies only if use_model is true")
```
mwielgus: Why it has to be >2*modelResultion?

mwielgus: Actually I think there is no reason for it to be > 2* modelResolution. If model resolution is 1 minute then it should sync every minute with the cache. Timestamp checking in update should do the rest (and if it doesn't then it should be fixed ASAP).

vishh:
add to whitelist

heapster.go
vishh: https://github.com/kubernetes/heapster/issues/642#issuecomment-146935609

vishh:
@jszczepkowski: I added you to the whitelist. Jenkins should run e2e for your PRs henceforth.

vishh:
LGTM.

heapster.go
mwielgus: Can we make it 1 min + offset?

jszczepkowski: It seems not: https://github.com/kubernetes/heapster/issues/642#issuecomment-146972283.

PR #644, opened 2015-10-09
Collect one stat sample every housekeeping.

### LINKED ISSUE #638: Heapster gets 1 sample/second from cAdvisor
Even if the stats resolution is 30 seconds. We drop 29 samples but they consume both network(transmission) and cpu&mem (parsing).

vishh:
Fixes #638 

cc @mwielgus @piosz

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -144,8 +144,11 @@ func (rm *realManager) HousekeepModel() {
 
 func (rm *realManager) Housekeep() {
 	for {
-		start := rm.lastSync
-		end := start.Add(rm.resolution)
+		end := rm.lastSync.Add(rm.resolution)
+		// We need only one sample every housekeeping. Issue #638.
+		// Collecting two samples to work-around situations where
+		// the source might have skipped a second.
+		start := end.Add(-2 * time.Second)
 		timeToNextSync := end.Sub(time.Now())
 		// TODO: consider adding some delay here
 		select {
--- a/sources/datasource/cadvisor.go
+++ b/sources/datasource/cadvisor.go
@@ -30,6 +30,7 @@ import (
 type cadvisorSource struct{}
 
 func (self *cadvisorSource) parseStat(containerInfo *cadvisor.ContainerInfo) *api.Container {
+	glog.V(5).Infof("got %d stats for container %q", len(containerInfo.Stats), containerInfo.Name)
 	container := &api.Container{
 		Name:  containerInfo.Name,
 		Spec:  api.ContainerSpec{ContainerSpec: containerInfo.Spec},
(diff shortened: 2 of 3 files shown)


manager/manager.go
```diff
@@ -144,8 +144,9 @@ func (rm *realManager) HousekeepModel() {
 
 func (rm *realManager) Housekeep() {
 	for {
-		start := rm.lastSync
-		end := start.Add(rm.resolution)
+		end := rm.lastSync.Add(rm.resolution)
+		// We need only one sample every housekeeping. Issue #638.
+		start := end.Add(-time.Second)
```
mwielgus: What if cadvisor didn't manage to catch the sample in the last second (or the samples are offseted slightly).

vishh: We rely on cAdvisor's predictable performance as of now. 
To be on the safe side, increasing the window a bit SGTM.

>Also - will 1 sec window work with metrics compaction/deduplication strategy in cAdvisor?
I don't understand this question. Can you elaborate?

vishh: I also think the right solution here will be for cAdvisor or kubelet to support resolution.

mwielgus: So, I head from @piosz that cAdvisor/Kubelet doesn't generate a new sample if the new one is more-or-less equal to the the previous and the previous was taken in the last 15 sec or so. I'm not sure where and how this logic is exactly implemented.  It is also possible that I misunderstood something.

piosz:
AFAIK this change won't help because NumStats field is ignored if Start and End fields are specified. See more details here:
https://github.com/google/cadvisor/blob/master/info/v1/container.go#L94

jimmidyson:
Yeah - see https://github.com/google/cadvisor/blob/893a5c15f59972d5ddb4d61e2630863ce6a98dd7/utils/timed_store.go#L95-L98

vishh:
@piosz: Thanks for pointing out. I will remove numstats. This change also reduces the time window. As of now, processing two stat samples is better than 30 of them. WDYT?

piosz:
In this case you should decrease [maxHousekeepingInterval](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/cadvisor_linux.go#L47) to assure we don't omit any stats. I'm not sure what would be the outcome of this change.

@vmarmol @rjnagal any thoughts?

vishh:
The dynamic housekeeping feature was added to reduce `cpu usage` of
cAdvisor. If we have empty cgroups, stats will never change for those
cgroups and cAdvisor spent cpu cycles on them.
From that perspective, if we do not receive a stat sample in a time
duration, we can safely assume that the previous metric values still hold
good. On the flipside, it sounds like a hack :)

piosz:
@vishh I know, but we don't have mechanism implemented which will use the previous value in case of lack the current one, so indeed there will be gaps in metrics. This will cause incorrect computation of autoscaling metrics.

piosz:
This will be fix by introducing Metrics API in Kubeletet https://github.com/kubernetes/kubernetes/issues/12483

Issue #227, opened 2015-04-10
Add a Elasticsearch Sink


vishh:
Elasticsearch is well suited for handling events. It might be useful for metrics as well, but not a priority. Kubernetes runs a ELK setup by default.

rohitsardesai83:
I am interested in contributing towards this. Do let me know I can pick this up . Thanks !

piosz:
PR in flight #733

huangyuqi:
thanks @piosz 
I will rebase #733 , and fix it.

piosz:
Thanks @huangyuqi. Looking forward for you change.

djsly:
HI @huangyuqi , we are also interested in the this PR. We though it was dead and planned in trying to see how easy it was to bring it back to life. Happy to see that you are back. Looking forward to this change!

titilambert:
Hello !
@huangyuqi I don't know if you already started the work, but due to my job deadlines, I need it quickly.
So, I started the merge for ES sink from your commits.
It's here: https://github.com/titilambert/heapster/tree/sink_es
Events are working. Metrics and tests are missing...

titilambert:
PR created here: https://github.com/kubernetes/heapster/pull/1144

huangyuqi:
I will update the PR #733 today
BTW, @titilambert , in PR #1144 , you can cherry-pick my commit .

titilambert:
@huangyuqi It seems I was not clear. I already did the job: I cherry-pick your commits and I merge it in 2 commits and create PR #1144.
It just missing the metrics parts (which is not really useful... generally people use influxDB for metrics)
You can rebase on PR #1144, before adding your changes, in order to avoid to redo the done work...

huangyuqi:
@piosz I have done the ES sink in #733, please review.

piosz:
#733 is merged.

PR #648, opened 2015-10-12
kafka sink 

huangyuqi:
Apache Kafka is publish-subscribe messaging rethought as a distributed commit log.
Kafka sink just transforms the monitoring metrics and events to kafka message with two topics: timeseriestopic, eventstopic.

To use the kafka sink add the following flag:
```
--sink=kafka:<KAFKA_SERVER_URL>[?<OPTIONS>]
```
Normally, kafka server has multi brokers, so brokers' list need be configured for producer.
So, we can set `KAFKA_SERVER_URL` to a dummy value, and provide kafka brokers' list in url's query string.
Besides,the following options need be set in query string:

* `timeseriestopic` - Kafka's topic for timeseries
* `eventstopic` - Kafka's topic for events

Like this:
```
--sink=kafka:http://kafka/?brokers=0.0.0.0:9092&brokers=0.0.0.0:9093&timeseriestopic=test&eventstopic=test 
```

### DIFF
--- a/docs/sink-configuration.md
+++ b/docs/sink-configuration.md
@@ -78,6 +78,25 @@ The following options are available:
 * `tenant` - Hawkular-Metrics tenantId (default: `heapster`)
 * `labelToTenant` - Hawkular-Metrics uses given label's value as tenant value when storing data
 
+### Kafka
+This sink supports monitoring metrics and events.
+To use the kafka sink add the following flag:
+
+```
+--sink=kafka:<KAFKA_SERVER_URL>[?<OPTIONS>]
+```
+Normally, kafka server has multi brokers, so brokers' list need be configured for producer.
+So, we can set `KAFKA_SERVER_URL` to a dummy value, and provide kafka brokers' list in url's query string.
+Besides,the following options need be set in query string:
+
+* `timeseriestopic` - Kafka's topic for timeseries
+* `eventstopic` - Kafka's topic for events
+
+Like this:
+```
+--sink=kafka:http://kafka/?brokers=0.0.0.0:9092&brokers=0.0.0.0:9093&timeseriestopic=test&eventstopic=test
+```
+
 ## Modifying the sinks at runtime
 
 Using the `/api/v1/sinks` endpoint, it is possible to fetch the sinks
--- a/docs/sink-configuration.md
+++ b/docs/sink-configuration.md
@@ -82,9 +82,8 @@ The following options are available:
 This sink supports monitoring metrics and events.
 To use the kafka sink add the following flag:
 
-```
---sink=kafka:<KAFKA_SERVER_URL>[?<OPTIONS>]
-```
+    --sink=kafka:<KAFKA_SERVER_URL>[?<OPTIONS>]
+
 Normally, kafka server has multi brokers, so brokers' list need be configured for producer.
 So, we can set `KAFKA_SERVER_URL` to a dummy value, and provide kafka brokers' list in url's query string.
 Besides,the following options need be set in query string:
@@ -93,9 +92,8 @@ Besides,the following options need be set in query string:
 * `eventstopic` - Kafka's topic for events
 
 Like this:
-```
---sink=kafka:http://kafka/?brokers=0.0.0.0:9092&brokers=0.0.0.0:9093&timeseriestopic=test&eventstopic=test
-```
+
+    --sink=kafka:http://kafka/?brokers=0.0.0.0:9092&brokers=0.0.0.0:9093&timeseriestopic=test&eventstopic=test
 
 ## Modifying the sinks at runtime
 
--- a/sinks/kafka/driver.go
+++ b/sinks/kafka/driver.go
@@ -32,7 +32,7 @@ const (
 	partition                = 0
 	brokerClientID           = "kafka-sink"
 	brokerDialTimeout        = 10 * time.Second
-	brokerDialRetryLimit     = 1
+	brokerDialRetryLimit     = 10
 	brokerDialRetryWait      = 500 * time.Millisecond
 	brokerAllowTopicCreation = true
 	brokerLeaderRetryLimit   = 10
@@ -116,7 +116,10 @@ func init() {
 
 func NewKafkaSink(uri *url.URL, _ extpoints.HeapsterConf) ([]sink_api.ExternalSink, error) {
 	var kafkaSink kafkaSink
-	opts, _ := url.ParseQuery(uri.RawQuery)
+	opts, err := url.ParseQuery(uri.RawQuery)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parser url's query string: %s", err)
+	}
 
 	if len(opts["timeseriestopic"]) < 1 {
 		return nil, fmt.Errorf("There is no timeseriestopic assign for config kafka-sink")
@@ -144,8 +147,7 @@ func NewKafkaSink(uri *url.URL, _ extpoints.HeapsterConf) ([]sink_api.ExternalSi
 
 	broker, err := kafka.Dial(kafkaSink.sinkBrokerHosts, brokerConf)
 	if err != nil {
-		glog.Errorf("connect to kafka cluster fail: %s", err)
-		return nil, fmt.Errorf("connect to kafka cluster fail: %s", err)
+		return nil, fmt.Errorf("failed to connect to kafka cluster: %s", err)
 	}
 	defer broker.Close()
 
--- a/sinks/kafka/driver.go
+++ b/sinks/kafka/driver.go
@@ -1,0 +1,159 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kafka
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/optiopay/kafka"
+	"github.com/optiopay/kafka/proto"
+	"k8s.io/heapster/extpoints"
+	sink_api "k8s.io/heapster/sinks/api"
+	kube_api "k8s.io/kubernetes/pkg/api"
+)
+
+const (
+	partition                int32         = 0
+	brokerClientID           string        = "kafka-sink"
+	brokerDialTimeout        time.Duration = 10 * time.Second
+	brokerDialRetryLimit     int           = 10
+	brokerDialRetryWait      time.Duration = 500 * time.Millisecond
+	brokerAllowTopicCreation bool          = true
+	brokerLeaderRetryLimit   int           = 10
+	brokerLeaderRetryWait    time.Duration = 500 * time.Millisecond
+)
+
+type kafkaSink struct {
+	producer        kafka.Producer
+	timeSeriesTopic string
+	eventsTopic     string
+	sinkBrokerHosts []string
+}
+
+// START: ExternalSink interface implementations
+
+func (self *kafkaSink) Register(mds []sink_api.MetricDescriptor) error {
+	return nil
+}
+
+func (self *kafkaSink) Unregister(mds []sink_api.MetricDescriptor) error {
+	return nil
+}
+
+func (self *kafkaSink) StoreTimeseries(timeseries []sink_api.Timeseries) error {
+	if timeseries == nil || len(timeseries) <= 0 {
+		return nil
+	}
+	for _, t := range timeseries {
+		err := self.produceKafkaMessage(t, self.timeSeriesTopic)
+		if err != nil {
+			return fmt.Errorf("failed to produce Kafka messages: %s", err)
+		}
+	}
+	return nil
+}
+
+func (self *kafkaSink) StoreEvents(events []kube_api.Event) error {
+	if events == nil || len(events) <= 0 {
+		return nil
+	}
+	for _, event := range events {
+		err := self.produceKafkaMessage(event, self.eventsTopic)
+		if err != nil {
+			return fmt.Errorf("failed to produce Kafka messages: %s", err)
+		}
+	}
+	return nil
+}
+
+// produceKafkaMessage produces messages to kafka
+func (self *kafkaSink) produceKafkaMessage(v interface{}, topic string) error {
+	if v == nil {
+		return nil
+	}
+	jsonItems, err := json.Marshal(v)
+	if err != nil {
+		return fmt.Errorf("failed to transform the items to json : %s", err)
+	}
+	message := &proto.Message{Value: []byte(string(jsonItems))}
+	_, err = self.producer.Produce(topic, partition, message)
+	if err != nil {
+		return fmt.Errorf("failed to produce message to %s:%d: %s", topic, partition, err)
+	}
+	return nil
+}
+
+func (self *kafkaSink) DebugInfo() string {
+	info := fmt.Sprintf("%s\n", self.Name())
+	info += fmt.Sprintf("There are two kafka's topics: %s,%s:\n", self.eventsTopic, self.timeSeriesTopic)
+	info += fmt.Sprintf("The kafka's broker list is: %s", self.sinkBrokerHosts)
+	return info
+}
+
+func (self *kafkaSink) Name() string {
+	return "Apache-Kafka Sink"
+}
+
+func init() {
+	extpoints.SinkFactories.Register(NewKafkaSink, "kafka")
+}
+
+func NewKafkaSink(uri *url.URL, _ extpoints.HeapsterConf) ([]sink_api.ExternalSink, error) {
+	var kafkaSink kafkaSink
+	opts, _ := url.ParseQuery(uri.RawQuery)
+
+	if len(opts["timeseriestopic"]) < 1 {
+		return nil, fmt.Errorf("There is no timeseriestopic assign for config kafka-sink")
+	}
+	kafkaSink.timeSeriesTopic = opts["eventstopic"][0]
+
+	if len(opts["eventstopic"]) < 1 {
+		return nil, fmt.Errorf("There is no eventstopic assign for config kafka-sink")
+	}
+	kafkaSink.eventsTopic = opts["eventstopic"][0]
+
+	if len(opts["brokers"]) < 1 {
+		return nil, fmt.Errorf("There is no broker assign for connecting kafka broker")
+	}
+	kafkaSink.sinkBrokerHosts = append(kafkaSink.sinkBrokerHosts, opts["brokers"]...)
+
+	//connect to kafka cluster
+	brokerConf := kafka.NewBrokerConf(brokerClientID)
+	brokerConf.DialTimeout = brokerDialTimeout
+	brokerConf.DialRetryLimit = brokerDialRetryLimit
+	brokerConf.DialRetryWait = brokerDialRetryWait
+	brokerConf.LeaderRetryLimit = brokerLeaderRetryLimit
+	brokerConf.LeaderRetryWait = brokerLeaderRetryWait
+	brokerConf.AllowTopicCreation = true
+
+	broker, err := kafka.Dial(kafkaSink.sinkBrokerHosts, brokerConf)
+	if err != nil {
+		glog.Errorf("connect to kafka cluster fail: %s", err)
+		return nil, fmt.Errorf("connect to kafka cluster fail: %s", err)
+	}
+	defer broker.Close()
+
+	//create kafka producer
+	conf := kafka.NewProducerConf()
+	conf.RequiredAcks = proto.RequiredAcksLocal
+	sinkProducer := broker.Producer(conf)
+	kafkaSink.producer = sinkProducer
+	glog.Infof("created kafka sink successfully with brokers: %v", kafkaSink.sinkBrokerHosts)
+	return []sink_api.ExternalSink{&kafkaSink}, nil
+}
--- a/sinks/kafka/driver.go
+++ b/sinks/kafka/driver.go
@@ -29,14 +29,14 @@ import (
 )
 
 const (
-	partition                int32         = 0
-	brokerClientID           string        = "kafka-sink"
-	brokerDialTimeout        time.Duration = 10 * time.Second
-	brokerDialRetryLimit     int           = 10
-	brokerDialRetryWait      time.Duration = 500 * time.Millisecond
-	brokerAllowTopicCreation bool          = true
-	brokerLeaderRetryLimit   int           = 10
-	brokerLeaderRetryWait    time.Duration = 500 * time.Millisecond
+	partition                = 0
+	brokerClientID           = "kafka-sink"
+	brokerDialTimeout        = 10 * time.Second
+	brokerDialRetryLimit     = 1
+	brokerDialRetryWait      = 500 * time.Millisecond
+	brokerAllowTopicCreation = true
+	brokerLeaderRetryLimit   = 10
+	brokerLeaderRetryWait    = 500 * time.Millisecond
 )
 
 type kafkaSink struct {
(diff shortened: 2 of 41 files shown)


vishh:
ok to test

huangyuqi:
https://github.com/Huawei-PaaS/heapster/commit/e081929cbe942d32bc82025420935bfbfd06e672 this commit just adds some contents for docs, but some checks were not successful.  So i want to reopen it.

vishh:
ok to test

huangyuqi:
Thanks , @vishh , would you please kindly review this PR? : )

huangyuqi:
Thanks , @vishh , would you please kindly review this PR? : )

mvdan:
@huangyuqi if this PR is ready on your end, please remove WIP from the title

docs/sink-configuration.md
```diff
@@ -78,6 +78,25 @@ The following options are available:
 * `tenant` - Hawkular-Metrics tenantId (default: `heapster`)
 * `labelToTenant` - Hawkular-Metrics uses given label's value as tenant value when storing data
 
+### Kafka
+This sink supports monitoring metrics and events.
+To use the kafka sink add the following flag:
+
+```
+--sink=kafka:<KAFKA_SERVER_URL>[?<OPTIONS>]
```
mvdan: Please indent via four spaces if it's just one line.

docs/sink-configuration.md
```diff
@@ -78,6 +78,25 @@ The following options are available:
...
+Normally, kafka server has multi brokers, so brokers' list need be configured for producer.
+So, we can set `KAFKA_SERVER_URL` to a dummy value, and provide kafka brokers' list in url's query string.
+Besides,the following options need be set in query string:
+
+* `timeseriestopic` - Kafka's topic for timeseries
+* `eventstopic` - Kafka's topic for events
+
+Like this:
+```
+--sink=kafka:http://kafka/?brokers=0.0.0.0:9092&brokers=0.0.0.0:9093&timeseriestopic=test&eventstopic=test
```
mvdan: Same as above.

sinks/kafka/driver.go
```diff
@@ -0,0 +1,166 @@
...
+
+const (
+	partition                int32         = 0
+	brokerClientID           string        = "kafka-sink"
+	brokerDialTimeout        time.Duration = 10 * time.Second
+	brokerDialRetryLimit     int           = 10
+	brokerDialRetryWait      time.Duration = 500 * time.Millisecond
+	brokerAllowTopicCreation bool          = true
+	brokerLeaderRetryLimit   int           = 10
+	brokerLeaderRetryWait    time.Duration = 500 * time.Millisecond
```
mvdan: None of these require typing.

sinks/kafka/driver.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (self *kafkaSink) StoreEvents(events []kube_api.Event) error {
+	if events == nil || len(events) <= 0 {
+		return nil
+	}
+	for index := range events {
+		err := self.produceKafkaMessage(events[index], self.eventsTopic)
+		if err != nil {
+			glog.Errorf("failed to produce Kafka messages: %s", err)
+			return fmt.Errorf("failed to produce Kafka messages: %s", err)
```
mvdan: I see logs and errors duplicated everywhere. Is this necessary? If it is, please deduplicate the strings.

sinks/kafka/driver_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+package kafka
+
+import (
+	"encoding/json"
+	"github.com/optiopay/kafka/proto"
+	"github.com/stretchr/testify/assert"
+	sink_api "k8s.io/heapster/sinks/api"
+	kube_api "k8s.io/kubernetes/pkg/api"
+	kube_api_unv "k8s.io/kubernetes/pkg/api/unversioned"
```
mvdan: Please run `goimports -l -w .` to fix the import grouping.

sinks/kafka/driver_test.go
```diff
@@ -0,0 +1,236 @@
...
+type fakeKafkaSink struct {
+	sink_api.ExternalSink
+	fakeProducer *fakeKafkaProducer
+}
+
+func NewFakeKafkaProducer() *fakeKafkaProducer {
+	return &fakeKafkaProducer{[]msgProducedToKafka{}}
+}
+
+func (producer *fakeKafkaProducer) Produce(topic string, partition int32, messages ...*proto.Message) (offset int64, err error) {
```
mvdan: You can drop the named returns.

sinks/kafka/driver.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (self *kafkaSink) Unregister(mds []sink_api.MetricDescriptor) error {
+	return nil
+}
+
+func (self *kafkaSink) StoreTimeseries(timeseries []sink_api.Timeseries) error {
+	if timeseries == nil || len(timeseries) <= 0 {
+		return nil
+	}
+	for index := range timeseries {
```
mvdan: This loop and all the rest are a bit weird. Why not just do `for _, t := range timeseries` and then `t` instead of `timeseries[index]`?

sinks/kafka/driver.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil
+	}
+	jsonItems, err := json.Marshal(v)
+	if err != nil {
+		glog.Errorf("failed to transform the items to json : %s", err)
+		return fmt.Errorf("failed to transform the items to json : %s", err)
+	}
+	message := &proto.Message{Value: []byte(string(jsonItems))}
+
+	if err != nil {
```
mvdan: What is this checking? The last generated `err` was already checked.

sinks/kafka/driver.go
```diff
@@ -0,0 +1,166 @@
...
+func NewKafkaSink(uri *url.URL, _ extpoints.HeapsterConf) ([]sink_api.ExternalSink, error) {
+	var kafkaSink kafkaSink
+	opts, _ := url.ParseQuery(uri.RawQuery)
+	if len(opts["timeseriestopic"]) >= 1 && len(opts["eventstopic"]) >= 1 {
+		kafkaSink.timeSeriesTopic = opts["timeseriestopic"][0]
+		kafkaSink.eventsTopic = opts["eventstopic"][0]
+	} else {
+		glog.Errorf("There is no topic assign for config kafka-sink")
+		return nil, fmt.Errorf("There is no topic assign for config kafka-sink")
+	}
```
mvdan: See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow - the else is unnecessary.

docs/sink-configuration.md
huangyuqi: @mvdan ,thanks for your comments so much. The adding lines of Kafka sink are copied from the above sinks to keep pace with their format. So i think these lines are not one line, i will retain this. : )

sinks/kafka/driver_test.go
huangyuqi: OK

sinks/kafka/driver.go
huangyuqi: thanks @mvdan,  but I cant catch your meaning, can you tell me more clearly?  : )

sinks/kafka/driver_test.go
huangyuqi: thanks @mvdan,  but I cant catch your meaning, can you tell me more clearly?  : )

sinks/kafka/driver.go
jimmidyson: I think what he means is you can drop the type declaration as type will be inferred.

mvdan: Not exactly - constants are only typed when used, not when declared. So it's a simplification of the code, but also dropping an unnecessary strict typing. For example, you could use `10` as both `int` and `uint`, but if you specify `int`, you can only use it for the former.

docs/sink-configuration.md
mvdan: Note that the format was changed weeks ago, the other sink configuration lines are indented with four spaces.

sinks/kafka/driver.go
jimmidyson: Consider me educated! Thanks!

sinks/kafka/driver_test.go
mvdan: Named returns are only useful for naked returns, which you don't use here. So return `(int64, error)`.

See https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters.

mvdan:
You applied some of the fixes by amending your original commit, i.e. rebasing the history. Please don't do that until the PR is ready, so that the history of changes is kept.

sinks/kafka/driver.go
mvdan: Lots of interesting stuff at blog.golang.org :) https://blog.golang.org/constants

docs/sink-configuration.md
huangyuqi: got it, thanks so much. : )

sinks/kafka/driver.go
```diff
@@ -0,0 +1,159 @@
...
+	brokerConf.DialRetryLimit = brokerDialRetryLimit
+	brokerConf.DialRetryWait = brokerDialRetryWait
+	brokerConf.LeaderRetryLimit = brokerLeaderRetryLimit
+	brokerConf.LeaderRetryWait = brokerLeaderRetryWait
+	brokerConf.AllowTopicCreation = true
+
+	broker, err := kafka.Dial(kafkaSink.sinkBrokerHosts, brokerConf)
+	if err != nil {
+		glog.Errorf("connect to kafka cluster fail: %s", err)
+		return nil, fmt.Errorf("connect to kafka cluster fail: %s", err)
```
mvdan: This duplicate error string remains.

sinks/kafka/driver_test.go
```diff
@@ -0,0 +1,237 @@
...
+			MetricDescriptor: &smd,
+			Point:            &p2,
+		},
+	}
+
+	err := fakeSink.StoreTimeseries(timeseries)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(fakeSink.fakeProducer.msgs))
+	timeseries0Json, _ := json.Marshal(timeseries[0])
+	timeseries1Json, _ := json.Marshal(timeseries[1])
```
mvdan: I'm not sure whether ignoring errors here makes sense. Maybe panic if `err != nil`, similar to `regexp.MustCompile`?

sinks/kafka/driver.go
```diff
@@ -0,0 +1,159 @@
...
+func (self *kafkaSink) Register(mds []sink_api.MetricDescriptor) error {
+	return nil
+}
+
+func (self *kafkaSink) Unregister(mds []sink_api.MetricDescriptor) error {
+	return nil
+}
+
+func (self *kafkaSink) StoreTimeseries(timeseries []sink_api.Timeseries) error {
+	if timeseries == nil || len(timeseries) <= 0 {
```
mvdan: `len()` of a nil slice is zero, so the test against nil is redundant.

sinks/kafka/driver.go
```diff
@@ -0,0 +1,159 @@
...
+		err := self.produceKafkaMessage(t, self.timeSeriesTopic)
+		if err != nil {
+			return fmt.Errorf("failed to produce Kafka messages: %s", err)
+		}
+	}
+	return nil
+}
+
+func (self *kafkaSink) StoreEvents(events []kube_api.Event) error {
+	if events == nil || len(events) <= 0 {
```
mvdan: Same as above.

sinks/kafka/driver.go
```diff
@@ -0,0 +1,159 @@
...
+	return "Apache-Kafka Sink"
+}
+
+func init() {
+	extpoints.SinkFactories.Register(NewKafkaSink, "kafka")
+}
+
+func NewKafkaSink(uri *url.URL, _ extpoints.HeapsterConf) ([]sink_api.ExternalSink, error) {
+	var kafkaSink kafkaSink
+	opts, _ := url.ParseQuery(uri.RawQuery)
```
mvdan: Ignored error?

sinks/kafka/driver.go
huangyuqi: thanks, it's indeed thoughtless.I will remove it.

sinks/kafka/driver.go
huangyuqi: OK,i will remove, thanks.

sinks/kafka/driver.go
huangyuqi: i will remove. thanks : )

sinks/kafka/driver_test.go
huangyuqi: It's my bad coding habit. T_T.  
I will add a `assert` to deal with the err. Like this: 
```golang
assert.NoError(t, err)
```

sinks/kafka/driver.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func NewKafkaSink(uri *url.URL, _ extpoints.HeapsterConf) ([]sink_api.ExternalSink, error) {
+	var kafkaSink kafkaSink
+	opts, err := url.ParseQuery(uri.RawQuery)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parser url's query string: %s", err)
+	}
+
+	if len(opts["timeseriestopic"]) < 1 {
```
mvdan: This is really nitpicky, but usually people test if strings are empty like `if s == ""`

mvdan:
Thanks for all the fixes! The code looks a lot better now. Someone with a better understanding of kafka can probably speak as to the quality of the implementation.

huangyuqi:
Thanks @mvdan for your meticulous comments.

mvdan:
Note that four comments are still standing - they will be hidden once fixed.

sinks/kafka/driver.go
huangyuqi: @mvdan , thanks for your attentive comments. Here, I think the `opt` of query string is often an array, so i test it by `len()`.  Do you agree with me?  ^_^

huangyuqi:
@mvdan , thanks for your attentive comments.  In the four remaining comments: 
1. https://github.com/kubernetes/heapster/pull/648#discussion-diff-42236519R18 : i have dealed with `goimports`. ^-^
2~3. https://github.com/kubernetes/heapster/pull/648#discussion-diff-42524964R59       https://github.com/kubernetes/heapster/pull/648#discussion-diff-42524964R59   i have removed the corresponding testcase, so it may not be hidden here. ^-^
4.    https://github.com/kubernetes/heapster/pull/648#discussion-diff-42626989R124 I think the opt of query string is often an array, so i test it by len().

sinks/kafka/driver.go
```diff
@@ -0,0 +1,161 @@
...
+// produceKafkaMessage produces messages to kafka
+func (self *kafkaSink) produceKafkaMessage(v interface{}, topic string) error {
+	if v == nil {
+		return nil
+	}
+	jsonItems, err := json.Marshal(v)
+	if err != nil {
+		return fmt.Errorf("failed to transform the items to json : %s", err)
+	}
+	message := &proto.Message{Value: []byte(string(jsonItems))}
```
vishh: You are exposing an internal data structure [`TimeSeries`](https://github.com/kubernetes/heapster/blob/master/sinks/api/types.go#L159) over the wire here. If we update the data structure in the future, this will break all client of this sink. That is not desirable. 

What is the minimum amount of data that you need? Are there any standard formats for exposing metrics and events kind of data through kafka?

sinks/kafka/driver.go
```diff
@@ -0,0 +1,161 @@
...
+	//connect to kafka cluster
+	brokerConf := kafka.NewBrokerConf(brokerClientID)
+	brokerConf.DialTimeout = brokerDialTimeout
+	brokerConf.DialRetryLimit = brokerDialRetryLimit
+	brokerConf.DialRetryWait = brokerDialRetryWait
+	brokerConf.LeaderRetryLimit = brokerLeaderRetryLimit
+	brokerConf.LeaderRetryWait = brokerLeaderRetryWait
+	brokerConf.AllowTopicCreation = true
+
+	broker, err := kafka.Dial(kafkaSink.sinkBrokerHosts, brokerConf)
```
vishh: Do you want to retry here? In a typical cluster, components can go down anytime. If you were to keep re-trying periodically to connect to kafka hoping for it to come up, that might provide better user experience. WDYT?

vishh:
Sorry for the delay @huangyuqi! Thanks for the PR! I have posted a couple of comments. Other than that, this PR LGTM. Lets get this in quickly.

sinks/kafka/driver.go
huangyuqi: @vishh ,thanks for your comments. Retry here is used for connecting kafka brokers  for the first time, by importing an opensoure kafka client. 
And your opinion is to check the link to kafka periodically, during the sink process, right?

sinks/kafka/driver.go
huangyuqi: Thanks for your comments.  @vishh .
I need all the ops information, includes metrics and events. 
Exposing an internal data structure TimeSeries over the wire is not well advised. I will pick out the effective information about metrics and events to give the kafka sink.Include: timestamp, value of metrics and events.

sinks/kafka/driver.go
vishh: Yeah. If the client doesn't attempt to connect to kafka brokers here, that is fine. If it attempts to establish connectivity to brokers, chances are it will fail most of the time, if kafka is also running as a cluster service.

vishh: You can take a look at the influxdb sink which retries instead of failing hard.

sinks/kafka/driver.go
vishh: SGTM. You can create a versioned structure in this sink.

huangyuqi: @vishh , Sorry for the delay. I have add two internal data structures for kafka sink. Would you please check again? Thanks  ^_^

sinks/kafka/driver.go
huangyuqi: @vishh , I think that setting the retry time to 1 is fine, I have modified. If the kafka server is not ready, the sink will return immediately.  : )

vishh:
@huangyuqi: Is this PR ready for review? Have you addressed the pending comments?

huangyuqi:
@vishh : this PR is ready, I have fixed all the comments. : )

huangyuqi:
@vishh , Would you please review the PR of kafka sink for some modification about data structure and kafka's client? : )
BTW, Our OPS subsystem uses the ES to collect the metrics and events,  so i realize a simple ES's sink. May I take another new PR about this ?

sinks/kafka/driver.go
vishh: If the brokers do not respond within the timeout, will the sink setup fail? I'd expect the sink to continue setup, and keep retrying to establish connection. From the code, that is not clear to me. Can you document the behavior using comments?

vishh:
> @vishh , Would you please review the PR of kafka sink for some modification about data structure and kafka's client? : )

The part about sink setup isn't clear to me yet. Once you clarify the behavior there, this PR is good to go.

> BTW, Our OPS subsystem uses the ES to collect the metrics and events, so i realize a simple ES's sink. May I take another new PR about this ?

SGTM! Just curious - does ES support retention policies?

huangyuqi:
Thanks, @vishh :
**sink setup**
+ Kafka is run as a cluster comprised of one or more servers each of which is called a broker; So we need to assign the broker list to the sink, with the query string `brokers`
+ Kafka maintains feeds of messages in categories called topics. So the topics should also be specified, with the query string `timeseriestopic` and `eventstopic`
+ In kafka sink, we use a client to connect these brokers, with the below config:
```go
type BrokerConf struct {
	// Kafka client ID.
	ClientID string

	// LeaderRetryLimit limits the number of connection attempts to a single
	// node before failing. Use LeaderRetryWait to control the wait time
	// between retries.
	// Defaults to 10.
	LeaderRetryLimit int

	// LeaderRetryWait sets a limit to the waiting time when trying to connect
	// to a single node after failure.
	// Defaults to 500ms.
	// Timeout on a connection is controlled by the DialTimeout setting.
	LeaderRetryWait time.Duration

	// AllowTopicCreation enables a last-ditch "send produce request" which
	// happens if we do not know about a topic. This enables topic creation
	// if your Kafka cluster is configured to allow it.
	// Defaults to False.
	AllowTopicCreation bool

	// Any new connection dial timeout.
	// Default is 10 seconds.
	DialTimeout time.Duration

	// DialRetryLimit limits the number of connection attempts to every node in
	// cluster before failing. Use DialRetryWait to control the wait time
	// between retries.
	// Defaults to 10.
	DialRetryLimit int

	// DialRetryWait sets a limit to the waiting time when trying to establish
	// broker connection to single node to fetch cluster metadata.
	// Defaults to 500ms.
	DialRetryWait time.Duration
...
}
```
In this sink, I set the count of all retries to 1, and waiting time to zero, means that if the client can not connect the broker, it will  return immediately
+ We'll call processes that publish messages to a Kafka topic producers.So we need instantiate a producer base on the broker. And use the producer to send message to kafka server.

**Elasticsearch**
About retention policies, ES support gateway( gateway module stores the cluster state and shard data), data persistence. Now , we use the gateway to save the data with local file system.

rvrignaud:
@huangyuqi I really would be interested to see a PR for an ES sink.  Is this for events only or also metrics ? This would fix https://github.com/kubernetes/heapster/issues/227 ?

huangyuqi:
Hi, @rvrignaud , Elasticsearch is a distributed full text search engine, ES sink is able to save metrics and events; And we can use the RestAPI to search.

vishh:
```
$ ./heapster --logtostderr --source=cadvisor:external?standalone=true --sink="kafka:?brokers=localhost:2181&timeseriestopic=test&eventstopic=test"
I1103 10:00:04.764864    2337 heapster.go:55] ./heapster --logtostderr --source=cadvisor:external?standalone=true --sink=kafka:?brokers=localhost:2181&timeseriestopic=test&eventstopic=test
I1103 10:00:04.764964    2337 heapster.go:56] Heapster version 0.18.0
I1103 10:00:04.765003    2337 external.go:88] Running in standalone mode, external nodes source will only use localhost
F1103 10:00:04.766070    2337 heapster.go:62] failed to connect to kafka cluster: cannot connect
```

This is the error that I was suggesting to avoid. I will post a patch against this PR to fix this.

vishh:
@huangyuqi: I opened #683 based on this PR. I tested that against a local kafka cluster and it seems to be resilient against cluster outages. If you are OK with that, we can merge that in.

We need a simple integration test for this client.

vishh:
I'm closing this PR.

huangyuqi:
@vishh , I think something wrong about the sink's format:   ^_^

**your format :** `--sink=kafka:?brokers=localhost:2181&timeseriestopic=test&eventstopic=test`
**mine :**  `--sink=kafka:"http://?brokers=localhost:9092&timeseriestopic=test&eventstopic=test"`

two issues, let me explain : )
+ symbol `"&"`

`"&"` is used for support multi sinks, like this:
```
--sink=influxdb:http://0.0.0.0:80&gcm:http://0.0.0.0:81
```
so, the kafka sink's url needs quotation mark to avoid confuse. Maybe, we can modify the use of symbol `"&"` .

+ url format

Kafka sink gets the brokers list from the url's query string.So kafka sink's url need follow the format of URL.

So, i use `--sink=kafka:"http://?brokers=localhost:9092&timeseriestopic=test&eventstopic=test"`, the sink works normally.  Do you have a better suggestion ?

vishh:
Enclosing the entire sink config inside `"` should do the trick. --sink="kafka:?brokers=localhost:9092" should work even if you have multiple parameters. Can you try that?

huangyuqi:
` ./heapster --logtostderr --source=cadvisor:external?standalone=true --sink="kafka:?brokers=localhost:9092&timeseriestopic=test&eventstopic=test"`
It can work normally.Enclosing the entire sink config inside  `""` :+1:  ^_^

rvrignaud:
@huangyuqi Hi. Would you mind open a PR or at least send a gist for the Elasticsearch sink as you mentioned above ? It would be very useful to us to get that feature ! The issue is describe here : https://github.com/kubernetes/heapster/issues/227.

Thanks !

huangyuqi:
@rvrignaud , i am very sorry for delay.  I am working on this, I will open a PR in recent two days. I have finished most of the sink code. ^_^

huangyuqi:
Hi, @rvrignaud , so sorry for delay. I have finished the ES sink, would you please help me to review it? 
Or, if you have any suggestion, please tell me. ^_^
https://github.com/kubernetes/heapster/pull/733

BTW, Hi, @vishh , I found that I can not test any sink in standalone mode.

rvrignaud:
@huangyuqi : Hi. Thanks a lot for the PR. Unfortunately I'm not a Golang developer but I'll try to find some time to build and test it.

vishh:
@huangyuqi: Why are you not able to test sink in standalone mode?

PR #656, opened 2015-10-16
correct the node level metrics api

bluebreezecf:
When I tried the node level metrics api of heapster according the api descriptions in current model.md, I found it was wrong. After correcting the wrong node level metrics api according to the modification in this pr, the heapster then can reply with right response.


afein:
LGTM, thanks for catching this

mvdan:
ok to test

mvdan:
Oh, no jenkins run required. Merging.

vishh:
Thanks for the cleanup @bluebreezecf! Much appreciated!

PR #660, opened 2015-10-19
Fix OpenShift case

danmcp:
Openshift -> OpenShift


vishh:
ok to test

vishh:
LGTM. Thanks for the PR @danmcp!

PR #667, opened 2015-10-23
Updated Controller Spec link to yaml version

efranford:
Just noticed this broken link while I was setting this up on my box.


mvdan:
My bad, I forgot to update the markdown docs when changing the formats. Thanks for noticing!

PR #670, opened 2015-10-23
Do not crash if one of the sinks fails to come up.

vishh:
Related to https://github.com/kubernetes/kubernetes/pull/16214

cc @piosz

### DIFF
--- a/heapster.go
+++ b/heapster.go
@@ -118,7 +118,8 @@ func doWork() ([]source_api.Source, sinks.ExternalSinkManager, manager.Manager, 
 	c := cache.NewCache(*argCacheDuration, time.Minute)
 	sources, err := newSources(c)
 	if err != nil {
-		return nil, nil, nil, err
+		// Do not fail heapster is source setup fails for any reason.
+		glog.Errorf("failed to setup sinks - %v", err)
 	}
 	sinkManager, err := sinks.NewExternalSinkManager(nil, c, *argSinkFrequency)
 	if err != nil {
@@ -138,7 +139,8 @@ func doWork() ([]source_api.Source, sinks.ExternalSinkManager, manager.Manager, 
 		return nil, nil, nil, err
 	}
 	if err := manager.SetSinkUris(argSinks); err != nil {
-		return nil, nil, nil, err
+		// Do not fail heapster is sink setup fails for any reason.
+		glog.Errorf("failed to setup sinks - %v", err)
 	}
 
 	manager.Start()
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -252,7 +252,9 @@ func onlyKeepLatestStat(cont *cache.ContainerElement) {
 func (rm *realManager) SetSinkUris(sinkUris Uris) error {
 	sinks, err := newSinks(sinkUris, rm.resolution)
 	if err != nil {
-		return err
+		// Skip sink setup errors here. We do not want to fail if one or more
+		// sinks fail.
+		glog.Error(err)
 	}
 	if err := rm.sinkManager.SetSinks(sinks); err != nil {
 		return err
--- a/manager/sinks.go
+++ b/manager/sinks.go
@@ -16,6 +16,7 @@ package manager
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"k8s.io/heapster/extpoints"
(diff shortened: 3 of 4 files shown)


vishh:
http://104.197.39.211/job/heapster-e2e-gce/2909/console

mvdan:
LGTM if this behaviour is intended

vishh:
In general I'm thinking logging and forging ahead is better than
crashlooping. Imagine a user adding a sink dynamically - we do not want
heapster to crash in such a scenario right?

mvdan:
Note that this just returns an error, it doesn't exit the program. It's up to the caller to decide whether to treat the error as fatal.

`heapster.go` does error and exit, since you want that if you cannot bring up the sinks specified via the command line.

OTOH, `api/v1/api.go` will just return a `StatusInternalServerError` to the HTTP client and not exit.

I think both cases are correct, so I don't see the point of this PR at the moment. It would make `--sinks` unintuitive and hide errors from the API - if you tell it to add a sink and it fails, it would not return the error response.

vishh:
What if one of the sinks were to crash if its endpoint isn't available?
That is not acceptable in a dynamic cluster environment.
From that perspective, even if I were to configure a sink via flags, it
must continue to retry ideally, and not cause heapster to crash.
Also sinks are just one functionality of heapster. The http APIs are
expected to function even if the sinks fail. May be we can raise events, if
running in a kube environment, in case of sink failures.

mvdan:
If you want sinks to not cause an exit when used via flags, then change `heapster.go`. I still think that modifying it this way is not the right way to do it. If you do an HTTP POST changing the sinks and you get a 200 response when setting the new sinks failed, that is definitely broken.

vishh:
Ah I get what you are saying now.. Will fix the PR!

piosz:
So what is the current plan to fix this?

jimmidyson:
IMO I feel there are a couple of different scenarios to cater for in different ways:

1. Sink unavailable on startup - fail startup. More likely to highlight misconfiguration.
2. Sink unavailable when adding at runtime. Don't add sink, return appropriate error with non-200 status code (not sure quite what to use).
3. Sink becomes unavailable after being added. Sink should buffer until sink is available again.

mvdan:
> Sink unavailable on startup - fail startup. More likely to highlight misconfiguration.

This was already in place, and I did not change it when adding sinks at runtime.

> Sink unavailable when adding at runtime. Don't add sink, return appropriate error with non-200 status code (not sure quite what to use).

I made it work this way when I added sinks at runtime. I made it return InternalServerError (500), suggestions on a better status code welcome. But we definitely want 4xx/5xx.

> Sink becomes unavailable after being added. Sink should buffer until sink is available again.

AFAIK nothing is done in this regard. Nothing changed about this when I added sinks at runtime, so it didn't even occur to me.

jimmidyson:
A 422, `Unprocessable Entity`, is probably the most appropriate error code.

@mvdan Do you think the current behaviour is wrong? What do you think should be correct behaviour?

mvdan:
@jimmidyson good suggestion. 1.5 is limited as far as http status codes, since they stick to the widely approved ones. So the 400s only go up to 418 on 1.5, for example. Maybe leave it as a TODO? I recently saw a commit in golang master that added many more status codes, so 422 will probably land in 1.6. Although starting to use it would mean dropping 1.5, so probably not going to happen soon.

I think the current behaviour on cases 1 and 2 is correct, hence me not understanding this PR. You raised a valid point on case 3, but that is a different issue - this issue is about "coming up".

jimmidyson:
We can add in 422 as kubernetes does [here](https://github.com/kubernetes/kubernetes/blob/75c977d200f29418fefb5519e5cf3388b4425d3a/pkg/api/errors/errors.go#L31-L41).

So if current behaviour is correct what is this PR for @vishh?

vishh:
Crashlooping is not desirable. It adds unnecessary overhead to the orchestration system - dead containers lying around, signals firing in all directions, etc.

Heapster now supports multiple sources and sinks. It is possible for some of the sources or sinks to not be available when heapster starts. In addition to that the sources and sinks lifecycles are decoupled from that of heapster. I think a good cluster application should be resilient to the dynamic nature of cluster workloads.

This PR deals with the configuration issues specifically. A user can look at the logs and then debug any issues that they encounter with heapster.

I have been updating a few sinks to be resilient to backend outages as well. As part of this I also added a utility library that can help manage backend availability for sinks.

vishh:
@mwielgus: Can you review this PR? This will help prevent heapster from crashlooping if GCM APIs or scopes are not enabled.

mwielgus:
LGTM

PR #671, opened 2015-10-23
Ignore pods that have terminated.

vishh:
Addresses https://github.com/kubernetes/kubernetes/issues/16168


vishh:
cc @dchen1107

vishh:
cc @piosz @mwielgus

piosz:
LGTM

dchen1107:
Thanks for fix.

PR #676, opened 2015-10-27
remove the type of constant variables

huangyuqi:
remove the type of constant string variables.


vishh:
ok to test

vishh:
Thanks for the cleanup @huangyuqi :+1:

PR #678, opened 2015-10-27
correct the comments of interface Cache

bluebreezecf:
I try to dig into the sources of heapster, and find some comments of interface Cache (in heapster/sinks/cache/cache.go line 91 and 97) may just copy from line 85, and if just
see the comments and method definition, user may get confused.

After reading the implementation in heapster/sinks/cache/impl.go, i modify the method comments 
of Cache, aiming at making the methods comments more readable and meaningful.


vishh:
ok to test

vishh:
LGTM. Thanks for the PR @bluebreezecf :+1:

PR #680, opened 2015-10-30
Add and use deletePodContainer and corresponding test for realModel

bluebreezecf:
Currently there is a runGC for the sinks/cache.Cache (actually sinks/cache.realCache)  every other gcDuration (the duration of *argCacheDuration) , which may triggher listener.PodContainerEvicted
(https://github.com/kubernetes/heapster/blob/master/sinks/cache/impl.go#L98). 

``` go
func (rc *realCache) runGC() {
    rc.Lock()
    defer rc.Unlock()
    for podName, podElem := range rc.pods {
        for contName, contElem := range podElem.containers {
            if rc.isTooOld(contElem.lastUpdated) {
                delete(podElem.containers, contName)
                for _, listener := range rc.cacheListeners {
                    if listener.PodContainerEvicted != nil {
                        listener.PodContainerEvicted(podElem.Namespace, podElem.Name, contName)
```

The actual used listener,which is created by model.GetCacheListener() , just does nothing (https://github.com/kubernetes/heapster/blob/master/model/impl.go#L387). 

``` go
func (rc *realModel) GetCacheListener() cache.CacheListener {
    return cache.CacheListener{
        NodeEvicted:      func(hostName string) { rc.deleteNode(hostName) },
        NamespaceEvicted: func(namespace string) { rc.deleteNamespace(namespace) },
        PodEvicted:       func(namespace string, podName string) { rc.deletePod(namespace, podName) },
        PodContainerEvicted: func(namespace string, podName string, containerName string) {
            // do nothing at this moment.
        },
```

But the aggregationStep of realModel, which is triggered in the realModel.Update()  for every other modelDuration, will finally aggregate the container metrics for the given pod
(https://github.com/kubernetes/heapster/blob/master/model/aggregation.go#L164)

``` go
    // Collect the Container InfoTypes
    sources := []*InfoType{}
    for _, container := range pod.Containers {
        sources = append(sources, &(container.InfoType))
    }
    c <- rc.aggregateMetrics(&pod.InfoType, sources, latestTime)
```

Meanwhile, the rest apis, which may use the ageing pod container metrics to aggregate the pod metrics in the response (https://github.com/kubernetes/heapster/blob/master/model/getters.go#L135), 

``` go
func (rc *realModel) GetPodMetric(req PodMetricRequest) ([]statstore.TimePoint, time.Time, error) {
    var zeroTime time.Time
    rc.lock.RLock()
    defer rc.lock.RUnlock()

    if len(rc.Namespaces) == 0 {
        return nil, zeroTime, errModelEmpty
    }
    ns, ok := rc.Namespaces[req.NamespaceName]
    if !ok {
        return nil, zeroTime, errNoSuchNamespace
    }
    pod, ok := ns.Pods[req.PodName]
    if !ok {
        return nil, zeroTime, errNoSuchPod
    }
    if len(pod.Metrics) == 0 {
        return nil, zeroTime, errNoEntityMetrics
    }
    ts, ok := pod.Metrics[req.MetricName]
    if !ok {
        return nil, zeroTime, errNoSuchMetric
    }

    res := (*ts).Hour.Get(req.Start, req.End)
    return res, rc.timestamp, nil
}
```

will be called by the outside to get k8s metrics via heapster.

Now, we can figure out that the aggregated pod metrics may not be so accurate. Besides, the aging pod container data in realModel will not be removed until the belonging pod is ageing and removed
in the PodEvicted(), which is also only triggered by the runGC().

So, i add deletePodContainer() for realModel, trigger to call it in PodContainerEvicted in model/impl.go, and add the corresponding test for deletePodContainer in model/impl_test.go. The details can be shown in the commit details of this PR.

### DIFF
--- a/model/impl.go
+++ b/model/impl.go
@@ -384,7 +384,7 @@ func (rc *realModel) GetCacheListener() cache.CacheListener {
 		NamespaceEvicted: func(namespace string) { rc.deleteNamespace(namespace) },
 		PodEvicted:       func(namespace string, podName string) { rc.deletePod(namespace, podName) },
 		PodContainerEvicted: func(namespace string, podName string, containerName string) {
-			// do nothing at this moment.
+			rc.deletePodContainer(namespace, podName, containerName)
 		},
 		FreeContainerEvicted: func(hostName string, containerName string) { rc.deleteFreeContainer(hostName, containerName) },
 	}
@@ -514,5 +514,20 @@ func (rc *realModel) deleteFreeContainer(hostname, name string) {
 	// Get Node pointer
 	if node, ok := rc.Nodes[hostname]; ok {
 		delete(node.FreeContainers, name)
+	}
+}
+
+// deletePodContainer deletes a container from the belonging pod in the given namespace.
+// deletePodContainer receives a name of the target namespace, a name of the belonging pod,
+// and a name of the container to be deleted.
+func (rc *realModel) deletePodContainer(namespace, podName, containerName string) {
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+
+	// Delete pod container from the belonging pod in the target namespace
+	if namespaceInfo, ok := rc.Namespaces[namespace]; ok {
+		if podInfo, ok := namespaceInfo.Pods[podName]; ok {
+			delete(podInfo.Containers, containerName)
+		}
 	}
 }
--- a/model/impl_test.go
+++ b/model/impl_test.go
@@ -1057,3 +1057,43 @@ func TestDeleteFreeContainer(t *testing.T) {
 	cluster.deleteFreeContainer(hostName, freeContainerName)
 	// No panic etc.
 }
+
+// TestDeletePodContainer tests all flows of deletePodContainer.
+func TestDeletePodContainer(t *testing.T) {
+	var (
+		cluster = newRealModel(time.Minute)
+		podElem = podElementFactory()
+		assert  = assert.New(t)
+	)
+
+	// Initialize the cluster model with a pod with fake pod container
+	timestamp, err := cluster.updatePod(podElem)
+	assert.NoError(err)
+	assert.NotEqual(timestamp, time.Time{})
+	assert.NotNil(cluster.Nodes[podElem.Hostname])
+	podInfo := cluster.Nodes[podElem.Hostname].Pods[getPodKey(podElem.Namespace, podElem.Name)]
+	assert.NotNil(cluster.Namespaces[podElem.Namespace])
+	assert.Equal(podInfo, cluster.Namespaces[podElem.Namespace].Pods[podElem.Name])
+	assert.Equal(podInfo.Labels, podElem.Labels)
+
+	// Record the number of pod containers after initialization
+	containersNumAfterAdded := len(podInfo.Containers)
+	assert.NotEqual(0, containersNumAfterAdded)
+
+	// Record the names to be used to invoke deletePodContainer
+	namespace := podInfo.Namespace
+	podName := podInfo.Name
+	var containerName string
+	for name, _ := range podInfo.Containers {
+		containerName = name
+		break
+	}
+
+	// First call : try to delete newly added pod container
+	cluster.deletePodContainer(namespace, podName, containerName)
+	assert.Equal(containersNumAfterAdded, len(podInfo.Containers)+1)
+
+	// Second call: already deleted
+	cluster.deletePodContainer(namespace, podName, containerName)
+	// No panic etc.
+}


vishh:
> The actual used listener,which is created by model.GetCacheListener() , just does nothing (https://github.com/kubernetes/heapster/blob/master/model/impl.go#L387).

It is deleting objects right?

> Now, we can figure out that the aggregated pod metrics may not be so accurate. Besides, the aging pod container data in realModel will not be removed until the belonging pod is ageing and removed
in the PodEvicted(), which is also only triggered by the runGC().

By ageing, I assume you are referring to terminated pods. Is that correct?

> So, i add deletePodContainer() for realModel, trigger to call it in PodContainerEvicted in model/impl.go, and add the corresponding test for deletePodContainer in model/impl_test.go. The details can be shown in the commit details of this PR.

AFAIK, k8s does not allow adding/removing containers in pods. That was probably the reason why the existing logic did not handle pod container deletions. 
In general though, if we are abstracting out k8s concepts in the model, the proposed change makes sense.

I will let @mwielgus and @piosz also take a look before this change is merged.

bluebreezecf:
As far as I know, heapster has two level cache of k8s metrics in some extent. The first level is cache.realCache  (in sinks/cache/impl.go), which is holding the original and raw metrics collected from k8s. It has an ageing and clean-up mechanism triggered by listener.PodContainerEvicted
(https://github.com/kubernetes/heapster/blob/master/sinks/cache/impl.go#L98). The runGC() method
cleans the metrics data of pod containers/pods/free containers/node for ageing in every other gcDuration (the duration of *argCacheDuration)
```go
func (rc *realCache) runGC() {
	rc.Lock()
	defer rc.Unlock()
	for podName, podElem := range rc.pods {
		for contName, contElem := range podElem.containers {
			if rc.isTooOld(contElem.lastUpdated) {
				delete(podElem.containers, contName)
				for _, listener := range rc.cacheListeners {
					if listener.PodContainerEvicted != nil {
						listener.PodContainerEvicted(podElem.Namespace, podElem.Name, contName)
					}
				}
			}
		}
		if rc.isTooOld(podElem.lastUpdated) {
			delete(rc.pods, podName)
			for _, listener := range rc.cacheListeners {
				if listener.PodEvicted != nil {
					listener.PodEvicted(podElem.Namespace, podElem.Name)
				}
			}
		}
	}

	for nodeName, nodeElem := range rc.nodes {
		for contName, contElem := range nodeElem.freeContainers {
			if rc.isTooOld(contElem.lastUpdated) {
				delete(nodeElem.freeContainers, contName)
				for _, listener := range rc.cacheListeners {
					if listener.FreeContainerEvicted != nil {
						listener.FreeContainerEvicted(nodeName, contName)
					}
				}
			}
		}

		if rc.isTooOld(nodeElem.node.lastUpdated) {
			delete(rc.nodes, nodeName)
			for _, listener := range rc.cacheListeners {
				if listener.NodeEvicted != nil {
					listener.NodeEvicted(nodeName)
				}
			}
		}
	}
}

func NewCache(bufferDuration, gcDuration time.Duration) Cache {
	rc := &realCache{
		events:         store.NewGCStore(store.NewTimeStore(), bufferDuration),
		eventUIDs:      make(map[string]struct{}),
		bufferDuration: bufferDuration,
	}
	go util.Until(rc.runGC, gcDuration, util.NeverStop)
	return rc
}
```

The second level cache of heapster is model.realModel (in model/impl.go), which stores the recomputed and aggregated metrics used by the api server of heapster to reply the outer-side
requests.

These two level caches are synchronized data by two ways:
- Add data periodically from cache.realCache to model.realModel by realModel.Update(), which is 
triggered by realManager.HousekeepModel()
```go
func (rm *realManager) HousekeepModel() {
	for {
		select {
		//TODO: better timing here
		case <-time.After(rm.modelDuration):
			if err := rm.model.Update(rm.cache); err != nil {
				glog.V(1).Infof("Model housekeeping returned error: %s", err.Error())
			}
		case <-rm.modelStopChan:
			return
		}
	}
}
```
- Callback the methods of listener to remove data from model.realModel, after runGC removes (ages)
the original metrics.The listener is initialized by realModel.GetCacheListener() in manager.NewManager (https://github.com/kubernetes/heapster/blob/master/manager/manager.go#L80). So it means heapster firstly removes the ageing of original metrics in cache.realCache, and secondly removes the corresponding aggregated metrics in model.realModel
```go
func NewManager(sources []source_api.Source, sinkManager sinks.ExternalSinkManager, res, bufferDuration time.Duration,
	c cache.Cache, useModel bool, modelRes, modelDuration time.Duration) (Manager, error) {

	var newModel model.Model
	if useModel {
		newModel = model.NewModel(modelRes)
		// Temporary semi-hack to get model storage garbage-collected.
		c.AddCacheListener(newModel.GetCacheListener())
	}
```

>>The actual used listener,which is created by model.GetCacheListener() , just does nothing 
>>(https://github.com/kubernetes/heapster/blob/master/model/impl.go#L387).

>It is deleting objects right?

So i think the added removal of  aggregated metrics data in model.realModel is just aiming at be 
accordance with the corresponding original metrics in cache.realCache, and it may make the aggregated metrics more accurate. 


>>Now, we can figure out that the aggregated pod metrics may not be so accurate. Besides, the 
>>aging pod container data in realModel will not be removed until the belonging pod is ageing and removed in the PodEvicted(), which is also only triggered by the runGC().

>By ageing, I assume you are referring to terminated pods. Is that correct?

I don't mean ageing is terminated pods, but just mean current realization of old data removal mechanism, which is triggered by cache.runGC(). Only the original metrics of pods containers/
pods/nodes/free containers are out of the date and will be removed from the cache.realCache beyond cache duration since current detecting (https://github.com/kubernetes/heapster/blob/master/sinks/cache/impl.go#L76).
```go
func (rc *realCache) isTooOld(lastUpdated time.Time) bool {
	if time.Now().Sub(lastUpdated) >= rc.bufferDuration {
		return true
	}
	return false
}
```

afein:
Regarding metric aging, the model uses the self-managing DayStore structure which does not require garbage collection for old metrics.

Also, as @vishh said, it is not possible to add/remove containers from Pods at runtime, which is probably why this functionality was originally left out.

If you are observing inaccuracies in aggregated Pod metrics, the root cause might be different than the garbage collection system. Feel free to elaborate more on your use case and your observations and I'll try to help pinpoint the problem

bluebreezecf:
@afein: Thanks for your kindly explanation.

AFAIK, the self-managing ability of DayStore used in model.realModel just takes effects in the scene of adding newly metrics, just like the reapOldData operation of gcStore, which is used in cache.realCache
to holds the original metrics. If currently something wrong has just happened to the related k8s cluster, or just in a certain duration, there may be no newly original metrics to be queries by heapster from k8s, and there may also be no newly aggregated metrics to store into model.realModel, this self-managing ability will not work in this scene.

The runGC runs periodically and never to stop to clear ageing original metrics in cache.realCache, as well as triggers the removal of ageing aggregated metrics stored in model.realModel. 

These two kinds of metrics ageing mechanism run with cooperation, and the runGC one of aging operations can be a good insurance of triggering removal of old (original/aggregrated) metrics.
Besides the inaccuracies of aggregated pod metrics, making runGC take care of the old metrics in
model.realModel is my PR's  consideration.

afein:
Thanks for the explanation of your concerns @bluebreezecf,
AFAIK, when a Pod is terminated, the existing deletePod method should remove all references to the corresponding PodInfo struct, which should eventually trigger Go's garbage collection for the referenced DayStore altogether.

Because of that, I suspect that the underlying cause of your observations may not be caused by the existing garbage collection mechanisms. 

May I ask the following:
- How did you observe that old metrics are not removed for pods that have been deleted?
- What types of inaccuracies did you observe in aggregated pod metrics?

bluebreezecf:
@afein:

> when a Pod is terminated, the existing deletePod method should remove all references to the corresponding PodInfo struct

Actually the pod original metrics in cache.realCache are deleted by runGC, just because the difference of now time and podElement's laste updated time is larger than realCache.bufferDuration.
The real pod entity in k8s is removed will do nothing to the corresponding data structure in heapster cache immediately; while the metrics are just ageing by realCache.runGC (ageing for original metrics in cache.realCache directly, and for aggregated metrics in model.realModel indirectly), gcStore.reapOldData() (for original metrics in cache.realCache) or StatStore.rewind() (ageing for aggregated metrics in model.realModel)

> - How did you observe that old metrics are not removed for pods that have been deleted?
> - What types of inaccuracies did you observe in aggregated pod metrics?

Actually I do not observe the inaccuracies or un-removal Influences, I just dig into the sources
codes of heapster and suppose there may exist such inaccuracies of aggregated metrics. 
I am curious if we can forcast it may happen something bad, maybe we can do some discussions
or precautions against it to make things better. ：）

Thank so much for your replies and discussion @afein @vishh

piosz:
I don't mind with this PR, but let @mwielgus to also take w look.

bluebreezecf:
@mwielgus 
Could you have spare time to check this PR and the above discussions ?

piosz:
@bluebreezecf we will take a look today together with @mwielgus

bluebreezecf:
Thanks a lot, bro! @piosz

piosz:
LGTM

bluebreezecf:
@vishh @piosz @afein @mwielgus:  Appreciate for your kindness and rigorous reviews, discussions and supports! Thanks Bros!

PR #683, opened 2015-11-03
kafka sink

vishh:
Based on #648.

Added client retry logic.


vishh:
cc @mwielgus

vishh:
I'm merging this assuming that there are no objects from @piosz and @mwielgus.

piosz:
Thanks @vishh!

PR #685, opened 2015-11-04
Initial skeleton of new data flow

piosz:
@mwielgus

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -15,131 +15,46 @@
 package manager
 
 import (
-	"fmt"
-	"sync"
 	"time"
 
-	"github.com/golang/glog"
-
-	"k8s.io/heapster/model"
+	"k8s.io/heapster/processor"
 	"k8s.io/heapster/sinks"
-	sink_api "k8s.io/heapster/sinks/api"
-	"k8s.io/heapster/sinks/cache"
-	source_api "k8s.io/heapster/sources/api"
+	"k8s.io/heapster/sources"
 )
 
 // Manager provides an interface to control the core of heapster.
 // Implementations are not required to be thread safe.
 type Manager interface {
-	// Export the latest data point of all metrics.
-	ExportMetrics() ([]*sink_api.Point, error)
-
-	// Set the sinks to use
-	SetSinkUris(Uris) error
-
-	// Get the sinks currently in use
-	SinkUris() Uris
-
-	// Get a reference to the cluster entity of the model, if it exists.
-	GetModel() model.Model
-
-	// Starts the manager.
 	Start()
-
-	// Stops the manager.
 	Stop()
 }
 
 type realManager struct {
-	sources       []source_api.Source
-	cache         cache.Cache
-	model         model.Model
-	modelDuration time.Duration
-	sinkManager   sinks.ExternalSinkManager
-	sinkUris      Uris
-	lastSync      time.Time
-	resolution    time.Duration
-	decoder       sink_api.Decoder
-	mainStopChan  chan struct{}
-	sinkStopChan  chan<- struct{}
-	modelStopChan chan struct{}
+	sources    sources.SourceManager
+	processors []processor.DataProcessor
+	sinks      sinks.SinkManager
+	lastSync   time.Time
+	resolution time.Duration
+	stopChan   chan struct{}
 }
 
-type syncData struct {
-	data  source_api.AggregateData
-	mutex sync.Mutex
-}
-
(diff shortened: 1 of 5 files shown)


aggregator/aggregator.go
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aggregator
+
+import (
+	. "k8s.io/heapster/types"
+)
+
+type Aggregator interface {
```
mwielgus: It should not be called aggregator. Lets call it DataBatchProcessor and we will have aggregators implementing this interface.

types/types.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"time"
+)
+
+type ValueType int
+
+const (
+	// An int64 value.
+	ValueInt64 ValueType = iota
```
mwielgus: it shouldrather be Int64Gauge, Int64Cumulative etc so that we know what is the interpretation of the value.

manager/manager.go
```diff
@@ -15,131 +15,46 @@
...
-	c cache.Cache, useModel bool, modelRes, modelDuration time.Duration) (Manager, error) {
-
-	var newModel model.Model
-	if useModel {
-		newModel = model.NewModel(modelRes)
-		// Temporary semi-hack to get model storage garbage-collected.
-		c.AddCacheListener(newModel.GetCacheListener())
-	}
-
+func NewManager(sources sources.SourceManager, aggregator aggregator.Aggregator, sinks sinks.SinkManager, res time.Duration) (Manager, error) {
```
mwielgus: There should be a list of DataBatchProcessors, that are fired one after another.

sources/manager.go
```diff
@@ -0,0 +1,35 @@
...
+	ScrapeSources(start, end time.Time) *DataBatch
+}
+
+func NewSourceManager() (SourceManager, error) {
+	return &realSourceManager{}, nil
+}
+
+type realSourceManager struct {}
+
+func (m *realSourceManager) ScrapeSources(start, end time.Time) *DataBatch {
```
mwielgus: it should return a list of data batches. Otherwise secifying start doesn't make any sense.

types/types.go
```diff
@@ -0,0 +1,44 @@
...
+	// A double-precision floating point number.
+	ValueDouble
+)
+
+type MetricValue struct {
+	Name string
```
mwielgus: Lets move name outside of the struct and keep it only as map key.

types/types.go
```diff
@@ -0,0 +1,44 @@
...
+	Type ValueType
+	Value interface{}
+}
+
+type MetricSet struct {
+	MetricValue []MetricValue
```
mwielgus: Repalce with metrics map[MetricValue]string

aggregator/aggregator.go
piosz: Done

manager/manager.go
piosz: Done

sources/manager.go
piosz: Discussed offline. Not changing

types/types.go
piosz: Done

types/types.go
piosz: Done

types/types.go
piosz: Done

piosz:
PTAL

processor/processor.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	. "k8s.io/heapster/types"
+)
+
+type DataProcessor interface {
+	Process(*DataBatch) *DataBatch
+}
+
+func NewDataProcessor() (DataProcessor, error) {
```
mwielgus: NewDummyDataProcessor

processor/processor.go
```diff
@@ -0,0 +1,33 @@
...
+	return &realDataProcessor{}, nil
+}
+
+type realDataProcessor struct{}
```
mwielgus: dummyDataProcessor

processor/processor.go
piosz: Done

processor/processor.go
piosz: Done

mwielgus:
LGTM, thanks for writing our discussion down. Most likely we will change/move the core types around.
I will take care of fixing the makefile.

PR #689, opened 2015-11-05
SourceManager implementation

mwielgus:


core/types.go
```diff
@@ -45,5 +45,5 @@ type MetricSet struct {
 
 type DataBatch struct {
 	Timestamp time.Time
-	MetricSet map[string]MetricSet
+	MetricSets map[string]MetricSet
```
piosz: How about changing also MetricValue to plural above?

sources/manager.go
```diff
@@ -15,21 +15,126 @@
...
 
+	"github.com/golang/glog"
 	. "k8s.io/heapster/core"
 )
 
+const (
+	MetricsScrapeTimeout = 20 * time.Second
+)
+
+//
```
piosz: I'm not sure if this is the right way to make comments;)

sources/manager.go
```diff
@@ -15,21 +15,126 @@
...
+	return fmt.Sprintf("kubelet:%s:%d", this.host, this.port)
+}
+
+func (this *KubeletMetricsSource) ScrapeMetrics(start, end time.Time) *DataBatch {
+	var tmp DataBatch
+	return &tmp
+}
+
+//
+// Provides a list of sources to be scaped.
```
piosz: This comments is rather related to GetMetricsSources() method.

sources/manager.go
```diff
@@ -15,21 +15,126 @@
...
+
+type KubeletProvider struct {
+}
+
+func (this *KubeletProvider) GetMetricsSources() []MetricsSource {
+	return []MetricsSource{}
+}
+
+//
+// Scrapes sources in parallel and combines the responses.
```
piosz: Same here

sources/manager.go
```diff
@@ -15,21 +15,126 @@
...
+		if !now.Before(timeout) {
+			glog.Warningf("Failed to get all responses in time (got %d/%d)", i, len(sources))
+			break
+		}
+
+		select {
+		case dataBatch := <-responseChannel:
+			if dataBatch != nil {
+				for key, value := range dataBatch.MetricSets {
+					response.MetricSets[key] = value
```
piosz: There will be nil pointer here. Create MetricSets field in response.

mwielgus: Done

core/types.go
mwielgus: Done

sources/manager.go
mwielgus: Done.

sources/manager.go
mwielgus: Done

sources/manager.go
mwielgus: Done.

piosz:
LGTM

sources/manager.go
```diff
@@ -15,21 +15,115 @@
...
-type realSourceManager struct{}
+func (this *realSourceManager) ScrapeSources(start, end time.Time) *DataBatch {
+	sources := this.metricsSourceProvider.GetMetricsSources()
+
+	responseChannel := make(chan *DataBatch)
+	timeout := time.Now().Add(MetricsScrapeTimeout)
+
+	for _, source := range sources {
+		go func(channel chan *DataBatch, start, end, timeout time.Time) {
+			glog.Infof("Querying source: %s", source)
```
vishh: nit: I don't think we want to spam log files with info logs.

vishh:
I don't know the overall plan behind these PRs. FYI: Certain metrics can have multiple values per container - for example, there can be multiple disks per container and we currently use resource identifiers to distinguish those disks. This can be applicable to per-cpu usage, network, etc.

PR #698, opened 2015-11-09
Better time/parallelism management in manager

mwielgus:


mwielgus:
Tests for the whole flow will come in the next PR.

manager/manager.go
```diff
@@ -59,10 +74,12 @@ func (rm *realManager) Stop() {
 
 func (rm *realManager) Housekeep() {
 	for {
+		// Always try to get the newest metrics
+		now := time.Now()
 		start := rm.lastSync
```
piosz: Let's have a resolution size of a window. Remove last sync

manager/manager.go
```diff
@@ -75,9 +92,27 @@ func (rm *realManager) Housekeep() {
 }
 
 func (rm *realManager) housekeep(start, end time.Time) {
-	data := rm.sources.ScrapeSources(start, end)
-	for _, p := range rm.processors {
-		data = p.Process(data)
+	if !start.Before(end) {
```
piosz: if end - start != resolution

manager/manager.go
```diff
@@ -75,9 +92,27 @@ func (rm *realManager) Housekeep() {
...
+		glog.Warningf("Wrong time provided to housekeep start:%s end: %s", start, end)
+		return
+	}
+
+	select {
+	case <-time.After(rm.resolution / 2):
```
piosz: housekeep timeout

manager/manager.go
```diff
@@ -75,9 +92,27 @@ func (rm *realManager) Housekeep() {
...
+		glog.Warningf("Spent too long waiting for housekeeping to start")
+		return
+
+	case <-rm.housekeepSemaphoreChan:
```
piosz: Move this up

manager/manager.go
```diff
@@ -75,9 +92,27 @@ func (rm *realManager) Housekeep() {
...
+		// ok, good to go
 	}
-	rm.sink.ExportData(data)
+
+	go func(rm *realManager) {
+		// should always give back the semaphore
```
piosz: Move to realManager

manager/manager.go
mwielgus: Done.

manager/manager.go
mwielgus: This is already enforced in Housekeep. Simple sanity check is probably more than enough.

manager/manager.go
mwielgus: Done.

manager/manager.go
mwielgus: Done.

manager/manager.go
mwielgus: We would have two functions that take start/end and do housekeeping. Lets leave it as is.

piosz:
unit test doesn't compile. please fix

manager/manager.go
```diff
@@ -59,14 +74,15 @@ func (rm *realManager) Stop() {
...
 func (rm *realManager) Housekeep() {
 	for {
-		start := rm.lastSync
+		// Always try to get the newest metrics
+		now := time.Now()
+		start := now.Truncate(rm.resolution)
 		end := start.Add(rm.resolution)
-		timeToNextSync := end.Sub(time.Now())
-		// TODO: consider adding some delay here
+		timeToNextSync := end.Add(scrapeOffset).Sub(end)
```
piosz: It seems to be incorrect: time to next sync will be always 5 sec

manager/manager.go
piosz: OK

manager/manager.go
piosz: OK

manager/manager.go
mwielgus: Of course,  should be end.Add(scrapeOffset).Sub(now). Done.

piosz:
LGTM

PR #716, opened 2015-11-19
Implemented decoding metrics

piosz:


core/metrics.go
```diff
@@ -0,0 +1,318 @@
...
+import (
+	"time"
+
+	source_api "k8s.io/heapster/sources/api"
+)
+
+// Stub out for testing
+var timeSince = time.Since
+
+var SupportedMetrics = []SupportedMetric{
```
mwielgus: We will support all metrics, not only these mentioned below. Name it as StandardMetrics or so.

core/metrics.go
```diff
@@ -0,0 +1,318 @@
...
+package core
+
+import (
+	"time"
+
+	source_api "k8s.io/heapster/sources/api"
+)
+
+// Stub out for testing
+var timeSince = time.Since
```
mwielgus: Is it still used?

piosz: Done

core/metrics.go
piosz: Done

vishh:
Cadvisor exposes docker image labels. Make sure we preserve that.

piosz:
I added TODO to export all supported labels.

piosz:
Oh, I can see Image field. I missed it previously. Will add in follow-up PR.

PR #729, opened 2015-11-25
Namespace aggregator

mwielgus:


core/labels.go
```diff
@@ -36,10 +36,15 @@ var (
 		Key:         "pod_name",
 		Description: "The name of the pod",
 	}
+	// TODO: Replace with namespace_name
```
piosz: We now have 2 labels which means exactly the same. Please fix in this PR

mwielgus: Fixed. Although I'm not sure what consequences this change will have for GKE/GCM. We will have to put this in release notes.

piosz:
LGTM

core/labels.go
```diff
@@ -36,10 +36,15 @@ var (
 		Key:         "pod_name",
 		Description: "The name of the pod",
 	}
+	// Deprecated label
 	LabelPodNamespace = LabelDescriptor{
 		Key:         "pod_namespace",
 		Description: "The namespace of the pod",
 	}
+	LabelNamespaceName = LabelDescriptor{
```
vishh: Can we create a new version of storage schema instead of deprecating labels in the existing version?
It will be much easier for users to understand. WDYT?

PR #730, opened 2015-11-25
Batch size limit for Influxdb sink

mwielgus:


sinks/influxdb/influxdb.go
```diff
@@ -55,27 +55,49 @@ const (
 	valueField = "value"
 	// Event special tags
 	dbNotFoundError = "database not found"
+
+	// Maximum number of influxdb Points to be sent in one batch.
+	maxSendBatchSize = 10000
```
piosz: Any explanation why did you choose this number?

piosz:
LGTM

sinks/influxdb/influxdb.go
mwielgus: I had to pick some. 10000 points sounds more-or-less reasonable. We will fine-tune it soon and put it behind a flag.

PR #738, opened 2015-11-27
Stubs for model/metrics API

mwielgus:
Long-term stats will be moved to a different component. There will be no apiserver like data - node, namespace and pod lists should be taken from the apiserver.


piosz:
LGTM

Issue #731, opened 2015-11-25
Add new label for the Node name


mwringe:
We should add support for a new label for the name of the node. This should correspond to the metadata.name element of the node.

Currently the closest thing we have is "host_id" but this corresponds to using a deprecated 'ExternalID' (https://github.com/kubernetes/kubernetes/blob/release-1.0/pkg/api/v1/types.go#L1256) which is probably not something we should continue to use.

mwringe:
I already have a patch getting ready for this, the bigger issue is that heapster broke compatibility with OpenShift about a month ago (which I also have a patch almost ready to commit, I just need to update the tests). 

How does one get added as someone who can take tasks?

vishh:
@mwringe: I have added you to the `kubernetes` github org. Issues and PRs can be assigned to you now.

piosz:
We do have a label https://github.com/kubernetes/heapster/blob/master/sinks/api/supported_labels.go#L45 which is usually set to node name https://github.com/kubernetes/heapster/blob/master/sources/nodes/kube.go#L71

mwringe:
The hostname is only set to the node name if the hostname is empty. Which I don't think should be the case in most situations.

There is also the host_id which corresponds to the external id of the node. If the system doesn't overwrite the external id, then it should default to the node name. The external id is also deprecated and probably not something we should continue to rely on (https://github.com/kubernetes/kubernetes/blob/release-1.0/pkg/api/v1/types.go#L1256)

The problem is that in both the case of the hostname and host_id, its non deterministic that you will ever get back the node name.

DirectXMan12:
this is quite old.  please re-open if you are still experiencing the issue

PR #753, opened 2015-12-01
Add new 'node_name' label to expose the node name.

mwringe:
See https://github.com/kubernetes/heapster/issues/731

### DIFF
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -101,6 +101,7 @@ var targetLabelNames = map[string]struct{}{
 	sinksApi.LabelLabels.Key:             {},
 	sinksApi.LabelHostname.Key:           {},
 	sinksApi.LabelHostID.Key:             {},
+	sinksApi.LabelNodeName.Key:           {},
 	sinksApi.LabelPodNamespaceUID.Key:    {},
 	sinksApi.LabelContainerBaseImage.Key: {},
 }
--- a/sinks/api/decoder.go
+++ b/sinks/api/decoder.go
@@ -71,6 +71,7 @@ func (self *decoder) getPodLabels(pod *cache.PodElement) map[string]string {
 	labels[LabelLabels.Key] = util.LabelsToString(pod.Labels, ",")
 	labels[LabelHostname.Key] = pod.Hostname
 	labels[LabelHostID.Key] = pod.ExternalID
+	labels[LabelNodeName.Key] = pod.NodeName
 
 	return labels
 }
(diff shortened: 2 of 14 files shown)


mwielgus:
Can we somehow fix hostname instead of adding another label which will do more-or-less the same?

mwringe:
This has nothing to do with the hostname. The hostname and the node name are completely separate.

This more similarities to the existing 'host_id' label. We could change the host_id label to be node name, but this changes existing behaviour and may cause backwards compatibility issues.

The problem is that there is currently no label which will always give you the name of the node. Being able to query based on the node name is an important feature, especially if you end up migrating your node from one machine to another.

mwielgus:
Can you make sure that this label doesn't go to GCM - we are already hitting the limit (10) of labels AFAIR.

mwringe:
I don't really have a setup for testing with GCM. I can add in something to prevent it from being added to the GCM sink if that is required, but I can't easily test and verify it.

Is there a unit test to take into account this limitation with GCM?

mwielgus:
Unfortunately there is no unit test

Filter it in these places:
https://github.com/kubernetes/heapster/blob/master/sinks/gcm/core.go#L254
https://github.com/kubernetes/heapster/blob/master/sinks/gcm/core.go#L281
https://github.com/kubernetes/heapster/blob/master/sinks/gcm/driver.go#L33

mwringe:
This sounds like a very specific GCM issue, which I think should be handled by someone who is more familiar with the GCM sink and this issue. I don't use GCM and I don't have a system configured to test on GCM at the moment.

vishh:
ok to test

vishh:
ok to test

vishh:
Ping @mwringe @mwielgus : How do we make progress on this PR?

mwringe:
@vishh I am not sure how exactly to make progress with this PR. There is an issue where the GCM sink can only support 10 labels and this PR pushes the label count to 11.

So with this PR the GCM sink wont be able to function properly. I don't have a GCM account setup to test, and I don't know how exactly to handle this situation.

It seems like someone familiar with the GCM sink should be making the decisions on how it should work.  You may need to open up a new configuration option for GCM to allow admins to specify which labels they actually want or not, or perhaps just hardcode it in the sink.

vishh:
We might have to define GCM labels which might be a subset of all the
labels.

jimmidyson:
@vishh @mwringe What needs doing to get this merged?

mwringe:
The GCM sink needs to be updated to support handling more than 10 labels. Or make it more generic and add an option to be able to filter to any sink the list of labels to be considered.

mwielgus:
Please do the change in heapster-scalability branch if it is still needed.

PR #767, opened 2015-12-07
Remove sink/api

mwielgus:
The relevant stuff has been moved/refactored to heapster/core.
Validate will be replaced by some more detailed debug/info page.


manager/sinks.go
```diff
@@ -14,6 +14,7 @@
 
 package manager
 
+/*
```
piosz: Remove this as well.

piosz:
LGTM modulo small nit

manager/sinks.go
mwielgus: done

PR #773, opened 2015-12-09
Implemented pod aggregator.

piosz:


processors/pod_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+package processors
+
+import (
+	"fmt"
+
+	"k8s.io/heapster/core"
+)
+
+type PodAggregator struct {
+	MetricsToAggregate []string
```
mwielgus: All metrics should be aggregated.

processors/pod_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+	LabelsToPopulate   []core.LabelDescriptor
```
mwielgus: This should rather be a constant.

processors/pod_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+
+				for _, metricName := range this.MetricsToAggregate {
+					metricValue, found := metricSet.MetricValues[metricName]
+					if !found {
+						continue
+					}
+					aggregatedValue, found := pod.MetricValues[metricName]
+					if found {
+						if aggregatedValue.ValueType != metricValue.ValueType {
+							return nil, fmt.Errorf("PodAggregator: type not supported in %s", metricName)
```
mwielgus: Inconsistent type.

processors/pod_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+						}
+
+						if aggregatedValue.ValueType == core.ValueInt64 {
+							aggregatedValue.IntValue += metricValue.IntValue
+						} else if aggregatedValue.ValueType == core.ValueFloat {
+							aggregatedValue.FloatValue += metricValue.FloatValue
+						} else {
+							return nil, fmt.Errorf("PodAggregator: type not supported in %s", metricName)
```
mwielgus: Same here.

processors/pod_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+						}
+					} else {
+						aggregatedValue = metricValue
+					}
+					pod.MetricValues[metricName] = aggregatedValue
+				}
+			} else {
+				return nil, fmt.Errorf("No namespace and/or pod info in container %s: %v", key, metricSet.Labels)
```
mwielgus: Do not crash.

processors/pod_aggregator.go
piosz: Done

processors/pod_aggregator.go
piosz: Done

processors/pod_aggregator.go
piosz: I don't agree. We should have pod and namespace specified for every pod container. If there is not probably source implementation doesn't work as expected.

processors/pod_aggregator.go
piosz: Discussed offline. Leaving as it was.

processors/pod_aggregator.go
piosz: Done

piosz:
PTAL

processors/pod_aggregator.go
piosz: Discussed offline,.

mwielgus:
LGTM

PR #774, opened 2015-12-09
0.19.0 release

mwielgus:
cc: @piosz

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,12 @@
 # Release Notes
 
+## 0.19.0 (9-12-2015)
+- Switched to InfluxDB 0.9. The data layout was changed.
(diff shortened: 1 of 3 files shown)


RELEASES.md
```diff
@@ -1,5 +1,12 @@
 # Release Notes
 
+## 0.19.0 (9-12-2015)
+- Switched to InfluxDB 9.4. The database schema was changed.
```
vishh: Let's just say InfluxDB v0.9. That will let us rev up InfluxDB without a heapster release.

vishh: BTW the schema hasn't changed. The way columns are stored was changed.

vishh:
LGTM. We will have to update the config files also.

PR #776, opened 2015-12-09
Clarify the doc about getting set up with Google Cloud Monitoring

a-robinson:
with respect to where Container Engine metrics will be. Multiple users
have been confused by this doc.

@vishh


vishh:
LGTM

PR #779, opened 2015-12-11
Use curl instead of wget in integration/

mwielgus:
So that the integration tests can be fired on Mac


piosz:
```
Run gofmt on the following files:
 ./integration/framework.go
```

piosz:
LGTM

PR #781, opened 2015-12-11
Refresh pods and nodes every 1 hour

mwielgus:
Extra safety mechanism preventing from getting out of sync with apiserver. 
@vish - see fwd email about HPA

cc: @jszczepkowski @piosz


piosz:
LGTM

piosz:
We should probably make a patch release for v0.18 with the change.

vishh:
LGTM. We need to cherrypick this change into the release-0.18 branch.

PR #783, opened 2015-12-11
Upgrade InfluxDB to v0.9.6

vishh:


vishh:
I tested this change manually. Merging.

PR #785, opened 2015-12-14
Set node list refresh to 1 hour

mwielgus:


piosz:
LGTM

piosz:
```
sinks/influxdb/driver.go:306: result of fmt.Errorf call not used
```

mwielgus:
sinks/influxdb/driver.go - there is no such file, something got mis-sync...

PR #793, opened 2015-12-15
Fix issue found by go vet - error is not returned by influxdb/driver.…

pwittrock:
…go in new

### DIFF
--- a/sinks/influxdb/driver.go
+++ b/sinks/influxdb/driver.go
@@ -303,7 +303,7 @@ func newClient(c config) (influxdbClient, error) {
 func new(c config) (sink_api.ExternalSink, error) {
 	client, err := newClient(c)
 	if err != nil {
-		fmt.Errorf("issues while creating an InfluxDB sink: %v, will retry on use", err)
+		return nil, fmt.Errorf("issues while creating an InfluxDB sink: %v, will retry on use", err)
 	}
 	return &influxdbSink{
 		client: client, // can be nil


sinks/influxdb/driver.go
```diff
@@ -303,12 +303,12 @@ func newClient(c config) (influxdbClient, error) {
 func new(c config) (sink_api.ExternalSink, error) {
 	client, err := newClient(c)
 	if err != nil {
-		fmt.Errorf("issues while creating an InfluxDB sink: %v, will retry on use", err)
+		err = fmt.Errorf("issues while creating an InfluxDB sink: %v, will retry on use", err)
```
vishh: Should we just `return nil, err` here instead?

vishh:
cc @mwielgus: This seems like a concerning bug. Should we cherry-pick this into v0.19?

pwittrock:
ok to test

vishh:
@pwittrock: Feel free to self-merge once you address the comment.

sinks/influxdb/driver.go
pwittrock: Probably?  The comment below says it is ok for the client to be nil, but I think you are right and it should return a nil sink in case of error.

pwittrock: Done

pwittrock:
SGTM, will wait until the checks are green.

PR #809, opened 2015-12-17
remove unused file http.go

huangyuqi:
In the location of https://github.com/kubernetes/heapster/blob/master/sources/datasource/kubelet.go#L42, a method called `postRequestAndGetValue` has been realized. So this file is unuseful. It can remove.


mwielgus:
LGTM

PR #810, opened 2015-12-18
Fix grafana dashboards

vicki-c:
- revert changes in 7bc9b591a74545fc2253d6c07dc8d3d8393ddd3f (the dashboards from that commit seem messed up - they're not even valid json)
- fix derivative queries for CPU (influxdb 0.9.4.1 correctly fails a query when a derivative is grouped by an aggregate, but there is no nested aggregate in the derivative)

Let me know if I missed some important change from 7bc9b591a74545fc2253d6c07dc8d3d8393ddd3f. This is just to make Grafana functional for me.


vicki-c:
I signed it!

vishh:
Thanks for the PR @vicki-c. For CPU usage, I think we need to use `non_negative_derivative` function instead of `derivative`.

vicki-c:
@vishh changed!

vishh:
LGTM

maklemenz:
@vicki-c thank you for your work.
@vishh can you please make a release of that?

vishh:
Docker Image updated.

PR #813, opened 2015-12-21
Core types for Eventer

mwielgus:
cc: @piosz

### DIFF
--- a/events/core/types.go
+++ b/events/core/types.go
@@ -1,0 +1,46 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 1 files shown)


events/core/types.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"time"
+
+	kube_api "k8s.io/kubernetes/pkg/api"
+)
+
+type EventBatch struct {
+	// When this batch was created.
+	Timestamp time.Time
```
gmarek: We're not using -stamp suffix in k8s API. This is not an external API, but I guess it'd be nice to be consistent: 
https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#naming-conventions

events/core/types.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+type EventSink interface {
+	Name() string
+
+	// Exports data to the external storge. The funciton should be synchronous/blocking and finish only
+	// after the given EventBatch was written. This will allow sink manager to push data only to these
+	// sinks that finished writing the previous data.
+	ExportData(*EventBatch)
+	Stop()
```
gmarek: Please add comment on assumed semantics - I guess that it lets in-flight ExportData to finish before killing the Sink (in case someone would want to implement Stop() by os.Exit(0))

events/core/types.go
mwielgus: We have timestamp here:
https://github.com/kubernetes/heapster/blob/heapster-scalability/metrics/core/types.go#L76

events/core/types.go
mwielgus: Added.

gmarek:
LGTM

PR #818, opened 2015-12-22
Fix container aggregation

### LINKED ISSUE #817: [Grafana] Container metrics aggregated across namespaces and pods
(Containers dashboard)
Container metrics are aggregated across namespaces and pods even when a namespace and/or pod is selected via Grafana's drop-down menus.

Let's say I have a container `web` running in both namespaces `ns1` and `ns2`, in 2 different pods `pod1` and `pod2`. The metrics I see for the container `web` in pod `pod1` running in namespace `ns1` is the aggregation of both `web` containers running in both aforementioned namespaces.

antoineco:
Fixes #817

### DIFF
--- a/grafana/dashboards/containers.json
+++ b/grafana/dashboards/containers.json
@@ -82,12 +82,22 @@
                 "groupByTags": [],
                 "hide": false,
                 "measurement": "memory/limit_bytes_gauge",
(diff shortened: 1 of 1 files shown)


antoineco:
Sample, before/after:

mwielgus:
LGTM

PR #825, opened 2015-12-28
Remove duplicate namespace in podname.

aclisp:
The `podname` in the for loop is already `<namespace> / <podname>`. 
See https://github.com/kubernetes/heapster/blob/master/model/impl.go#L181

So this `GetNodePods` will return duplicate namespace in podname.


aclisp:
I signed it!

piosz:
Closing as obsolete.

PR #830, opened 2015-12-29
Flow manager for events

mwielgus:
cc:@piosz

### DIFF
--- a/events/manager/manager.go
+++ b/events/manager/manager.go
@@ -1,0 +1,80 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


events/manager/manager.go
```diff
@@ -0,0 +1,80 @@
...
+	rm.stopChan <- struct{}{}
+}
+
+func (rm *realManager) Housekeep() {
+	for {
+		// Always try to get the newest metrics
+		now := time.Now()
+		start := now.Truncate(rm.frequency)
+		end := start.Add(rm.frequency)
+		timeToNextSync := end.Sub(now)
```
gmarek: Please add a comment why you're doing it this way, instead of using time.After(rm.frequency). My guess is that you want to do all polls in the same time.

events/manager/manager.go
```diff
@@ -0,0 +1,80 @@
...
+			return
+		}
+	}
+}
+
+func (rm *realManager) housekeep() {
+	// No parallelism. Assumes that the events are pushed to Heapster. Add paralellism
+	// when this stops to be true.
+	events := rm.source.GetNewEvents()
+	glog.Infof("Exporting %d events", len(events.Events))
```
gmarek: Please add an explicit verbosity level - even if you want it to be 0.

gmarek:
Only nits. LGTM after you fix them.

events/manager/manager.go
mwielgus: Done

events/manager/manager.go
mwielgus: Done

PR #833, opened 2015-12-29
Sink/Source factories for Eventer

mwielgus:
cc: @piosz

### DIFF
--- a/events/sinks/factory.go
+++ b/events/sinks/factory.go
@@ -1,0 +1,53 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


events/sinks/factory.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, fmt.Errorf("Sink not recognized: %s", uri.Key)
+	}
+}
+
+func (this *SinkFactory) BuildAll(uris flags.Uris) []core.EventSink {
+	result := make([]core.EventSink, 0, len(uris))
+	for _, uri := range uris {
+		sink, err := this.Build(uri)
+		if err != nil {
+			glog.Errorf("Failed to create sink: %v", err)
```
gmarek: Wouldn't it be useful to somehow expose those errors? Eg. return list of them or slice of pairs, or sth? I don't want to block this PR on this, so feel free to ignore this comment if you feel it'd be pointless.

gmarek:
LGTM.

events/sinks/factory.go
mwielgus: I have a plan to make sink creation error a fatal one and force all sinks to handle retries/reconnections appropriately. In some next PR.

PR #834, opened 2015-12-30
Eventer main function

mwielgus:
cc: @piosz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ deps:
 build: clean deps
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build ./...
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -o heapster k8s.io/heapster/metrics
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -o eventer k8s.io/heapster/events
 
 sanitize:
 	hooks/check_boilerplate.sh
--- a/events/eventer.go
+++ b/events/eventer.go
@@ -1,0 +1,113 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -13,6 +13,7 @@ deps:
 build: clean deps
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build ./...
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -o heapster k8s.io/heapster/metrics
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -o eventer k8s.io/heapster/eventes
```
gmarek: s/eventes/events/

events/eventer.go
```diff
@@ -0,0 +1,113 @@
...
+		glog.Fatal(err)
+	}
+	if len(sources) != 1 {
+		glog.Fatalf("Wrong number of sources: %d", len(sources))
+	}
+
+	// sinks
+	sinksFactory := sinks.NewSinkFactory()
+	sinkList := sinksFactory.BuildAll(argSinks)
+	for _, sink := range sinkList {
```
gmarek: Generally it's more efficient to use `for i := range bla { bla[i] }` as this way you're avoiding making copies. I guess that Sinks are small enough for it to doesn't matter though.

events/eventer.go
```diff
@@ -0,0 +1,113 @@
...
+	setMaxProcs()
+	glog.Infof(strings.Join(os.Args, " "))
+	glog.Infof("Eventer version %v", version.HeapsterVersion)
+	if err := validateFlags(); err != nil {
+		glog.Fatal(err)
+	}
+
+	// sources
+	if len(argSources) != 1 {
+		glog.Fatal("wrong number of sources specified")
```
gmarek: Add "Requires exactly one source"

events/eventer.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		glog.Fatal(err)
+	}
+	manager.Start()
+
+	glog.Infof("Starting eventer")
+	<-quitChannel
+}
+
+func validateFlags() error {
```
gmarek: Maybe 'validateFlagValues'? My first thought was that it's some flag-library thingy.

gmarek:
One bigger issue and few nits. Otherwise LGTM.

events/eventer.go
mwielgus: They are interfaces/pointers.

Makefile
mwielgus: Done. Need to change this file so i have the same on mac and linux.

events/eventer.go
mwielgus: Done

events/eventer.go
mwielgus: Ack.

gmarek:
I guess that heapster-integration errors are still expected. LGTM in such case.

PR #840, opened 2015-12-31
Influxdb sink for events

mwielgus:
cc: @piosz

### DIFF
--- a/events/sinks/influxdb/influxdb.go
+++ b/events/sinks/influxdb/influxdb.go
@@ -1,0 +1,192 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package influxdb
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"k8s.io/heapster/events/core"
+	metrics_core "k8s.io/heapster/metrics/core"
+	influxdb_common "k8s.io/heapster/util/influxdb"
+	kube_api "k8s.io/kubernetes/pkg/api"
+
+	"github.com/golang/glog"
+	influxdb "github.com/influxdb/influxdb/client"
+)
+
+type influxdbSink struct {
+	client influxdb_common.InfluxdbClient
+	sync.RWMutex
+	c        influxdb_common.InfluxdbConfig
+	dbExists bool
+}
+
+const (
+	eventMeasurementName = "log/events"
+	// Event special tags
+	eventUID = "uid"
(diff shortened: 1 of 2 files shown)


events/sinks/influxdb/influxdb.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (sink *influxdbSink) resetConnection() {
+	glog.Infof("Influxdb connection reset")
+	sink.dbExists = false
+	sink.client = nil
+}
+
+// Generate point value for event
+func getEventValue(event *kube_api.Event) (string, error) {
+	bytes, err := json.MarshalIndent(event, "", " ")
```
fgrzadkowski: Why do you want to indent this? I think it will consume more bytes as a result and since you sent it to influxdb anyway it probably doesn't matter how it's formatted.

mwielgus: It is how it is written in the old/current Heapster.

events/sinks/influxdb/influxdb.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if sink.dbExists {
+		return nil
+	}
+	q := influxdb.Query{
+		Command: fmt.Sprintf("CREATE DATABASE %s", sink.c.DbName),
+	}
+	if resp, err := sink.client.Query(q); err != nil {
+		if !(resp != nil && resp.Err != nil && strings.Contains(resp.Err.Error(), "already exists")) {
```
fgrzadkowski: It took quite some time to understand this statement. Can you please add a comment? Something like:

```We want to return error only if it is not "already exists" error```

File: events/sinks/influxdb/influxdb.go [hunk 3]
```diff
@@ -0,0 +1,190 @@
...
+			return fmt.Errorf("Database creation failed: %v", err)
+		}
+	}
+	sink.dbExists = true
+	glog.Infof("Created database %q on influxDB server at %q", sink.c.DbName, sink.c.Host)
+	return nil
+}
+
+// Returns a thread-compatible implementation of influxdb interactions.
```
fgrzadkowski: What does it mean `influxdb interactions`? Does `interactions` have some special meaning in heapster?

events/sinks/influxdb/influxdb.go
mwielgus: Added. This code was copy-pasted from the old heapster.

events/sinks/influxdb/influxdb.go
mwielgus: Comment replaced.

events/sinks/influxdb/influxdb.go
fgrzadkowski: Interesting... Maybe a TODO to check this?

events/sinks/influxdb/influxdb.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if sink.dbExists {
+		return nil
+	}
+	q := influxdb.Query{
+		Command: fmt.Sprintf("CREATE DATABASE %s", sink.c.DbName),
+	}
+	if resp, err := sink.client.Query(q); err != nil {
+		//We want to return error only if it is not "already exists" error
```
fgrzadkowski: Please add space after `//` and `.` at the end.

mwielgus: Done

events/sinks/influxdb/influxdb.go
mwielgus: Added.

fgrzadkowski:
LGTM

PR #841, opened 2015-12-31
Rename util to common

mwielgus:
cc: @piosz


fgrzadkowski:
I'm not sure why, but lgtm :)

mwielgus:
There were "utils" at two levels - one at the "global" level and one in metrics and events.

PR #844, opened 2016-01-01
Include Eventer binary in Docker image

mwielgus:
cc: @piosz


deploy/docker/Dockerfile
```diff
@@ -9,4 +9,6 @@ RUN for cert in `ls -1 /etc/ssl/certs/*.crt | grep -v /etc/ssl/certs/ca-certific
 # cAdvisor discovery via external files.
 VOLUME /var/run/heapster/hosts
 ADD heapster /heapster
+ADD eventer /eventer
```
piosz: eventer is not started anywhere

mwielgus: Yes it is not, you can can specify what to start in yaml. By default Heapster4Metrics is started.

piosz: ack

piosz:
LGTM

PR #847, opened 2016-01-04
add prometheus for heapster

huangyuqi:
@mwielgus 
**Add status information to Heapster**
including:
+ time of the last metrics scrape
+ histogram of Kubelet/source response times
+ number of scraped nodes and containers
+ time spent in scrape (single run, not cumulative)
+ time spent in processors (single run, not cumulative)
+ time spent in exporting data to sinks (single run, not cumulative)
+ number of http model api requests
+ number of other http requests

### DIFF
--- a/metrics/api/v1/model_handlers.go
+++ b/metrics/api/v1/model_handlers.go
@@ -193,21 +193,33 @@ func (a *Api) RegisterModel(container *restful.Container) {
 
 // availableMetrics returns a list of available cluster metric names.
 func (a *Api) availableClusterMetrics(request *restful.Request, response *restful.Response) {
+
+	//number of http model api requests add 1
+	core.ModelApiRequestCount.Inc()
 	a.processMetricNamesRequest(core.ClusterKey(), response)
 }
 
 // availableMetrics returns a list of available node metric names.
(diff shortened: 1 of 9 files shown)


huangyuqi:
@mwielgus , Please help me to review this, thanks so much. ^_^

mwielgus:
LGTM, thank you for the PR

huangyuqi:
thanks @mwielgus ,
BTW, Are there other tasks or proposals to let me join?  ^_^

mwielgus:
So:
* We need the similar set of stats thing for Events/Eventer:
  - last event timestamp
  - last housekeep timestamp
  - total number of events
  - number of runs in the current houskeep interval
  - time spent for scraping events
  - time spent for writing events (total and on per sink basis, please add this to metrics as well)

* We are moving close to code complete stage but some packages have poor test coverage. Please look around and check where tests can be added (I created an issue for Influxdb sink but there are probably other places that require tests as well)

huangyuqi:
thanks, @mwielgus ,
In recent two days, I have anatomized all codes of reconstituted version.So i think i am suitable for adding these.

PR #851, opened 2016-01-04
Fix pod list in e2e integeration tests

mwielgus:
The current code produces a list with N pointers to the same iterator.


piosz:
LGTM

PR #852, opened 2016-01-04
Expose network stats for pods

### LINKED ISSUE #368: Heapster not getting network rx and tx bytes info
Hi

I am running heapster on kubernetes. Several pods are running but all of them giving 0 values in rx and tx network stats.
On doing docker stats <container-id>, I found that network stats are also coming 0 for that container but network stats are coming in cadvisor correctly.
Network stats for the pod gcr.io/google_containers/pause:0.8.0 are coming correctly but they are not communicated into the influxdb by the heapster. The containers corresponding to this gcr.io/google_containers/pause:0.8.0 are showing 0 network stats. 

What should I do to get correct network stats ?
Please help !

jimmidyson:
Fixes #368 

Dirty: using leaky infra container name, regex for container name to pod mapping, but seems to work OK...

/cc @vishh @akash010 @smarterclayton @simon3z

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -26,6 +26,7 @@ import (
 	sink_api "k8s.io/heapster/sinks/api"
 	"k8s.io/heapster/sinks/cache"
 	source_api "k8s.io/heapster/sources/api"
+	"k8s.io/heapster/sources/datasource"
 )
 
 // Manager provides an interface to control the core of heapster.
@@ -172,6 +173,9 @@ func (rm *realManager) housekeep(start, end time.Time) {
 			errors = append(errors, err.Error())
 		}
 	}
+
+	rm.mapInfraContainersToPods(&sd)
+
 	glog.V(2).Infof("completed scraping data from sources. Errors: %v", errors)
 	if err := rm.cache.StorePods(sd.data.Pods); err != nil {
 		errors = append(errors, err.Error())
@@ -184,6 +188,44 @@ func (rm *realManager) housekeep(start, end time.Time) {
 	}
 	if len(errors) > 0 {
 		glog.V(1).Infof("housekeeping resulted in following errors: %v", errors)
+	}
+}
+
+// Map infra containers to pods - only if we have any pods of course.
+// Clunky but required without significant API changes (coming soon!).
+func (rm *realManager) mapInfraContainersToPods(sd *syncData) {
+	if len(sd.data.Pods) > 0 {
+		sd.mutex.Lock()
+		defer sd.mutex.Unlock()
+
+		// Iterate slice in reverse as we are going to be deleting containers if we manage to match infra containers.
+		for containerIndex := len(sd.data.Containers) - 1; containerIndex >= 0; containerIndex-- {
+			cont := sd.data.Containers[containerIndex]
+			podName, podNameFound := cont.Spec.Labels[datasource.KubernetesPodNameLabel]
+			podNamespace, podNamespaceFound := cont.Spec.Labels[datasource.KubernetesPodNamespaceLabel]
+
+			// Not a pod container then continue.
+			if !podNameFound || !podNamespaceFound {
+				continue
+			}
+
+			// Find a pod with these details.
+			for podIndex := range sd.data.Pods {
+				pod := &sd.data.Pods[podIndex]
+				// If we find a matching pod then add the container to the pod's containers slice.
+				if pod.Name == podName && pod.Namespace == podNamespace {
+					cont.Hostname = pod.Hostname
+					cont.ExternalID = pod.ExternalID
+					cont.Name = datasource.KubernetesPodInfraContainerName
+					pod.Containers = append(pod.Containers, cont)
+
+					// Delete the matched container so we don't duplicate containers.
+					sd.data.Containers = append(sd.data.Containers[:containerIndex], sd.data.Containers[containerIndex+1:]...)
+
+					break
+				}
+			}
+		}
 	}
 }
 
--- a/sources/datasource/kubelet.go
+++ b/sources/datasource/kubelet.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"regexp"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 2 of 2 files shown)


smarterclayton:
If leaky is imported just to get pod infra container name, it's not worth it.  I'd just have it be a constant or parameter in this code.

jimmidyson:
retest this please

jimmidyson:
Removed `leaky` package dependency.

jimmidyson:
Got more work to do on this to expose via the API so changing to WIP - please don't merge.

@vishh @mwielgus Do you know if the `Jenkins GCE e2e` should pass? Seeing unrelated flake in logs:

    No error is expected but got expected [gcm] sinks, found []

mwielgus:
Yeah, the problem seems to be unrelated. BTW, we will have to redo this work in heapster-scalability branch.

jimmidyson:
@mwielgus Are you targeting the new metrics APIs in the scalability branch? Sorry I've not kept up to speed with it.

mwielgus:
If the new API is delivered on time then yes we will switch, if not we will keep using the old one.

jimmidyson:
Ready for review please. Jenkins GCE e2e is flaky & needs to either disabled or fixed separately to this PR.

api/v1/types/model_types.go
```diff
@@ -50,10 +50,12 @@ type StatsResponse struct {
...
 }
 
-// An ExternalEntityListEntry represents the latest CPU and Memory usage of a model entity.
+// An ExternalEntityListEntry represents the latest CPU, Memory & network usage of a model entity.
 // A model entity can be a Pod, a Container, a Namespace or a Node.
 type ExternalEntityListEntry struct {
 	Name     string `json:"name"`
 	CPUUsage uint64 `json:"cpuUsage"`
 	MemUsage uint64 `json:"memUsage"`
+	RxBytes  uint64 `json:"rx_bytes"`
```
vishh: Do we need network metrics in the model or is it necessary only for monitoring purposes?
The cost for adding metrics to the model is kindda high for now.

sources/datasource/kubelet.go
```diff
@@ -176,11 +185,23 @@ func (self *kubeletSource) getAllContainers(url string, start, end time.Time) ([
 	}
 
 	// TODO(vmarmol): Use this for all stats gathering.
-	// Don't include the Kubernetes containers, those are included elsewhere.
+	// Only include the Kubernetes infra containers, the others are collected via the kubelet pod stats API.
```
vishh: I think we want to collect the system containers like `/kubelet`, `/docker-daemon`, etc which should have been the existing behavior.

api/v1/types/model_types.go
jimmidyson: So are you saying it shouldn't be returned by API queries, only passed to sinks? I'd prefer it to be exposed via the REST API & that means using the model, doesn't it?

sources/datasource/kubelet.go
jimmidyson: Sorry that's a *bad* comment! Will update - raw containers are of course retained.

api/v1/types/model_types.go
vishh: Yeah. Is it **required** to be exposed via APIs as well?

jimmidyson: Point me to the requirements :) Don't we make it up as we go along? ;)

Honestly, if it's too expensive to store a couple of extra values per pod then I don't mind dropping it from there & just leave it as passed to sinks.

sources/datasource/kubelet.go
jimmidyson: Updated

api/v1/types/model_types.go
vishh: I'd personally prefer exposing this and filesystem stats as well. Its just that the default resource limits in Kube has to change once these metrics are added. So as long as we can go fix the limits, then adding these metrics is OK by me :)

jimmidyson: OK how about I remove from the model for this version & revisit it for @mwielgus' rewritten version?

vishh: SGTM. So is monitoring the primary use case?

jimmidyson: Monitoring & accounting, yes.

jimmidyson: Can't think of a use for autoscaling on network traffic right now.

manager/manager.go
```diff
@@ -187,6 +191,44 @@ func (rm *realManager) housekeep(start, end time.Time) {
...
+			if !podNameFound || !podNamespaceFound {
+				continue
+			}
+
+			// Find a pod with these details.
+			for podIndex := range sd.data.Pods {
+				pod := &sd.data.Pods[podIndex]
+				// If we find a matching pod then add the container to the pod's containers slice.
+				if pod.Name == podName && pod.Namespace == podNamespace {
+					cont.Hostname = pod.Hostname
```
vishh: Ideally, the pod infra container should be hidden inside heapster. That way if we collect metrics from rocket for example, which does not need an infra container, we will not break users of heapster.

sources/datasource/kubelet.go
```diff
@@ -32,7 +33,12 @@ import (
 )
 
 // TODO(vmarmol): Use Kubernetes' if we export it as an API.
-const kubernetesPodNameLabel = "io.kubernetes.pod.name"
+// Copied from k8s.io/kubernetes/pkg/kubelet/dockertools/labels.go - not exported there :(
```
vishh: The reason is that these labels are not intended to be exposed :)

vishh:
General structure LGTM. Thanks @jimmidyson !!

manager/manager.go
jimmidyson: Suggestions on how to do that? Right now I can't think of one tbh.

jimmidyson:
Network stats are now only sent to sinks, not retrievable via Heapster API as discussed with @vishh.

manager/manager.go
vishh: We need stats at the pod level in addition to container level.

jimmidyson: Bit confused by this. We're adding the infra container to the appropriate pod so this is done.

jimmidyson:
@vishh Please can you let me know what is outstanding for this PR to be merged?

vishh:
The only issue with this PR is that of exposing the infra container. I'd prefer adding pod level metrics and exposing network as a pod level metrics.

jimmidyson:
This is how's it's been done in `heapster-scalability` branch & I'm not going to duplicate that functionality as this is just a quick fix until `heapster-scalability` branch becomes master.

vishh:
Ok then. No issues in that case.

jimmidyson:
Thanks @vishh.

Merging.

PR #859, opened 2016-01-05
Include infra container metrics in pod metrics - enables network metrics

jimmidyson:
@mwielgus As discussed - pretty simple in the end following your advice!


jimmidyson:
@mwielgus: One question I have: with this approach (infra container metrics only being included in pod metrics, not as separate container) will the infra container metrics be sent to sinks?

mwielgus:
LGTM, thanks for the PR.

mwielgus:
Sinks get all metric sets (containers as well as pods, namespaces and cluster aggregates) and all data that are in metricsets. And of course they can write only a subset of data. Moreover, the exactly same data are available through the API (api is implemented as sink).

jimmidyson:
Thanks for the explanation.

PR #861, opened 2016-01-06
Improve Prometheus instrumentation for REST API, sources, sinks & processors

jimmidyson:
This reworks the previous Prometheus instrumentation to label sources, sinks & processors so they can be timed independently, as well as adding instrumentation to REST API handlers, labelled appropriately.


jimmidyson:
@mwielgus @huangyuqi Take a look when you get a chance please.

huangyuqi:
Yes, i will.This instrumentation is more better. :+1:

metrics/util/metrics/http.go
```diff
@@ -0,0 +1,295 @@
...
+	case "OPTIONS", "options":
+		return "options"
+	case "NOTIFY", "notify":
+		return "notify"
+	default:
+		return strings.ToLower(m)
+	}
+}
+
+func sanitizeCode(s int) string {
```
mwielgus: Why do we need this?

jimmidyson: Labels are text values in Prometheus metrics, this is converting & (normalizing where necessary) to a string in an efficient way I guess. This is lifted straight from Prometheus codebase but let me do a quick benchmark to see if it's worthwhile.

mwielgus: I guess that the Heapster API traffic will be in most cases smaller than 10qps and this code is a bit of overkill.

jimmidyson: Fair enough.

metrics/util/metrics/http.go
```diff
@@ -0,0 +1,295 @@
...
+func (f *fancyResponseWriterDelegator) ReadFrom(r io.Reader) (int64, error) {
+	if !f.wroteHeader {
+		f.WriteHeader(http.StatusOK)
+	}
+	n, err := f.ResponseWriter.(io.ReaderFrom).ReadFrom(r)
+	f.written += n
+	return n, err
+}
+
+func sanitizeMethod(m string) string {
```
mwielgus: Same as below.

jimmidyson: Removed.

metrics/util/metrics/http.go
jimmidyson: Removed.

mwielgus:
LGTM

Issue #787, opened 2015-12-14
No data received if one or more nodes not ready in kubernetes


FlavioF:
Hi,

I am using Heapster with Kubernetes. My cluster has a size of 4 nodes.
When one of the nodes stops (and changed state to not ready), the Heapster stops to send data to InfluxDB.

In the example, I stopped the node 3 of my cluster.

Heapster is logging this message:

```
E1214 12:37:55.008564       1 kube_nodes.go:59] Failed to get container stats from Kubelet on node "node-3-samantha.internal"
E1214 12:38:00.008963       1 kube_nodes.go:59] Failed to get container stats from Kubelet on node "node-3-samantha.internal"
E1214 12:38:35.000742       1 kube_nodes.go:59] Failed to get container stats from Kubelet on node "node-3-samantha.internal"
E1214 12:39:05.003094       1 kube_nodes.go:59] Failed to get container stats from Kubelet on node "node-3-samantha.internal"
E1214 12:39:35.004320       1 kube_nodes.go:59] Failed to get container stats from Kubelet on node "node-3-samantha.internal"
```

```
NAME                                        LABELS                                                             STATUS
node-1-samantha.internal   kubernetes.io/hostname=node-1-samantha.internal   Ready
node-2-samantha.internal   kubernetes.io/hostname=node-2-samantha.internal   Ready
node-3-samantha.internal   kubernetes.io/hostname=node-3-samantha.internal   NotReady
node-4-samantha.internal   kubernetes.io/hostname=node-4-samantha.internal   Ready
```

Data presented in Grafana:
<a href='http://postimg.org/image/u1plc0ejj/' target='_blank'><img src='http://s11.postimg.org/u1plc0ejj/Selection_012.jpg' border='0' alt="Selection 012" /></a>

Versions:

> Heapster Canary
> Kubernetes 1.0.1
> InfluxDB 0.9.4.1

vishh:
That's not expected. Heapster will continue its operations even if it encounters any errors.
cc @mwielgus

mwielgus:
1. Can you tell me the exact Heapster version (last git commit)?
2. Can you share some more details about your deployment? Is it Amazon, GCE or some bare-metal cluster.
3. Did Heapster stop writing the metrics completely? (No metrics for other pods as well)
4. Did the node come back to life? If yes, did Heapster recover?

FlavioF:
1. I was using the canary docker image in docker hub, but now I changed the version to 0.19.0 and it is still happening
[...]

PR #862, opened 2016-01-06
bug fix: add node state check in getting node list

huangyuqi:
~~fixed~~ ref https://github.com/kubernetes/heapster/issues/787
when structuring node list used for getting metrics, heapster needs to check node's state.

### DIFF
--- a/sources/nodes/kube.go
+++ b/sources/nodes/kube.go
@@ -97,6 +97,15 @@ func (self *kubeNodes) getNodeInfoAndHostname(node api.Node) (Info, string, erro
 	mem := node.Status.Capacity[api.ResourceMemory]
 	nodeInfo.CpuCapacity = uint64(cpu.MilliValue())
 	nodeInfo.MemCapacity = uint64(mem.Value())
+
+	for _, condition := range node.Status.Conditions {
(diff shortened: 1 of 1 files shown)


jimmidyson:
This is a good check but i think there must be some other problem too in the scraping loop in the case where a node becomes unavailable during scrape?

huangyuqi:
yes, i agree.
When a node becomes unavailable during scrape, heapster will have some other problem. Like this:

```
E0107 04:32:55.001706    8037 kube_nodes.go:59] Failed to get container stats from Kubelet on node "8.8.8.239"
```
Because, the `unnormal` node still exists in cache, and heapster will get metrics as before.
But I think when the node is delete from cache, in `runGc()`. Heapster will work normally.

piosz:
LGTM
Thanks for the change, although I'm not sure whether it actually fixes #787, so no closing the issue before it's verified.

piosz:
add to whitelist

piosz:
@k8s-bot ok to test

huangyuqi:
thanks, @piosz

piosz:
Fail not related.

PR #864, opened 2016-01-06
Add Eventer to e2e test yaml

mwielgus:


deploy/kube-config/standalone-test/heapster-controller.yaml
```diff
@@ -22,10 +22,22 @@ spec:
         command:
         - /heapster
         - --source=kubernetes:https://kubernetes.default
+        - --sink=log
```
piosz: I'm afraid there is too many logs that come from log sink.

piosz:
LGTM

deploy/kube-config/standalone-test/heapster-controller.yaml
mwielgus: This is only for tests.

mwielgus:
Some strange jenkins flakiness - ignoring for now as it seems to be unrelated to the pr.

PR #866, opened 2016-01-07
GKE http endpoint

mwielgus:


piosz:
@a-robinson could you please take a look?

piosz:
LGTM, please rebase

a-robinson:
Sorry I missed this the last couple days. Taking a look now

metrics/core/labels.go
```diff
@@ -101,6 +101,7 @@ var podLabels = []LabelDescriptor{
 	LabelPodNamespace,
 	LabelPodNamespaceUID,
 	LabelLabels,
+	LabelCustomMetricName,
```
a-robinson: Is this used anywhere by this PR?

metrics/api/v1/types/types.go
```diff
@@ -44,38 +45,10 @@ type Point struct {
 // TimeseriesSchema represents all the metrics and labels.
 type TimeseriesSchema struct {
 	// All the metrics handled by heapster.
-	Metrics []MetricDescriptor `json:"metrics,omitempty"`
+	Metrics []core.MetricDescriptor `json:"metrics,omitempty"`
```
a-robinson: I'm a little worried that moving the definitions of the API types into the core makes it more likely that they'll be changed in the future without considering that they're part of what needs to be a stable API.

Is there any particular motivation for this change that justifies the risk?

metrics/api/v1/api.go
```diff
@@ -16,20 +16,38 @@ package v1
...
 
 type Api struct {
 	runningInKubernetes bool
 	metricSink          *metricsink.MetricSink
+	gkeMetrics          map[string]core.MetricDescriptor
+	gkeLabels           map[string]core.LabelDescriptor
 }
 
 // Create a new Api to serve from the specified cache.
 func NewApi(runningInKuberentes bool, metricSink *metricsink.MetricSink) *Api {
```
a-robinson: It's an existing problem, but it wouldn't hurt to fix the "Kuberentes" typo while you're here

integration/heapster_api_test.go
```diff
@@ -268,24 +268,28 @@ func runHeapsterMetricsTest(fm kubeFramework, svc *kube_api.Service) error {
...
-			if !exists {
-				return fmt.Errorf("timeseries: %v does not contain common label: %v", ts, label)
-			}
-		}
 		podName, podMetric := ts.Labels[core.LabelPodName.Key]
-		if podMetric {
-			for _, label := range core.PodLabels() {
+
+		/*
+			TODO(mwielgus): Uncoment once all labels are populated from POD watch
```
a-robinson: Without having made it through the actual code changes, commenting out an existing test scares me. Does this mean the new API output won't contain everything the old API output does?

mwielgus: We hope to have exactly same data. Just FYI - this we are rewriting Heapster so some functionality is not yet present. In the mean time we want to have the tests passing so some of them (that will obviously fail) are commented out.

metrics/api/v1/types/types.go
mwielgus: Ok good point.

metrics/core/labels.go
mwielgus: No

PR #867, opened 2016-01-07
Drop set/get sinks endpoint

mwielgus:
Volatile configuration changes are IMHO a bad idea. You never know when Heapster is rebooted and restored with default settings.

### DIFF
--- a/integration/heapster_api_test.go
+++ b/integration/heapster_api_test.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


jimmidyson:
:+1:

piosz:
LGTM

vishh:
This was added originally to simplify configuration. Instead of supporting 10 different configuration files and making our salt scripts complex, we can deploy a standalone heapster and then dynamically add/remove sinks.

mwielgus:
Thank you for the explanation. Allowing non-persistent configuration is just asking for troubles/outage. And crashes/restarts do happen. For that reasons we run Heapster under replication controller and not as a standalone Pod.

vishh:
I meant standalone configuration, not a pod. The default kube setup can
bring up heapster without any sink and once the sinks are setup, admins can
register those sinks.
Static configuration is one of the biggest pain points today with existing
monitoring solutions.

piosz:
And what would happen in case when Heapster is restarted?

vishh:
Config needs to be persisted.

PR #870, opened 2016-01-08
Watch-based Pod processor that adds uids and cluster autoscaling metrics

mwielgus:


metrics/processors/pod_enricher.go
```diff
@@ -0,0 +1,132 @@
...
+	if !found {
+		glog.V(3).Infof("Pod not found: %s", core.PodKey(pod.Namespace, pod.Name))
+		return
+	}
+
+	// Add UID
+	podMs.Labels[core.LabelPodId.Key] = string(pod.UID)
+
+	// Add cpu requests
+	cpuRequestMilli := int64(0)
```
piosz: How about just setting cpu/mem request for containers and then let the PodAggregator to do its work?

piosz: Also how about setting limits here rather than relying on cadvisor? Limits are tightly coupled with requests.

metrics/processors/pod_enricher.go
```diff
@@ -0,0 +1,132 @@
...
+		cpuRequestMilli += thisCpuRequestMilli
+		memRequest += thisMemRequest
+	}
+
+	// set in pod
+	podMs.MetricValues[core.MetricCpuRequest.Name] = intValue(cpuRequestMilli)
+	podMs.MetricValues[core.MetricMemoryRequest.Name] = intValue(memRequest)
+
+	// set in node
+	if nodeMs, found := batch.MetricSets[core.NodeKey(pod.Spec.NodeName)]; found {
```
piosz: Again, this is just duplication of the logic of NodeAggreator.

piosz: Also you are 'enriching' nodes in PodEnricher which is confusing.

mwielgus: We don't have node aggregator so there is no duplication, but i can add it :).

metrics/processors/pod_enricher.go
mwielgus: OK

mwielgus:
All comments addressed.

mwielgus:
Merging to unblock further work as the 1.2 code complete deadline is near. @piosz please take a look - I will fix your concerns in the next PR.

metrics/processors/pod_based_enricher.go
```diff
@@ -0,0 +1,145 @@
...
+				core.LabelPodNamespace.Key:  pod.Namespace,
+				core.LabelPodName.Key:       pod.Name,
+			},
+		}
+		batch.MetricSets[core.PodKey(pod.Namespace, pod.Name)] = podMs
+	}
+	// Add UID to pod
+	podMs.Labels[core.LabelPodId.Key] = string(pod.UID)
+
+	// Add cpu requests and limits to pods
```
piosz: I believe you are adding mem requests and limits as well. Also you are adding them to containers rather that pods

piosz:
LGTM, just a nit. Feel free to fix it some time in the future.

PR #886, opened 2016-01-14
Node aggregator

mwielgus:


metrics/processors/node_aggregator.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return &result, nil
+}
+
+func nodeMetricSet(nodeName string) *core.MetricSet {
+	return &core.MetricSet{
+		MetricValues: make(map[string]core.MetricValue),
+		Labels: map[string]string{
+			core.LabelMetricSetType.Key: core.MetricSetTypeNode,
```
piosz: Populate LabelHostID as well.

piosz:
LGTM, just one comment.

metrics/processors/node_aggregator.go
mwielgus: There is no host id in pod/container.

piosz: You are right. We should add it somewhere.

PR #890, opened 2016-01-15
Update events timestamp in store

wulonghui:
Previously，`Put` events to store is by `time.Now()`, but `Get` events form store is by `LastUpdate`, it will cause sink events repeatedly.so this PR is to fix this.


piosz:
@k8s-bot ok to test

piosz:
LGTM
Thanks for the catch and fix!

PR #902, opened 2016-01-26
Add float to heapster model api

mwielgus:
This change has to be made in both master (which will be imported to K8S) and heapster-scalability.


piosz:
LGTM

PR #915, opened 2016-01-29
Do not connect null values (Grafana)

antoineco:
Prevents this:

from being falsely drawn as:


mwielgus:
LGTM

mwielgus:
Thanks for fixing this!

Issue #912, opened 2016-01-28
[heapster-scalability] Lots of "missing metadata for container" errors in the logs


mwringe:
When running on OpenShift with the heapster-scalability branch, there are a lot of error messages in the logs about missing metadata for the system level metrics. These are the metrics which are not coming from a container or pod, but from the system itself.

Eg:
> kubelet.go:132] Missing metadata for container /system.slice/libvirtd.service. Got: map[]
kubelet.go:132] Missing metadata for container /system.slice/run-user-42-gvfs.mount. Got: map[]
kubelet.go:132] Missing metadata for container /system.slice/systemd-logind.service. Got: map[]
kubelet.go:132] Missing metadata for container /system.slice/gdm.service. Got: map[]

jimmidyson:
The problem here is this bit of code - https://github.com/kubernetes/heapster/blob/heapster-scalability/metrics/sources/kubelet/cadvisor_utils.go#L23-L41. It doesn't take into account systemd based distros systems or indeed nested cgroups.

jimmidyson:
System containers are now properly identified.

PR #919, opened 2016-01-29
Only enrich running pods

jimmidyson:


jimmidyson:
/cc @mwielgus @mwringe 

Partial fix for #912 - think we need to handle this more gracefully though in general.

mwringe:
Can we get metrics for pods which are not currently in the running state? Wondering if other states, such as pending, may also have metrics which could be exposed

jimmidyson:
If pods aren't running, i.e. containers aren't started, then there are no stats for them. Docker creates & cleans up cgroups that stats are gathered from on start/stop of containers.

vishh:
FYI: The pod is not deemed running until all containers have started and
readiness checks are satisfied. For monitoring, we might want to track
individual containers even if their parent pod is not ready.

jimmidyson:
AFAIK readiness checks don't affect `Running` phase, but rather `Ready` condition.

The only phase before `Running` is `Pending` which is [documented](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L973-L990) as:

> PodPending means the pod has been accepted by the system, but one or more of the containers has not been started. This includes time before being bound to a node, as well as time spent pulling images onto the host.

So perhaps we just have to ignore pods that we can't find details for in scraped metrics rather than crash as we currently do?

mwielgus:
I think i agree with @jimmidyson.  Just in case someone needs more data - can have a flag that disables status checking (set as default to false => pass only the running pods)?

mwielgus:
Or to be more precise - enrich whatever is already there but don't create any missing/not running stuff unless the flag is set.

jimmidyson:
Due to the distributed, asynchronous nature of Kubernetes I think we should just ignore data we can't find - best effort. If there is stuff that should be visible then users should rely on their alerting infra to notify them.

mwielgus:
Ok, but it is possible that a pod is running but cadvisor doesn't know about it yet which will result in creating a half-empty metric set. Your PR doesn't handle that.

jimmidyson:
Yes this PR was really an attempt at a band-aid - not a good idea & I will close!

Cadvisor picks up cgroups as they're created via inotify so there should be no point in time that cadvisor doesn't know about running containers/pods: cgroups are created on container start, cleaned up on container stop in the filesystem & cadvisor is notified via inotify of these changes. Having said that there could be points where containers are started but stats haven't been gathered for those containers yet. Cadvisor has no concept of pods so it doesn't aggregate pods at all right now. Stats are collected on a per-container basis so no concept of atomic pod stat collection.

I still think the best idea is to just report stats that we know about, ignore pods we haven't got any stats for via the kubelet.

PR #921, opened 2016-02-01
[heapster-scalability] Handle hierarchical system containers & systemd containers

jimmidyson:


mwielgus:
Can we please not have "/" in the container name? It will create issues in the REST api as container name can be a part of the path.

jimmidyson:
Can do. It's more common to keep the leading slash for cgroup hierarchies though but it dies mess with rest api as you said.

jimmidyson:
Removed leading `/` in cgroup names.

metrics/sources/kubelet/kubelet.go
```diff
@@ -132,25 +157,12 @@ func (this *kubeletMetricsSource) decodeMetrics(c *cadvisor.ContainerInfo) (stri
 			}
 		}
 
+		// No Kubernetes metadata so treat this as a system container.
 		if cName == "" || ns == "" || podName == "" {
```
mwielgus: Are you sure you want to treat containers without name as system?

jimmidyson: Hmm I thought all containers in pods were either explicitly named in pod spec or assigned a name when created. Is that not right?

mwielgus: Yes, when everything works then everything has a name :).

But if you take a look couple lines above you will see that we sometimes extract the container name from an obfuscated docker container name and sometimes from a kubelet-assigned label. Both mechanisms can possibly fail and in that case I would rather drop the metric set/container with a warning instead of renaming it as a system.

jimmidyson: From https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L843-L845:

```go
type Container struct {
	// Required: This must be a DNS_LABEL.  Each container in a pod must
	// have a unique name.
	Name string `json:"name"`
```

mwielgus: Yes, but at that moment we don't know to which pod the container belongs. Ideally we would take container id and look up the correct pod/container, but here we don't use this mechanisms (for simplicity, maybe we should) and we need a name. In Kubernetes 1.0.x the name was NOT available as an annotation so we extract it from docker-specific container name, what can possibly (but unlikely) fail.

jimmidyson: Do we have to support Kubernetes 1.0.x btw? We could get rid of this bit of logic & rely on Docker labels if not.

Having said that, I don't see any chance of failure in extracting the k8s container name from the Docker container name. This is what is done in the logging component successfully. Pod name is also present in the Docker container name, as is namespace - we can extract all that info from the Docker container name, but as I said above I'd prefer to drop support for 1.0.x & rely on Docker labels in 1.1.x.

mwielgus: I believe that we have to support 2 last major releases.

jimmidyson: So I think the extraction of container, pod & namespace looks right to me & I think we should trust that this will work. I can't see any problems with it.

jimmidyson: So in summary, yes I personally am happy to consider any pods without ns, pod name, container name to be system containers.

mwielgus:
OK, lets see what happens.

mwielgus:
LGTM

vishh:
FYI: @jimmidyson: https://github.com/kubernetes/kubernetes/pull/20687 combined with the new metrics API in the kubelet should solve the issue of identifying interesting `system` containers.

PR #922, opened 2016-02-01
[heapster-scalability] Skip node aggregation for pods with no node metadata

jimmidyson:


mwielgus:
Do you have an idea how this pods are created?

jimmidyson:
These are pods either in pending, waiting to be scheduled to a node, or completed, containers exited with restart policy of never (e.g. build pods, jobs pods, etc).

mwielgus:
OK, lgtm

Issue #911, opened 2016-01-28
[heapster-scalability] 'No node info in pod' crashes Heapster


mwringe:
When Heapster tries to access a pod which is not currently running, it will not receive information about that pod. Currently with the heapster-scalability branch this will cause the Heapster container to crash:
>F0128 18:33:35.546050       1 manager.go:135] Error in processor: No node info in pod namespace:test-1454000850/pod:metrics-deployer-436wu: map[labels:component:deployer,metrics-infra:deployer,provider:openshift namespace_id:a323cf1e-c5e1-11e5-a5bc-3c970e88a56f type:pod namespace_name:test-1454000850 pod_namespace:test-1454000850 pod_name:metrics-deployer-436wu pod_id:423a890c-c5ed-11e5-a5bc-3c970e88a56f]

`oc get pods metrics-deployer-436wu`
> NAME                     READY     STATUS      RESTARTS   AGE
metrics-deployer-436wu   0/1       Completed   0          6m

In the `master` branch this looks to cause 404 errors to be logged, but the Heapster instance will continue to run properly.

mwringe:
Is there anyway I could get this assigned to me?

jimmidyson:
@mwringe It doesn't need to be assigned for you to work on it & provide a PR :)

mwielgus:
Does #922 fix the problem?

jimmidyson:
It should do. Sorry I didn't mark it as fixed in the PR.

mwringe:
Can we please re-open this issue?

#922 just adds a check to get around the issue and is not addressing the main problem.

There are a few things happening here and it has to do with how we handle metrics coming in from /stats endpoint and how we handle metrics which are not available under /stats.

For metrics which we get from the /stats endpoint, we go through and add the hostname label to them (https://github.com/kubernetes/heapster/blob/heapster-scalability/metrics/sources/kubelet/kubelet.go#L94)

But, we also get the list of pods from the cluster. If any of those pods are not currently accounted for in the list of metrics, then we try and add them: https://github.com/kubernetes/heapster/blob/heapster-scalability/metrics/processors/pod_based_enricher.go#L56

The issue is we are forgetting to add the hostname attribute here. So all that we need to do is to add `core.LabelNodename.Key: pod.Spec.NodeName,` and then this issue goes away (at least for completed containers, I suspect it would be the same for the other scenarios mentioned in #922). At the very least we should be logging a warning in #922 and not logging this situations as info level.

[...]

PR #930, opened 2016-02-02
Skip pods that do not have any stats collected

jimmidyson:
@mwringe As discussed in #911 this only adds pod info for pending & running pods. Pending pods need to be considered as they might be partially started, e.g. one container of a multi-container pod started. If there are no metrics collected for a pod then it is also dropped.

Thoughts @mwielgus?

### DIFF
--- a/metrics/processors/pod_based_enricher.go
+++ b/metrics/processors/pod_based_enricher.go
@@ -15,6 +15,7 @@
 package processors
 
 import (
+	"fmt"
 	"net/url"
 	"time"
 
@@ -28,7 +29,6 @@ import (
 	"k8s.io/kubernetes/pkg/client/cache"
 	kube_client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/fields"
-	"k8s.io/kubernetes/pkg/labels"
 )
 
 type PodBasedEnricher struct {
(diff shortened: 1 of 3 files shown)


jimmidyson:
Correction: as discussed in #911

metrics/processors/pod_based_enricher.go
```diff
@@ -46,26 +46,22 @@ func (this *PodBasedEnricher) Process(batch *core.DataBatch) (*core.DataBatch, e
...
 	}
 	return batch, nil
 }
 
 func addPodInfo(pod *kube_api.Pod, batch *core.DataBatch) {
-	podMs, found := batch.MetricSets[core.PodKey(pod.Namespace, pod.Name)]
+	podKey := core.PodKey(pod.Namespace, pod.Name)
+	podMs, found := batch.MetricSets[podKey]
 	if !found {
-		// create pod stub.
```
mwielgus: Is it possible for cAdvisor that a PodStatsContainer is not present but regular pod containers are? If yes then we should clean up the mess somewhere - either create the missing pod or drop the "orphaned" containers.

metrics/processors/pod_based_enricher.go
```diff
@@ -46,26 +46,22 @@ func (this *PodBasedEnricher) Process(batch *core.DataBatch) (*core.DataBatch, e
 		return nil, err
 	}
 	for _, pod := range pods {
-		addPodInfo(pod, batch)
+		// Only keep stats for pending & running pods. We keep pending pods as well
+		// as running because pending pods might be partially started (e.g. one container
+		// in a multi-container pod).
+		if pod.Status.Phase == kube_api.PodPending || pod.Status.Phase == kube_api.PodRunning {
+			addPodInfo(pod, batch)
+		}
```
mwielgus: I wonder what to do if the status is different (deleted/failed/whatever)? Shouldn't we drop the remains?

jimmidyson: Other statuses shouldn't have any running containers so will not be present in cadvisor stats.

metrics/processors/pod_based_enricher.go
jimmidyson: Cadvisor doesn't know anything about pods. Any pods that are started/running should have a pod infra container running that will ensure the PodStatsContainer exists ready here.

metrics/processors/pod_based_enricher.go
mwielgus: Yes, if everything is in perfect sync. But there is a chance that the state of pod/containers differs in api server and kubelete when a pod is killed, started etc. What then?

jimmidyson: Maybe we need to switch this round then & loop through the stats we've collected & correlate to pods from pod list? That way we can't miss anything at all & we only keep state for the pods we actually have stats for.

jimmidyson:
I've reworked this now to only loop through collected stats, creating stubs for containers that have no metrics collected for them (e.g. partially started pod) & creating stubs for containers that have no pod metrics collected (e.g. running pod containers without pod infra container - strange but possible scenario I guess).

mwielgus:
metrics/processors/pod_based_enricher.go:55: possible formatting directive in Error call
make: *** [sanitize] Error 1
Build step 'Execute shell' marked build as failure

jimmidyson:
Yeah sorry about that. Also trying to update tests to use a fake `StoreToPodLister` - any examples of that anywhere that you know?

mwielgus:
I'm not aware of any

jimmidyson:
OK working on it now & expanding test coverage to test for stubbing as required.

jimmidyson:
Something up with e2e:

```
Starting cluster using provider: gce
... calling verify-prereqs
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
```

Repeated a lot in the logs.

jimmidyson:
@k8s-bot test this please

jimmidyson:
@mwielgus This now only adds pods/containers that have stats collected. If there are missing pods without all containers or containers without pods, these are stubbed with requests/limits from pod spec (tests added for this).

jimmidyson:
@mwringe FYI in this PR the stubbed pods/containers now have proper node info for node aggregation.

Issue #926, opened 2016-02-02
Add autoscaling metrics to GCM sink


mwielgus:
Right now autoscaling suite in k8s is failing due to lack of these metrics.

cc @piosz @fgrzadkowski

PR #931, opened 2016-02-02
Node autoscaling enricher

mwielgus:
#926

Part 1 - calculate percentage based on aggregated values.
Part 2 - next PR - enable enricher and set correct metric names for GCM autoscaling

cc: @fgrzadkowski


piosz:
LGTM

PR #936, opened 2016-02-03
GCA node autoscaling support in GCM

mwielgus:
Continuation of #931 (issue #926).

### DIFF
--- a/metrics/core/labels.go
+++ b/metrics/core/labels.go
@@ -82,6 +82,14 @@ var (
 		Key:         "custom_metric_name",
 		Description: "User-defined name of the exported custom metric",
 	}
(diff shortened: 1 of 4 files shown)


metrics/core/labels.go
```diff
@@ -127,6 +135,14 @@ var gcmLabels = []LabelDescriptor{
 	LabelContainerName,
 	LabelContainerBaseImage,
 	LabelCustomMetricName,
+	LabelGCEResourceID,
```
fgrzadkowski: I guess this should be only in `gcmNodeAutoscalingLabels`

fgrzadkowski:
LGTM, except one comment.

PR #941, opened 2016-02-04
Support all available metrics in REST API

jimmidyson:
This makes the `/metrics/{metrics-name:*}` endpoint consistent with the names returned from `/metrics` & supports all available metrics for each resource type (cluster, node, pod, etc).

/cc @mwielgus @piosz

### DIFF
--- a/metrics/api/v1/model_handlers.go
+++ b/metrics/api/v1/model_handlers.go
@@ -33,7 +33,26 @@ import (
 // when manager.model has not been initialized.
 var errModelNotActivated = errors.New("the model is not activated")
 
-var metricNamesConversion = map[string]string{
+var supportedMetrics = map[string]struct{}{
+	"memory/usage":             {},
+	"network/tx":               {},
+	"memory/limit":             {},
+	"cpu/request":              {},
+	"network/rx_errors":        {},
+	"memory/major_page_faults": {},
+	"cpu/usage":                {},
+	"cpu/usage_rate":           {},
+	"network/tx_errors":        {},
+	"cpu/limit":                {},
+	"memory/working_set":       {},
+	"network/rx":               {},
+	"uptime":                   {},
+	"memory/request":           {},
+	"memory/page_faults":       {},
+}
+
+// Deprecated - clients should switch to full metric names ASAP.
+var deprecatedMetricNamesConversion = map[string]string{
 	"cpu-usage":      "cpu/usage_rate",
 	"cpu-limit":      "cpu/limit",
 	"memory-limit":   "memory/limit",
@@ -58,7 +77,7 @@ func (a *Api) RegisterModel(container *restful.Container) {
 		Operation("availableClusterMetrics"))
 
 	// The /metrics/{metric-name} endpoint exposes an aggregated metric for the Cluster entity of the model.
-	ws.Route(ws.GET("/metrics/{metric-name}").
+	ws.Route(ws.GET("/metrics/{metric-name:*}").
 		To(metrics.InstrumentRouteFunc("clusterMetrics", a.clusterMetrics)).
 		Doc("Export an aggregated cluster-level metric").
 		Operation("clusterMetrics").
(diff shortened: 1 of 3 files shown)


metrics/api/v1/model_handlers.go
```diff
@@ -33,14 +33,6 @@ import (
 // when manager.model has not been initialized.
 var errModelNotActivated = errors.New("the model is not activated")
 
-var metricNamesConversion = map[string]string{
```
mwielgus: Unfortunately you cannot remove cpu-usage mapping. This is a legacy constant used by Horizontal Pod Autoscaler when talking to Heapster.

mwielgus:
Looks good except for the mapping removal. It has to remain there.

metrics/api/v1/model_handlers.go
jimmidyson: Ah damn. How about keeping the mappings, deprecated, & supporting all metrics in the way this PR does? So check if metric is supported, if not check if there's a mapped metric?

mwielgus: Other way round, but yes. First check for mapping.

jimmidyson: Done.

metrics/api/v1/model_handlers.go
```diff
@@ -313,15 +314,22 @@ func (a *Api) podListMetrics(request *restful.Request, response *restful.Respons
...
-	metricName := getMetricName(request)
-	if metricName == "" {
-		response.WriteError(http.StatusBadRequest, fmt.Errorf("Metric not supported: %v", request.PathParameter("metric-name")))
-		return
+	metricName := request.PathParameter("metric-name")
+	convertedMetricName := metricName
+	for idx, podName := range strings.Split(request.PathParameter("pod-list"), ",") {
+		key := core.PodKey(ns, podName)
+		if idx == 0 {
+			validMetricName, valid := a.validateMetricName(metricName, key)
```
mwielgus: Some metrics may not be yet present in a metric set. Please do not enforce that.

metrics/api/v1/model_handlers.go
```diff
@@ -370,24 +378,35 @@ func (a *Api) processMetricRequest(key string, request *restful.Request, respons
 		response.WriteError(http.StatusBadRequest, err)
 		return
 	}
-	metricName := getMetricName(request)
-	if metricName == "" {
-		response.WriteError(http.StatusBadRequest, fmt.Errorf("Metric not supported: %v", request.PathParameter("metric-name")))
+	metricName := request.PathParameter("metric-name")
+	convertedMetricName, valid := a.validateMetricName(metricName, key)
```
mwielgus: Same here.

metrics/api/v1/model_handlers.go
jimmidyson: So what would you prefer? List all metrics that could be available? Have to remember that we're not aggregating all metrics, are we?

mwielgus: I'm just saying that the lack of a metric in one of the pods should not break the whole batch request. Right now the result for that pod will be empty and I would rather keep this behavior.

jimmidyson: Agreed - I'll update to validate for any known metric names & it will return empty for those that are not populated.

jimmidyson: So now this PR return all possible metrics in `/metrics` requests, validate possible ones in `/metrics/{metric-name:*}`, return only for unknown metrics, regardless of what is cached.

metrics/api/v1/model_handlers.go
```diff
@@ -33,7 +33,26 @@ import (
 // when manager.model has not been initialized.
 var errModelNotActivated = errors.New("the model is not activated")
 
-var metricNamesConversion = map[string]string{
+var supportedMetrics = map[string]struct{}{
```
mwielgus: there can also be custom metrics that can have arbitrary name. Please only hardcode the deprecated metrics that require conversions.

jimmidyson: So what do you want to return in `/metrics` requests? Only available (collected) metrics for the requested resource?

mwielgus: Yes.

mwielgus: Can you apply the comment?

mwielgus: Or let's do it other way round, I will merge and apply a quick fix. I need enable custom metrics ASAP and the current code block them.

jimmidyson:
Sorry -on vacation for a couple of weeks. Thanks for picking this up.

PR #959, opened 2016-02-10
Startup retry logic in GCM sink

mwielgus:


piosz:
Let's discuss offline tomorrow.

piosz:
LGTM

PR #960, opened 2016-02-11
Added waiting for GCE token.

### LINKED ISSUE #958: Debug k8s gce autoscaling suite
It is failing despite gcm sink with autoscaling enabled.

piosz:
fix #958


mwielgus:
LGTM but we will have to reorganize managers and factories after 1.2.

PR #966, opened 2016-02-11
Reuse databatch in processors.

mwielgus:
There is no reason to copy all of the data in every processor.

cc: @jimmidyson


piosz:
LGTM

Issue #967, opened 2016-02-11
[heapster-scalability] certificate based authentication broken


mwringe:
It looks like certificate authentication was broken when this commit was put into place: https://github.com/kubernetes/heapster/commit/8ad5353a3dd6dc00eeb24eb4da778e566d26178c

As it stands right now its not possible to use client based authentication to secure the heapster endpoints.

fgrzadkowski:
Fixed in #968

Please let us know if this is still the problem.

PR #968, opened 2016-02-11
Add back in support for client based authentication.

mwringe:
Resolves https://github.com/kubernetes/heapster/issues/967

### DIFF
--- a/metrics/heapster.go
+++ b/metrics/heapster.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


liggitt:
Any chance we could add a test to make sure this doesn't regress again?

mwielgus:
LGTM

mwielgus:
@liggitt AFAIR this code has never been in heapster-scalability.

mwringe:
@mwielgus It believe it did work in the heapster-scalablity branch before this commit: https://github.com/kubernetes/heapster/commit/8ad5353a3dd6dc00eeb24eb4da778e566d26178c

@liggitt Yeah, we should really have a test for this to prevent regressions

mwielgus:
Ok sorry, I must have missed it during the review. Yes, definitely a test would be helpful.

PR #969, opened 2016-02-12
Increased timeout while waiting for GCE token

piosz:


piosz:
Merging without LGTM. PTAL later

PR #982, opened 2016-02-16
Update kubernetes godeps to include latest kubelet stats API

timstclair:
Prerequisite for https://github.com/kubernetes/heapster/pull/927

/cc @dchen1107 @vishh

PR #983, opened 2016-02-17
More robust file checks for setting up integration tests

timstclair:
Somehow heapster jenkins nodes got into a state where the kubernetes tmp directory was present but empty. By checking for the actual required files rather than the directory, this should avoid such problems in the future.

See https://github.com/kubernetes/heapster/pull/982 for problematic build logs.

/cc @vishh @pwittrock


vishh:
LGTM

mwielgus:
LGTM

PR #985, opened 2016-02-17
Rate metrics converter

### LINKED ISSUE #984: [heapster-scalability] Convert missing rate metrics from cumulatives
https://github.com/kubernetes/heapster/blob/master/sinks/gcm/rate_metrics.go

mwielgus:
Fixes #984 
Needs unit tests and manual verification.

cc: @piosz @timstclair

### DIFF
--- a/metrics/core/types.go
+++ b/metrics/core/types.go
@@ -86,7 +86,6 @@ type MetricValue struct {
 	FloatValue float32
 	MetricType MetricType
 	ValueType  ValueType
-	Start      time.Time
 }
 
 func (this *MetricValue) GetValue() interface{} {
@@ -119,6 +118,8 @@ func (this *LabeledMetric) GetValue() interface{} {
 }
 
 type MetricSet struct {
+	CreateTime     time.Time
+	ScrapeTime     time.Time
 	MetricValues   map[string]MetricValue
 	Labels         map[string]string
 	LabeledMetrics []LabeledMetric
--- a/metrics/processors/rate_calculator.go
+++ b/metrics/processors/rate_calculator.go
@@ -1,0 +1,80 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package processors
+
+import (
+	"k8s.io/heapster/metrics/core"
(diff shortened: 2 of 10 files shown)


metrics/core/types.go
```diff
@@ -119,6 +118,8 @@ func (this *LabeledMetric) GetValue() interface{} {
 }
 
 type MetricSet struct {
+	CreateTime     time.Time
+	ScrapeTime     time.Time
```
timstclair: What about metrics that could be scraped at different times? Volume / filesystem stats will be updated on a different schedule from other stats, as will UserDefinedMetrics.

metrics/processors/rate_calculator.go
```diff
@@ -0,0 +1,77 @@
...
+				continue
+			}
+
+			for metricName, targetMetric := range this.rateMetricsMapping {
+				metricValNew, foundNew := newMs.MetricValues[metricName]
+				metricValOld, foundOld := oldMs.MetricValues[metricName]
+				if foundNew && foundOld {
+					if metricName == core.MetricCpuUsageRate.MetricDescriptor.Name {
+						// cpu/usage values are in nanoseconds; we want to have it in millicores (that's why constant 1000 is here).
+						newVal := 1000 * (metricValNew.IntValue - metricValOld.IntValue) /
```
timstclair: How about definining a scale somewhere (either with the RateMetricsMapping, or maybe as part of the `core.UnitsType` in the MetricsDescriptor), then you don't need to branch on the metric name here

metrics/processors/rate_calculator.go
```diff
@@ -0,0 +1,77 @@
...
+							(newMs.ScrapeTime.UnixNano() - oldMs.ScrapeTime.UnixNano())
+
+						newMs.MetricValues[targetMetric.MetricDescriptor.Name] = core.MetricValue{
+							ValueType:  core.ValueInt64,
+							MetricType: core.MetricGauge,
+							IntValue:   newVal,
+						}
+
+					} else if targetMetric.MetricDescriptor.ValueType == core.ValueFloat {
+						newVal := 1000000000 * float32(metricValNew.IntValue-metricValOld.IntValue) /
```
timstclair: (nit) prefer `time.Second.Nanoseconds()` or `1e9`

metrics/core/types.go
mwielgus: Scrape time is (at least now) used only for converting rate metrics to cumulative. Do you have any cumulative metric that is scraped at a very different time? This is an internal structure that can be changed in the future. Right now I'm trying to reduce a size of a single metric as we are adding couple more of them.

metrics/processors/rate_calculator.go
mwielgus: Good idea - will take a look.

mwielgus:
Unit test added

metrics/core/types.go
timstclair: This should be ok for now, especially if it's internal-only.

mwielgus:
Added couple other changes required for this to work:
* do not aggregate container cumulative metrics 
* print scrape time in logsink

piosz:
LGTM

PR #995, opened 2016-02-19
Handle 'delta' metrics in the API conversion

timstclair:
Missed this in https://github.com/kubernetes/heapster/pull/927

/cc @vishh @mwielgus


mwielgus:
LGTM

Issue #1001, opened 2016-02-24
[heapster scalability] Port OpenTSDB


mwielgus:
cc: @piosz @fgrzadkowski

mwielgus:
I will port the code but someone should test if it really works.

mwielgus:
#1000 needs porting.

mwielgus:
#1007 - #1000 ported.

mwielgus:
Ok, so we have a basic support for metrics in opentsdb. Events are pending/TBD.

piosz:
Fixed by @mwielgus

PR #703, opened 2015-11-11
Add opentsdb sink

bluebreezecf:
OpenTSDB is a distributed, scalable Time Series Database (TSDB) written on top of HBase. 
OpenTSDB was written to address a common need: store, index and serve metrics collected from computer systems (network gear, operating systems, applications) at a large scale, and make this data easily accessible and graphable.

This PR adds opentsdb sink to heapster and makes metrics and events of k8s can be pushed to OpenTSDB, which could be a scalable store-end.

PR #1004, opened 2016-02-24
Ported OpenTSDB sink from master

mwielgus:
Partial fix for #1001
Based on #703 by @bluebreezecf


piosz:
LGTM Though I'm not familiarized with OpenTSDB

PR #1017, opened 2016-02-29
Fix influxdb float export

### LINKED ISSUE #1015: [heapster-scalability] Floats in influxdb are aggregated to 0
```
select sum(value),count(value) from "network/rx_rate" where time >'2016-02-26T16:00:00Z' group by time(60s)
```
produces only zeros
Either we are putting the floats somehow as strings or "wrong" floats.

Blocks graphana updata.

cc: @piosz

mwielgus:
Fixes #1015


piosz:
LGTM

PR #1018, opened 2016-02-29
Increased timeout while waiting for GCE metadata

piosz:
ref https://github.com/kubernetes/kubernetes/issues/22176


mwielgus:
LGTM

PR #1022, opened 2016-03-01
Fix backwards summary_api supported logic

timstclair:
Unfortunately the summarySupported logic was backwards, so the summary_api has not actually been enabled. This invalidates some of my previous testing, but not all of it (in many cases I manually disabled the fallback to ensure that it was testing the summary API).

/cc @dchen1107 @vishh @fgrzadkowski @mwielgus @piosz

### DIFF
--- a/metrics/sources/summary/summary.go
+++ b/metrics/sources/summary/summary.go
@@ -76,7 +76,7 @@ func NewSummaryMetricsSource(node NodeInfo, client *kubelet.KubeletClient, fallb
 	return &summaryMetricsSource{
 		node:          node,
 		kubeletClient: client,
-		useFallback:   summarySupported(node.KubeletVersion),
+		useFallback:   !summarySupported(node.KubeletVersion),
 		fallback:      fallback,
 	}
 }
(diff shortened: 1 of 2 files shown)


timstclair:
On the positive side, the fallback code is thoroughly tested :)

fgrzadkowski:
Tim, it seems that this code hasn't been tested at all, including fallback. The only part that has been tested is using old kubelet API from new source, which doesn't bring much value and information. Summary API is a very significant change and we need to be sure we don't break anything.

Initially we decided not to rollback your previous PR (https://github.com/kubernetes/kubernetes/pull/22101) as we believed that the tests were passing. Now it seems that nothing of this is true.

I propose to do the following:
 * revert https://github.com/kubernetes/kubernetes/pull/22101 
 * review & merge this PR
 * enable summary API only in a separate jenkins job (you can use `GCE autoscaling` suite)
 * do some thorough tests and write down the scenario with results

If it turns out that the new API is working we can enable it back.

If it's needed I'm happy to have a VC to discuss it further.

fgrzadkowski:
@timstclair Thanks for the explanation! Please let us know if you need any assistance with these tests.

fgrzadkowski:
@mwielgus Can you please review this PR?

mwielgus:
LGTM

PR #1023, opened 2016-03-01
Do not export pod level metrics to GKE

### LINKED ISSUE #1021: GKE endpoint should return only 1 sample per metric
@a-robinson @piosz @fgrzadkowski

mwielgus:
Fixes #1021


piosz:
LGTM

PR #1029, opened 2016-03-01
DO NOT MERGE - summary API correctness testing

timstclair:
Add test-only `CompareSource` which performs a side-by-side comparison of the metrics gathered from the `kubernetes.summary_api` source and the `kubernetes` source. See [Heapster Summary Testing](https://docs.google.com/document/d/1ZUYslYiSBjI6V7rhCRxyh4lH-CdMjX45F_0DZ65ZaxA/edit#heading=h.dulh3aqcsdfv) for more details (sorry that the link is google-only, LMK if you'd like access).

### Results

First, a few special cases:

- times (uptime, scrape time, create time) use a fuzzy comparison since the calculation is based on the exact scrape / parse time.
- pod cpu & memory metrics (`cpu/usage, memory/{working_set, usage, page_faults, major_page_faults}`) are ignored, since these metrics measured the pod infra container usage, which is not surfaced in the summary API.
- label `container_base_image` - Filled in by the pod enricher from the k8s API, and hence ignored here.
- labeled metrics - Handled differently by the summary API. These metrics will need to be validated through other means.

Aside from these exceptions, the only remaining discrepancy is in networking:
```
I0301 23:08:05.000222       1 manager.go:79] Scraping metrics start: 2016-03-01 23:07:30 +0000 UTC, end: 2016-03-01 23:08:00 +0000 UTC
node:e2e-test-stclair-minion-gilj metric network/tx mismatch: summary = {902197626 0 0 0} kubelet = {699783929 0 0 0}
node:e2e-test-stclair-minion-gilj metric network/rx mismatch: summary = {2916806622 0 0 0} kubelet = {2695792187 0 0 0}
namespace:kube-system/pod:kube-proxy-e2e-test-stclair-minion-gilj metric network/rx mismatch: summary = {2916841897 0 0 0} kubelet = {2695826628 0 0 0}
namespace:kube-system/pod:kube-proxy-e2e-test-stclair-minion-gilj metric network/tx mismatch: summary = {902226555 0 0 0} kubelet = {699792580 0 0 0}
```
This is due to the summary API reporting the sum over all interfaces, while the kubelet source just reports the stats from eth0. From offline discussions, we probably want to change the summary API to use that same behavior.

### DIFF
--- a/deploy/kube-config/standalone-test/heapster-controller.yaml
+++ b/deploy/kube-config/standalone-test/heapster-controller.yaml
@@ -21,7 +21,7 @@ spec:
         imagePullPolicy: Always
         command:
         - /heapster
-        - --source=kubernetes:https://kubernetes.default
+        - --source=kubernetes.compare:https://kubernetes.default
         - --sink=log
         volumeMounts:
         - name: ssl-certs
--- a/metrics/sources/factory.go
+++ b/metrics/sources/factory.go
@@ -17,6 +17,8 @@ package sources
 import (
 	"fmt"
 
+	"github.com/golang/glog"
+
 	"k8s.io/heapster/common/flags"
 	"k8s.io/heapster/metrics/core"
 	"k8s.io/heapster/metrics/sources/kubelet"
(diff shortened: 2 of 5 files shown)


timstclair:
FYI - @fgrzadkowski @mwielgus @piosz @vishh @pwittrock @dchen1107

timstclair:
Filed https://github.com/kubernetes/kubernetes/issues/22319 for the networking stats.

PR #1035, opened 2016-03-02
Do not aggregate cpu/mem usage on node level

### LINKED ISSUE #1032: Memory value is doubled in GKE

piosz:
fix #1032


mwielgus:
lgtm

PR #1042, opened 2016-03-03
Use curl instead od wget in Dockerfile

mwielgus:
Something is broken with wget


vishh:
Did you accidentally include two other files?

mwielgus:
Yep, removed.

vishh:
LGTM

vishh:
Let's wait for the tests to pass.

PR #1047, opened 2016-03-04
Sort log sink output

timstclair:
For easier comparison of log output.


mwielgus:
Thanks for fixing this.

LGTM

PR #1064, opened 2016-03-14
Export pod level metrics for GKE

piosz:
cc @a-robinson


metrics/api/v1/api.go
```diff
@@ -218,6 +218,9 @@ func (a *Api) processMetricsRequest(shortStorage []*core.DataBatch) []*types.Tim
 			if msType == core.MetricSetTypeNode {
 				ts.Labels[core.LabelContainerName.Key] = "machine"
 			}
+			if msType == core.MetricSetTypePod {
+				ts.Labels[core.LabelContainerName.Key] = "/pod"
```
vishh: nit: Why not just `pod`?

PR #1070, opened 2016-03-15
Hawkular: Update to use the nodename when storing system level metrics

mwringe:
Update to use the node name when generating the id for non-container metrics instead of the deprecated externalID value (eg HostID).

Parameter can be used to get back the old behaviour if required

### DIFF
--- a/metrics/sinks/hawkular/client.go
+++ b/metrics/sinks/hawkular/client.go
@@ -107,7 +107,17 @@ func (self *hawkularSink) idName(ms *core.MetricSet, metricName string) string {
 	if ms.Labels[core.LabelPodId.Key] != "" {
 		n = append(n, ms.Labels[core.LabelPodId.Key])
 	} else {
-		n = append(n, ms.Labels[core.LabelHostID.Key])
+		if &self.labelNodeId != nil {
+			if v, found := ms.Labels[self.labelNodeId]; found {
+				n = append(n, v)
+			} else {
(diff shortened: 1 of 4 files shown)


piosz:
@jimmidyson could you please take a look?

metrics/sinks/hawkular/client.go
```diff
@@ -107,7 +107,17 @@ func (self *hawkularSink) idName(ms *core.MetricSet, metricName string) string {
 	if ms.Labels[core.LabelPodId.Key] != "" {
 		n = append(n, ms.Labels[core.LabelPodId.Key])
 	} else {
-		n = append(n, ms.Labels[core.LabelHostID.Key])
+		if &self.labelNodeId != nil {
+			if v, found := ms.Labels[self.labelNodeId]; found {
+				n = append(n, v)
+			} else {
+				glog.Infof("The labelNodeId was set to %s but there is no label with this value."+
```
jimmidyson: Should this be a debug log rather than info?

jimmidyson:
One nit but other LGTM.

metrics/sinks/hawkular/client.go
mwringe: Which V level exactly then? just V(1) or something higher?

jimmidyson: Let's go with 4.

mwringe:
Updated to use V(4) logging as suggested. Squashed commits.

jimmidyson:
LGTM - waiting for tests before merge. Thanks @mwringe.

PR #1072, opened 2016-03-16
Debugging doc update

mwielgus:


docs/debugging.md
```diff
@@ -10,11 +10,59 @@ Depending on the deployment setup, the issue could be either with heapster, cadv
...
+ ``` 
+This is enabled for metrics only.
+
 #### Extra Logging
 
-If the '/validate' endpoint does not provide enough information, additional logging can be enabled by setting an extra flag. This requires restarting heapster though.
-Add `--vmodule=*=4` flag to heapster. When using the docker image or when running in kubernetes, pass an extra environment variable `FLAGS="--vmodule=*=4`. 
-If you are running heapster on kubernetes, the environment variable needs to be added to the `env` section in [heapster-controller.yaml](../deploy/kube-config/standalone/heapster-controller.yaml).
+Moreover additional logging can be enabled by setting an extra flag `--vmodule=*=4`. 
+You can also enable a sink that writes all metrics or events to stdout with `--sink=log` added to command line paramenters.
```
piosz: typo: should be parameters

piosz:
LGTM just a nit

PR #1077, opened 2016-03-17
Source configuration doc update

mwielgus:


docs/source-configuration.md
```diff
@@ -1,8 +1,8 @@
 Configuring sources
 ===================
 
-Heapster can get data from multiple sources. These are specified on the command line
-via the `--source` flag. The flag takes an argument of the form `PREFIX:CONFIG[?OPTIONS]`.
+Heapster can get data from multiple sources (although at this momeent we support only one kind - Kuberntes).
```
piosz: typo: kubernetes

docs/source-configuration.md
```diff
@@ -1,8 +1,8 @@
...
+They specified on the command line via the `--source` flag. The flag takes an argument of the form `PREFIX:CONFIG[?OPTIONS]`.
```
piosz: They are specified?

docs/source-configuration.md
mwielgus: Done.

docs/source-configuration.md
mwielgus: Done.

piosz:
LGTM, though we should probably improve this document since summary-api is now a default source in Kubernetes and it seems to be an optional one from what is written here.

@timstclair would you like to elaborate a little bit about summary-api?

timstclair:
Do we want to make the summary API the only kubernetes source? I believe this was stated as a goal at some point. If that's still the case, then I don't think we need to add much here...

piosz:
Not yet. My concern here is that after reading the doc it seems that kubernetes source is a default one while kubernetes.summary_api is just an alternative. It'd great to update the docs thought it's rather low priority.

mwielgus:
From the shared node doc it is not clear how do you want to handle custom metrics. As long as the presence of custom metrics in the summary api is not fully confirmed and agreed I won't drop the old API that has them.

PR #1086, opened 2016-03-22
Fix warning formatting in NamespaceBasedEnricher

mwielgus:
%s is printed as %!s(bool=false)

### DIFF
--- a/metrics/processors/namespace_based_enricher.go
+++ b/metrics/processors/namespace_based_enricher.go
@@ -61,7 +61,11 @@ func (this *NamespaceBasedEnricher) addNamespaceInfo(metricSet *core.MetricSet) 
 					glog.Errorf("Wrong namespace store content")
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


metrics/processors/namespace_based_enricher.go
```diff
@@ -61,7 +61,12 @@ func (this *NamespaceBasedEnricher) addNamespaceInfo(metricSet *core.MetricSet)
 					glog.Errorf("Wrong namespace store content")
 				}
 			} else {
-				glog.Warningf("Failed to get namespace: exist: %s error: %v", exists, err)
+				if !exists {
+					glog.Warningf("Namespace doesn't exist: %s", namespaceName)
+				}
+				if err != nil {
```
piosz: else if? We don't want to have 2 error messages for one problem.

mwielgus: Done.

piosz:
LGTM

PR #1091, opened 2016-03-23
Temporary workaround for flaky metadata server on GCE

piosz:
ref https://github.com/kubernetes/kubernetes/issues/22176


mwielgus:
LGTM

Issue #1090, opened 2016-03-23
no cpu/usage_rate for nodes?


perhapszzy:
I see data in cpu/usage for nodes, but not for cpu/usage_rate

mwielgus:
Can you please provide more information? What Heapster version, what deployment, where do you miss cpu/usage_rate, etc.

perhapszzy:
I'm running heapster v1.0.0 on k8s v1.1. The k8s is set up on ubuntu 14.04 machines on a cloud. 

I can find result with this query:
```
select hostname,type,value from "cpu/usage"  where type='node' limit 40
```
but not this one:
```
select hostname,type,value from "cpu/usage_rate"  where type='node' limit 40
```

mwielgus:
Ok, can you enable log sink (`--sink=log`)for a moment and send us a sample from the logs?

perhapszzy:
Some log is enclosed:
[log.txt](https://github.com/kubernetes/heapster/files/186447/log.txt)

It seems all metrics are present in the log

mwielgus:
Thanks for the log. Looks like for unknown reason the rates are not generated for nodes and node system containers. One reason is that something is missing in the node metadata in cAdvisor.
What cloud are you running on?

perhapszzy:
I'm running k8s on a local IaaS provider, could the problem be caused by the version of k8s or the way I'm running the k8s on the infrastructure?

mwielgus:
Please add -v=8 for more logging - if my suspicions are correct you should find: "Skipping rates for %s - different crate time new: xxxx  old: xxx"  there. Please paste these lines and/or all logs if possible.

perhapszzy:
Right, I did find something like:
```
I0323 17:54:05.088835       1 rate_calculator.go:51] Skipping rates for node:i-075twq8l - different crate time new:2016-03-22 20:06:16.842345 +0800 +0800  old:2016-03-22 20:06:16.842345 +0800 +0800
```

perhapszzy:
All logs are enclosed:
[log1.txt](https://github.com/kubernetes/heapster/files/186572/log1.txt)

mwielgus:
I spotted a bug in our code. We will push a new image tomorrow.

mwielgus:
Thanks for reporting!

perhapszzy:
Sure. Thanks for fixing the problem so quick :)

PR #1094, opened 2016-03-23
Print create time of metrics sets in logsink

mwielgus:


mwielgus:
Merging without LGTM to move #1090 forward. @piosz - please review at the earliest convenience.

piosz:
LGTM

PR #1095, opened 2016-03-23
Compare time with equals

### LINKED ISSUE #1090: no cpu/usage_rate for nodes?

mwielgus:
Should fix #1090. Time struct contains a pointer and == doesn't always work for it.


piosz:
LGTM

PR #1096, opened 2016-03-24
1.1.0-beta2

mwielgus:
Inc version after time comparison fix.


piosz:
The version 1.1.0-beta1 haven't been released yet, so closing this.

PR #1097, opened 2016-03-24
Revert "Temporary workaround for flaky metadata server on GCE"

piosz:
Reverts kubernetes/heapster#1091

This is no longer needed once https://github.com/GoogleCloudPlatform/gcloud-golang/issues/194 is fixed.


mwielgus:
LGTM

PR #1105, opened 2016-03-30
Enable namespace_id for GKE

mwielgus:
cc: @a-robinson

### DIFF
--- a/metrics/core/labels.go
+++ b/metrics/core/labels.go
@@ -115,8 +115,7 @@ var podLabels = []LabelDescriptor{
 	LabelPodName,
 	LabelPodId,
 	LabelPodNamespace,
-	// TODO: Enable label once it is supported by NamespaceEnricher
-	//	LabelPodNamespaceUID,
+	LabelPodNamespaceUID,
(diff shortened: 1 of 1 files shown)


a-robinson:
LGTM, but can we have a unit test? I thought that there used to be a test that covered all the fields we expected to be present?

mwielgus:
There is a test that checks whether the exported metrics have the labels mentioned in the schema. But if a label is not mentioned in the schema (as it was in this case) then the test won't indicate any troubles.

If you have a separate list of labels that have to be in the dump we can use it to double-check that everything is exported as expected.

a-robinson:
That test sounds a little self-fulfilling :)

We always want:
namespace_id
pod_id
host_id
container_name

And for disk metrics, we expect a resource_id field.

mwielgus:
Yeah, kind of :). 
I will add the explicit checks in couple minutes in the next PR. Merging this one.

piosz:
@mwielgus we should cherry-pick this, make patch release and bump Heapster version in Kubernetes.

PR #1106, opened 2016-03-30
Explicit label checks for GKE endpoint

mwielgus:


mwielgus:
cc: @a-robinson

piosz:
LGTM

integration/heapster_api_test.go
```diff
@@ -287,6 +287,20 @@ func runHeapsterMetricsTest(fm kubeFramework, svc *kube_api.Service) error {
 
 		if podMetric {
 			actualPods[podName] = true
+			// Extra explicit check that the expecte metrics are there:
+			requiredLabels := []string{
+				core.LabelPodNamespaceUID.Key,
```
a-robinson: Just a suggestion, but using the raw string literal (e.g. "namespace_id") rather than the same constant that's shared with the code would be more future-proof against accidental changes to the value of the constant.

a-robinson:
Thanks!

PR #1111, opened 2016-04-01
Don't fetch metrics from a node that we have not yet been able to reach. Fixes NPE

mwringe:
This fixes a nil pointer exception which occurs when we try and fetch data from a node which could not be reached.

### DIFF
--- a/metrics/processors/node_aggregator.go
+++ b/metrics/processors/node_aggregator.go
@@ -42,8 +42,7 @@ func (this *NodeAggregator) Process(batch *core.DataBatch) (*core.DataBatch, err
 				node, found := batch.MetricSets[nodeKey]
 				if !found {
 					glog.Warningf("Failed to find node: %s", nodeKey)
-				}
-				if err := aggregate(metricSet, node, this.MetricsToAggregate); err != nil {
+				} else if err := aggregate(metricSet, node, this.MetricsToAggregate); err != nil {
 					return nil, err
 				}
 			} else {


metrics/processors/node_aggregator.go
```diff
@@ -42,8 +42,7 @@ func (this *NodeAggregator) Process(batch *core.DataBatch) (*core.DataBatch, err
 				node, found := batch.MetricSets[nodeKey]
 				if !found {
 					glog.Warningf("Failed to find node: %s", nodeKey)
```
jimmidyson: When can we get into this state? Is this really a warning or lower log level?

I'd also prefer to just continue if not found rather than `if else`.

mwringe: > When can we get into this state?

If you set the metric_resolution to 10s then this error occurs. The default 30s doesn't have this issue. Not sure why exactly its happening in this situation, but the bug still remains that we trying to get the node, finding it doesn't exist, and then trying on the next line to use this nil object when we shouldn't be using it.

> Is this really a warning or lower log level?

I left the log level at its current state, if you think it should be a lower log level then we can change it.

> I'd also prefer to just continue if not found rather than if else.

We can't just continue because the next 'if' line tries to call the aggregate method with the nil node which causes a nil pointer exception. That is what this simple patch is trying to avoid.

mwielgus:
LGTM

metrics/processors/node_aggregator.go
jimmidyson: I meant continue as in continue with the for loop. Reads better to me but only stylistic preference.

I'm still unsure why changing the resolution makes it likely to hit this condition. Think that needs looking at more.

PR #1116, opened 2016-04-04
Update how Hawkular metric ids are handled

mwringe:
Now that we have more than just pod container metrics and node metrics we need to handle the metric ID better when storing into Hawkular Metrics.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1321349

### DIFF
--- a/metrics/sinks/hawkular/client.go
+++ b/metrics/sinks/hawkular/client.go
@@ -103,25 +103,52 @@ func (self *hawkularSink) groupName(ms *core.MetricSet, metricName string) strin
 
 func (self *hawkularSink) idName(ms *core.MetricSet, metricName string) string {
 	n := make([]string, 0, 3)
-	n = append(n, ms.Labels[core.LabelContainerName.Key])
-	if ms.Labels[core.LabelPodId.Key] != "" {
+
+	metricType := ms.Labels[core.LabelMetricSetType.Key]
+	switch metricType {
+	case core.MetricSetTypeNode:
+		n = append(n, "machine")
+		n = append(n, self.nodeName(ms))
+	case core.MetricSetTypeSystemContainer:
+		n = append(n, core.MetricSetTypeSystemContainer)
+		n = append(n, ms.Labels[core.LabelContainerName.Key])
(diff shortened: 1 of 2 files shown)


metrics/sinks/hawkular/client.go
```diff
@@ -103,20 +103,32 @@ func (self *hawkularSink) groupName(ms *core.MetricSet, metricName string) strin
...
 func (self *hawkularSink) idName(ms *core.MetricSet, metricName string) string {
 	n := make([]string, 0, 3)
-	n = append(n, ms.Labels[core.LabelContainerName.Key])
-	if ms.Labels[core.LabelPodId.Key] != "" {
+
+	metricType := ms.Labels[core.LabelMetricSetType.Key]
+	if metricType == core.MetricSetTypeNode {
+		n = append(n, "machine")
+		n = append(n, self.nodeName(ms))
+	} else if metricType == core.MetricSetTypeSystemContainer {
```
jimmidyson: This would be more idiomatic to write as a switch rather than if-else.

metrics/sinks/hawkular/client.go
```diff
@@ -124,6 +136,20 @@ func (self *hawkularSink) idName(ms *core.MetricSet, metricName string) string {
 	return strings.Join(n, separator)
 }
 
+func (self *hawkularSink) nodeName(ms *core.MetricSet) string {
+	if &self.labelNodeId != nil {
```
jimmidyson: Would prefer to flip this check & `return ms.Labels[core.LabelNodename.Key]` - no need for the else part, reduces nesting.

jimmidyson:
Couple of stylistic things, but LGTM otherwise.

mwringe:
Updated

metrics/sinks/hawkular/client.go
mwringe: I updated this to remove one of the else statements in there, not sure if that was exactly what you were expecting or not though.

jimmidyson:
LGTM - merging. Thanks @mwringe.

PR #1118, opened 2016-04-05
Remove duplicate processing from Hawkular sink

burmanm:
Transforms the old map[string]MetricValue to LabeledMetric and avoids the duplication of code in several places.

This also begs the question, why is there even two metric types? Why not remove the old map[string]MetricValue and modify this behavior to be global in the Heapster? Is there some sort of wish for different types (if so - why are they on the different depth in MetricSet type) ?

### DIFF
--- a/metrics/sinks/hawkular/client.go
+++ b/metrics/sinks/hawkular/client.go
@@ -28,37 +28,37 @@ import (
 )
 
 // Fetches definitions from the server and checks that they're matching the descriptors
-func (self *hawkularSink) updateDefinitions(mt metrics.MetricType) error {
-	m := make([]metrics.Modifier, len(self.modifiers), len(self.modifiers)+1)
-	copy(m, self.modifiers)
+func (h *hawkularSink) updateDefinitions(mt metrics.MetricType) error {
+	m := make([]metrics.Modifier, len(h.modifiers), len(h.modifiers)+1)
+	copy(m, h.modifiers)
 	m = append(m, metrics.Filters(metrics.TypeFilter(mt)))
 
-	mds, err := self.client.Definitions(m...)
+	mds, err := h.client.Definitions(m...)
 	if err != nil {
 		return err
 	}
 
-	self.regLock.Lock()
-	defer self.regLock.Unlock()
+	h.regLock.Lock()
+	defer h.regLock.Unlock()
 
 	for _, p := range mds {
 		// If no descriptorTag is found, this metric does not belong to Heapster
 		if mk, found := p.Tags[descriptorTag]; found {
-			if model, f := self.models[mk]; f {
-				if !self.recent(p, model) {
-					if err := self.client.UpdateTags(mt, p.Id, p.Tags, self.modifiers...); err != nil {
+			if model, f := h.models[mk]; f {
+				if !h.recent(p, model) {
+					if err := h.client.UpdateTags(mt, p.Id, p.Tags, h.modifiers...); err != nil {
 						return err
 					}
 				}
 			}
-			self.reg[p.Id] = p
+			h.reg[p.Id] = p
 		}
 	}
 	return nil
 }
 
 // Checks that stored definition is up to date with the model
-func (self *hawkularSink) recent(live *metrics.MetricDefinition, model *metrics.MetricDefinition) bool {
+func (h *hawkularSink) recent(live *metrics.MetricDefinition, model *metrics.MetricDefinition) bool {
 	recent := true
 	for k := range model.Tags {
 		if v, found := live.Tags[k]; !found {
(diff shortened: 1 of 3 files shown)


jimmidyson:
ok to test

jimmidyson:
@mwielgus Question for you:

> This also begs the question, why is there even two metric types? Why not remove the old map[string]MetricValue and modify this behavior to be global in the Heapster? Is there some sort of wish for different types (if so - why are they on the different depth in MetricSet type) ?

@mwringe Could you take a look please?

metrics/sinks/hawkular/client.go
```diff
@@ -28,37 +28,37 @@ import (
...
+	defer h.regLock.Unlock()
 
 	for _, p := range mds {
 		// If no descriptorTag is found, this metric does not belong to Heapster
 		if mk, found := p.Tags[descriptorTag]; found {
-			if model, f := self.models[mk]; f {
-				if !self.recent(p, model) {
-					if err := self.client.UpdateTags(mt, p.Id, p.Tags, self.modifiers...); err != nil {
+			if model, f := h.models[mk]; f {
+				if !h.recent(p, model) {
```
jimmidyson: Can join this in the previous condition (`f && !h.recent(p, model)`)

metrics/sinks/hawkular/client.go
```diff
@@ -138,83 +138,33 @@ func (self *hawkularSink) idName(ms *core.MetricSet, metricName string) string {
 	return strings.Join(n, separator)
 }
 
-func (self *hawkularSink) nodeName(ms *core.MetricSet) string {
-	if &self.labelNodeId != nil {
-		if v, found := ms.Labels[self.labelNodeId]; found {
+func (h *hawkularSink) nodeName(ms *core.MetricSet) string {
+	if &h.labelNodeId != nil {
```
jimmidyson: Shouldn't this just check if `labelNodeId` is zero string (`len(labelNodeId) != 0`)?

jimmidyson: I've just seen this in a few other places actually, unrelated to this PR - all probably need updating.

metrics/sinks/hawkular/driver.go
```diff
@@ -83,64 +83,31 @@ func (self *hawkularSink) ExportData(db *core.DataBatch) {
 	if len(db.MetricSets) > 0 {
 		tmhs := make(map[string][]metrics.MetricHeader)
 
-		if &self.labelTenant == nil {
-			tmhs[self.client.Tenant] = make([]metrics.MetricHeader, 0, totalCount)
+		if &h.labelTenant == nil {
```
jimmidyson: Same as above - should be zero length string check.

metrics/sinks/hawkular/driver.go
```diff
@@ -83,64 +83,31 @@ func (self *hawkularSink) ExportData(db *core.DataBatch) {
...
+						continue Store
 					}
 				}
 
-				tenant := self.client.Tenant
+				tenant := h.client.Tenant
 
-				if &self.labelTenant != nil {
-					if v, found := ms.Labels[self.labelTenant]; found {
+				if &h.labelTenant != nil {
```
jimmidyson: Zero length string check.

metrics/sinks/hawkular/client.go
jimmidyson: Or make optional fields `*string` & check against nil that way. But length check preferred.

metrics/sinks/hawkular/driver.go
```diff
@@ -165,29 +132,42 @@ func (self *hawkularSink) ExportData(db *core.DataBatch) {
...
-		info += fmt.Sprintf("Using label '%s' as tenant information\n", self.labelTenant)
+	h.regLock.Lock()
+	defer h.regLock.Unlock()
+	info += fmt.Sprintf("Known metrics: %d\n", len(h.reg))
+	if &h.labelTenant != nil {
+		info += fmt.Sprintf("Using label '%s' as tenant information\n", h.labelTenant)
 	}
-	if &self.labelNodeId != nil {
-		info += fmt.Sprintf("Using label '%s' as node identified in resourceid\n", self.labelNodeId)
+	if &h.labelNodeId != nil {
```
jimmidyson: Zero length string check.

mwringe:
Is there any reason we are calling the function 'pointToLabeledMetricHeader' instead of the old 'pointToMetricHeader'? If it is now handling both normal metrics and labeled metric the same, having 'LabeledMetric' in the function name might be a bit confusing.

Other than that, LGTM

burmanm:
I figured leaving it as "LabeledMetric" was easier to read since we're transforming older metrics to the "Labeled" ones (and not the other way)

jimmidyson:
@mwringe Happy?

LGTM

mwringe:
LGTM

Issue #665, opened 2015-10-21
Decide and document the scope of heapster's responsibilities


bgrant0607:
Since we're extending heapster in various ways for auto-scaling, and we're about to add support for collecting and aggregating custom metrics, we should decide and document what the boundaries for heapster's responsibilities should be. In particular, I don't want it to grow into a general-purpose application and/or infrastructure monitoring, dashboarding, and alerting system, such as [Prometheus](https://developers.soundcloud.com/blog/prometheus-monitoring-at-soundcloud). There are many such systems, and we shouldn't compete with them.

We should also document how heapster can be integrated with components that provide functionality outside its boundaries. For instance, it is already documented that Heapster supports multiple storage backends, with a few examples. Additionally, my understanding is that we plan to use collectd in order to convert a multitude of data formats to the Prometheus format. Other obvious areas include dashboards, such Graphana, and alerting, such as AlertManager. 

We should also clarify intended usage. For instance, I understand collecting a handful of metrics, such as QPS, to drive horizontal pod auto-scaling and other core functionality, but how about collection of hundreds of arbitrary metrics from a single application?

cc @vishh @jszczepkowski @mwielgus @davidopp @lavalamp @smarterclayton

jimmidyson:
I think this is an incredibly important thing to nail down. I've previously raised a proposal in #645 to utilise Prometheus for metric collection & Heapster to become a REST API with Kubernetes API semantics over the top of Prometheus.

I'd also like to point out the work I've done with Prometheus to provide autodiscovery for master, node, container & application targets for metric collection.

But that doesn't help with nailing down exactly what the function & limits of Heapster is but I feel this would provide a configurable & extensible option to users.

piosz:
cc @fgrzadkowski

fgrzadkowski:
Reassigning to @mwielgus who is driving now changes in heapster.

mwielgus:
Ref: #769

mwielgus:
The target scope and vision documented in #769. Closing this for now.

PR #1124, opened 2016-04-11
Proposal: Introduce Oldtimer

DirectXMan12:
This is a proposal for Oldtimer, the Heapster historical metrics
access component.  Oldtimer was original proposed in the vision
statement, but was not specified in any particular detail previously.

### DIFF
--- a/docs/proposals/old-timer.md
+++ b/docs/proposals/old-timer.md
@@ -1,0 +1,228 @@
+# Heapster Oldtimer
+
+## Overview
+
+Prior to the Heapster refactor, the Heapster model presented aggregations of
+metrics over certain time periods (the last hour and day).  Post-refactor, the
+concern of presenting an interface for historical metrics was to be split into
+a separate Heapster component: Oldtimer.
+
+Oldtimer will run as part of the main Heapster executable, and will present
+common interfaces for retrieving historical metrics over longer periods of time
+than the Heapster model, and will allow fetching aggregations of metrics (e.g.
+averages, 95 percentile, etc) over different periods of time.  It will do this
+by querying the sink to which it is storing metrics.
+
+Note: even though we are retrieving metrics, this document refers to the
+metrics storage locations as "sinks" to be consistent with the rest
+of Heapster.
+
+## Motivation
+
+There are two major motivations for exposing historical metrics information:
+
+1. Using aggregated historical data to make size-related decisions
+   (for example, idling requires looking for traffic over a long time period)
+
+2. Providing a common interface for users to view historical metrics
+
+Before the Heapster refactoring (see the
+[Heapster Long Term Vision Proposal](https://github.com/kubernetes/heapster/blob/master/docs/proposals/vision.md)),
+Heapster supported querying metrics aggregated over certain extended time
+periods (the last hour and day) via the Heapster model.
+
+However, since the Heapster model is stored in-memory, and not persisted to
+disk, this historical data would be "lost" whenever Heapster was restarted.
+This made it unreliable for use by system components which need a historical
+view.
+
+Since we already persist metrics into a sink, it does not make sense for
+Heapster itself to persist long-term metrics to disk itself.  Instead, we can
+just query the sink directly.
+
+## API
+
+Oldtimer will present an api somewhat similar to the normal Heapster model.
+The structure of the URLs is designed to mirror those exposed by the model API.
+When used simply to retrieve historical data points, Oldtimer will return the
+same types as the model API.  When the used to retrieve aggregations, Oldtimer
+will return special data types detailed under the "Return Types" section.
+
+### Paths
+
+`/api/v1/historical/{prefix}/metrics/`: Returns a list of all available
+metrics.
+
+`/api/v1/historical{prefix}/metrics/{metric-name}?start=X&end=Y`: Returns a set
+of (Timestamp, Value) pairs for the requested {prefix}-level metric, over the
+given time range.
+
+`/api/v1/historical{prefix}/metrics-aggregated/{aggregations}/{metric-name}?start=X&end=Y&bucket=B`
+Returns the requested {prefix}-level metric, aggregated with the given
+aggregation over the requested time period (potentially split into several
+different bucket of duration `B`).  `{aggregations}` may be a comma-separated
+list of aggregations to retrieve multiple at once.
+
+Where `{prefix}` is normally either empty (cluster-level),
+`/namespaces/{namespace}` (namespace-level),
+`/namespaces/{namespace}/pods/{pod-name}` (pod-level),
+`/namespaces/{namespace}/pod-list/{pod-list}` (multi-pod-level), or
+`/namespaces/{namespace}/pods/{pod-name}/containers/{container-name}`
+(container-level).
+
+Additionally, since pod names are not temporally unique (i.e. it is possible to
+delete a pod, and then create a new, completely different pod with the same
+name), `{prefix}` may also be `/pod-id/{pod-id}` (pod-level metrics),
+`/pod-id-list/{pod-id-list}` (multi-pod-level), or
+`/pod-id/{pod-id}/containers/{container-name}` (container-level metrics).
+
+In addition, when `{prefix}` is not empty, there will be a url of the form:
+`/api/v1/historical/{prefix-without-final-element}` which allows fetching the
+list of available nodes/namespaces/pods/containers.
+
+Note that queries by pod name will return metrics from the latest pod with the
+given name.  This may require an extra trip to the database in some cases, in
+order to determine which pod id that actually is.  For this reason, if a
+component knows the pod ids for which it is querying, using these is preferred
+to using the pod names.  The pod-name-based API is retained for the sake of
+easy queries and to match up with the model API.
+
+### Parameter Types
+
+The `start` and `end` parameters are defined the same way as for the model:
+each should be a timestamp formatted according to RFC 3339, if no start time is
+specified, it defaults to zero in Unix epoch time, and if no end time is
+specified, all data after the start time will be considered.
+
+The `bucket` (bucket duration) parameter is a number followed by any of the
+following suffixes:
+
+- `ms`: milliseconds
+- `s`: seconds
+- `m`: minutes
+- `h`: hours
+- `d`: days
+
+### Return Types
+
+For requests which simply fetch data points or list available objects, the
+return format will be the same as that used in the Heapster model API.
+
+The the case of aggregations, a different set of types is used: each bucket is
+represented by a `MetricAggregationBucket`, which contains the timestamp for
+that bucket (the start of the bucket), the count of entries in that bucket (if
+requested) as an unsigned integer, as well as each of the other requested
+aggregations, in the form of a `MetricValue` (which just holds an unsigned int
+or a float).
+
+All buckets for a particular metric are grouped together in a
+`MetricAggregationResult`, which also holds the bucket size (duration) for the
+buckets.  If multiple pods are requested, the result will be returned as a
+`MetricAggregationResultList`, similarly to the `MetricResultList` for the
+model API.
+
+```go
+type MetricValue struct {
+    IntValue *uint64
+    FloatValue *float64
+}
+
+type MetricAggregationBucket struct {
+    Timestamp time.Time
+    Count *uint64
+
+    Average *MetricValue
+    Maximum *MetricValue
+    Minimum *MetricValue
+    Median *MetricValue
+    Percentiles map[uint64]MetricValue
+}
+
+type MetricAggregationResult struct {
+    Buckets []MetricAggregationBucket
+    BucketSize time.Duration
+}
+
+type MetricAggregationResultList struct {
+    Items []MetricAggregationResult
+}
+```
+
+### Aggregations
+
+Several different aggregations will be supported.  Aggregations should be
+performed in the metrics sink.  If more aggregations later become supported
+across all metrics sinks, the list can be expanded.
+
+- Average (arithmetic mean): `/metrics-aggregated/average`
+- Maximum: `/metrics-aggregated/max`
+- Minimum: `/metrics-aggregated/min`
+- Percentile: `/metrics-aggregated/{number}-perc`
+- Median: `/metrics-aggregated/median`
+- Count: `/metrics-aggregated/count`
+
+Note: to support all the existing sinks, the supported percentiles will be
+limitted to 50, 95, and 99.  If additional percentile values later become
+supported by other sinks, this list may be expanded (see the Sink Support
+section below).
+
+### Example
+
+Suppose that one wanted to retrieve the 95th percentile of CPU usage for a
+given pod over the past 30 days, in 1 hour intervals, along with the maximum
+usage for each interval.  Call the pod "somepod", in the namespace "somens".
+To fetch the results, you'd perform:
+
+```
+GET /api/v1/historical/namespaces/somens/pods/somepod/metrics-aggregated/95-perc,average/cpu/usage?start=2016-03-20T10:57:37-04:00&bucket=1h
+```
+
+Which would then return:
+
+```json
+{
+    "bucketSize": "3600000000000",
+    "buckets": [
+        {
+            "timestamp": "2016-03-20T10:57:37-04:00",
+            "average": "32",
+            "percentiles": {
+                "95": "27"
+            }
+        },
+        ...
+    ]
+}
+```
+
+## Sink Support and Functionality
+
+When Oldtimer receives a request, it will compose a query to the sink, send the
+query to the sink, and the transform the results into the appropriate API
+formats.  Note that Oldtimer is designed to retrieve information that was
+originally written by Heapster itself.  Any information read by Oldtimer must
+have been stored according to the Heapster storage schema.
+
+All computations, filtering, etc should be performed in the sink.  Oldtimer
+should only be composing queries.  Ergo, the feature set of Oldtimer must
+represent the lowest-common-denominator of features supported by the sinks.
+Oldtimer is meant to be an API for performing basic aggregations supported by
+all of the sinks, and is not meant to be a general purpose query tool.
+
+At the time of writing of this proposal, the following sinks were considered:
+Hawkular, InfluxDB, GCM, and OpenTSDB.  However, the aggregations supported are
+fairly basic, so if new sinks are added, it should be fairly likely that they
+support the required Oldtimer features.
+
+## Scaling and Performance Considerations
+
+Since Oldtimer itself does not store any data, it should have a fairly low
+memory footprint.  The current plan is to have Oldtimer run as part of the main
+Heapster executable.  However, in the future it may be advantageous to have the
+ability to split Oldtimer out into a separate executable in order to scale it
+independently of Heapster.
+
+The metrics sinks themselves should already have clustering support, and thus
+can be scaled if needed.  Since Oldtimer queries the metrics sinks themselves,
+response latency should depend mainly on how quickly the sinks can respond to
+queries.


docs/proposals/old-timer.md
```diff
@@ -0,0 +1,144 @@
...
+of Heapster.
+
+## Motivation
+
+There are two major motivations for exposing historical metrics information:
+
+1. Using aggregated historical data to make size-related decisions
+   (for example, idling requires looking for traffic over a long time period)
+
+2. Providing a common interface to for users to view historical metrics
```
ncdc: remove the "to" after interface

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,144 @@
...
+
+Before the Heapster refactoring (see the "Heapster Long Term Vision" proposal),
```
ncdc: Link to proposal?

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,144 @@
...
+just query the sink directly.
+
+## Design
+
+### API
+
+Oldtimer will present an api somewhat similar to the normal Heapster model.
+The urls will take the forms:
+
+`/api/v1/old-timer/{prefix}/metrics/`: Returns a list of all available metrics.
```
ncdc: I can't say I particularly like "old-timer". `/api/v1/historical` or something along those lines sounds better to me.

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,144 @@
...
+Several different aggregations will be supported.  Aggregations should be
+performed in the metrics sink.  If more aggregations later become supported
+across all metrics sinks, the list can be expanded (and the API version
+should probably be bumped, since the supported aggregations should be part of
+the API).
+
+- Average (arithmetic mean): `/{metric-name}/average`
+- Maximum: `/{metric-name}/max`
+- Minimum: `/{metric-name}/min`
+- Percentile: `/{metric-name}/{number}-perc`
```
ncdc: Is `perc` currently in use? If not, maybe `pct` or not abbreviating at all would be better?

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,144 @@
...
+deploy multiple replicas of Oldtimer.  The metrics sinks themselves should
+already have clustering support, and thus can be scaled as well.  Since
+Oldtimer queries the metrics sinks themselves, response latency should
+depend mainly on how quickly the sinks can respond to queries.
+
+## Open Questions
+
+- Do the choice of percentiles need to be limited?  InfluxDB and Hawkular
+appear to support arbitrary percentile values in queries, while GCM v3 appears
+to support 99, 95, 50, 5, and OpenTSDB appears to support 50, 75, 90, 95, 99,
```
ncdc: Suggest ordering both in ascending or descending instead of mixing them

docs/proposals/old-timer.md
DirectXMan12: :heavy_check_mark:

docs/proposals/old-timer.md
DirectXMan12: :check:

docs/proposals/old-timer.md
DirectXMan12: Fair.  I like "historical" as well. :heavy_check_mark:

docs/proposals/old-timer.md
DirectXMan12: I was trying to keep it "readable".  IMO, "perc" is more readable than "pct".  I'd prefer either "perc" or not abbreviating it.

docs/proposals/old-timer.md
DirectXMan12: :heavy_check_mark:

piosz:
cc @fgrzadkowski

fabxc:
As hinted in #1129 I have my concerns about service-level metrics not being clearly separated. Or better, Heapster not holding up to its intent to _not_ be a general purpose monitoring system.

This was previously touched in #665 and I don't feel it ever actually happened. I'm worried that adding a "generic" read-path at this point will lead to Heapster eventually growing a full meta-QL for all its sinks. If it's trying to be the system through which writes happen, it just makes _sense_ to read through it as well.

Aside from the scope-issue, this has many semantical issues as different sinks have different evaluation behaviors (e.g. extrapolation), which can have drastic effects.
Also all these sinks have different query languages. I see a lot of time being spent working on mapping features onto each other and being a general contention point. You touched that point already at the end of your document.

I would be very interested in the use case for this and the benefit within the next 6 months.
As I've heard there are companies with very elaborate scheduling and auto-scaling that hardly had a need for long-term data.
My instinct tells me to defer such features until its absolutely clear what's required.

DirectXMan12:
> I would be very interested in the use case for this and the benefit within the next 6 months.
> As I've heard there are companies with very elaborate scheduling and auto-scaling that hardly had a need for long-term data.
> My instinct tells me to defer such features until its absolutely clear what's required.

Here's an example use case (in fact, this is an actual thing that we'd like to be able to do):

You would like to write an auto-idler component for Kubernetes.  This component would run at some sort of long-term regular interval (say, for instance, every night).  Then, it would, for some set of scalables (e.g. RCs, deployments, etc), check to see if there had been any activity on a particular metric (e.g. network activity or hits on an HTTP server) over the past 24h by using oldtimer.  Then, if they had no traffic, you could idle them, and set up an auto-unidler.

Without oldtimer, you *could* build this component to talk directly to one particular metrics sink, but that would mean that you could not upstream the component, since it would not work with just any Kubernetes deployment.  Alternatively, we could keep longer-term aggregated metrics in Heapster, but the Heapster vision doc *specifically* talks about moving such functionality to a different component (which I think is valuable, for reasons that I talked about a bit in the proposal).

> I'm worried that adding a "generic" read-path at this point will lead to Heapster eventually growing a full meta-QL for all its sinks. If it's trying to be the system through which writes happen, it just makes sense to read through it as well.

I really *do not* want a full query language for Heapster.  This proposal is about a minimal set of aggregations that would be useful for writing Kubernetes components that deal with historical data (e.g. the idling controller talked about above).

> Also all these sinks have different query languages. I see a lot of time being spent working on mapping features onto each other and being a general contention point. You touched that point already at the end of your document.

So, hopefully this came across in the proposal, but if not I can make it clearer there:  This should basically represent the lowest-common-denominator amongst the sinks.  There should **not** be any "this sink does not support feature X, so fake it in the Oldtimer code".  I did spend some time looking at what the different sinks are capable of, but let me know if I missed anything major.

DirectXMan12:
@fabxc you can also look at the original discussion around the Heapster long-term vision document (#769) to see some other proposed use cases for Oldtimer.

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
...
+
+In the case of aggregations, the normal `MetricsResult` and `MetricsResultList`
+are wrapped in order to differentiate between different aggregations.  Each
+metric point represents one bucket (if no buckets are requested, only one point
+is returned).  The timestamp in the case of aggregations is the timestamp of
+the start of that bucket.
+
+```go
+type MetricAggregationResult struct {
+    Average *MetricResult
```
mwielgus: Is MetricResult the best structure to pass this data?

mwielgus: Would you consider having multiple "MetricAggregationResult" and then a single data point in average, max, min, median, count etc?

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
...
+## Scaling and Performance Considerations
+
+Since Oldtimer itself does not store any data, it should be fairly easy to
+deploy multiple replicas of Oldtimer.  The metrics sinks themselves should
+already have clustering support, and thus can be scaled as well.  Since
+Oldtimer queries the metrics sinks themselves, response latency should
+depend mainly on how quickly the sinks can respond to queries.
+
+## Open Questions
+
```
mwielgus: Which sinks are you planning to support?

DirectXMan12: So, I'd definitely support Hawkular, and I'm also willing to write the initial support for other sinks that people think are important.  I've tried to make sure all the features talked about in the proposal work across InfluxDB, Hawkular, GCM, and OpenTSDB (the main metrics sinks, AFAICT).

DirectXMan12: The goal is to have a set of features that all the current metrics sinks support, and which seem likely for any future metrics sinks to support.

docs/proposals/old-timer.md
DirectXMan12: I'm definitely open to changes here.  Were you thinking something like:

```go
type MetricAggregationResultList struct {
    Items []MetricAggregationResult
}

type MetricAggregationResult struct {
    Average *MetricPoint
    Maximum *MetricPoint
    ...
}
```

That seems reasonable to me.

DirectXMan12: Whoops, that last comment needs bucket support:

```go
type MetricAggregationResultList struct {
    Items []MetricAggregationResult
}

type MetricAggregationResult struct {
    Buckets []MetricAggregationBucket
    // this below isn't strictly necessary, but puts the result into context nicely
    BucketSize time.Duration
}

type MetricAggregationBucket struct {
    Average *MetricPoint
    Maximum *MetricPoint
    ...
}
```

DirectXMan12: Alternatively, the timestamp could be moved into the `MetricAggregationBucket` type,
and have each field just be a numeric value.

mwielgus: Yeah, something like that. Although I would introduce start/end time to MetricAggregationBucket. The question is also whether you want to name all the aggregations upfront or have a map[string]MetricResult (I don't have a super strong opinon on that).

piosz:
@bryk @kubernetes/dashboard-maintainers this is something you are interested in. PTAL

bryk:
@piosz @DirectXMan12 I'll review this on behalf of Dashboard UI team.

cheld:
CC @taimir

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
+# Heapster Oldtimer
+
+## Overview
+
+Prior to the Heapster refactor, the Heapster model presented aggregations of
+metrics over certain time periods (the last hour and day).  Post-refactor, the
+concern of presenting an interface for historical metrics was to be split into
+a separate Heapster component: Oldtimer.
```
bryk: Will this be built into Heapster? Or maybe it'll be running in a separate Pod/Replication Controller/Service?

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
...
+
+Since we already persist metrics into a sink, it does not make sense for
+Heapster itself to persist long-term metrics to disk itself.  Instead, we can
+just query the sink directly.
+
+## Design
+
+### API
+
+Oldtimer will present an api somewhat similar to the normal Heapster model.
```
bryk: Can you highlight the differences? Is this the same API, but with different prefix and bucketing options? Or something else?

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
...
+different bucket of duration `B`).  `{aggregation}` may be a comma-separated
+list of aggregations to retrieve multiple at once.
+
+Where `{prefix}` is either empty (cluster-level), `/namespaces/{namespace}`
+(namespace-level), `/namespaces/{namespace}/pods/{pod-name}` (pod-level),
+`/namespaces/{namespace}/pod-list/{pod-list}` (multi-pod-level), or
+`/namespaces/{namespace}/pods/{pod-name}/containers/{container-name}`
+(container-level).
+
+In addition, when `{prefix}` is not empty, there will be a url of the form:
```
bryk: Out of curiosity, what is this for?

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
...
+
+- `ms`: milliseconds
+- `s`: seconds
+- `m`: minutes
+- `h`: hours
+- `d`: days
+
+### Functionality
+
+When Oldtimer receives a request at one of the given URLs, it will compose a
```
bryk: Does Oldtimer always talk to a persistent storage? Or, when it knows that everything is in memory already, it returns data from memory? 

Asking to asses performance for lists of pods.

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,146 @@
...
+The urls will take the forms:
+
+`/api/v1/historical/{prefix}/metrics/`: Returns a list of all available
+metrics.
+
+`/api/v1/historical{prefix}/metrics/{metric-name}?start=X&end=Y`: Returns a set
+of (Timestamp, Value) pairs for the requested {prefix}-level metric, over the
+given time range.
+
+`/api/v1/historical/{prefix}/metrics/{metric-name}/{aggregation-name}?start=X&end=Y&bucket=B`:
```
bryk: Can you formalize X, Y and B?

Whats the URL to get, say, hourly 95-percentiles for last 30 days?

bryk:
Looks nice. A few questions.

docs/proposals/old-timer.md
DirectXMan12: > Although I would introduce start/end time to MetricAggregationBucket

AFAICT, Influx doesn't actually provide an end time -- it just has a timestamp (otherwise I would have).

docs/proposals/old-timer.md
DirectXMan12: So, you'd probably want to run it separately from Heapster so that you could scale it, but we might want to also provide an easy option to run it as part of Heapster as well (some sort of "allinone" installation for quickly getting up and running).

docs/proposals/old-timer.md
DirectXMan12: Listing the available items is to be roughly in line with the Heapster model API.  In general, I think both APIs should follow the same conventions.

docs/proposals/old-timer.md
DirectXMan12: I was envisioning Oldtimer always querying the storage.  Otherwise, how will we know if "everything" is in memory?  Things could have changed since our last fetch.  The way things are tagged in the sinks shouldn't make it *too* bad to list all pods for a namespace, I would image, although I'd welcome additional data/feedback there.

docs/proposals/old-timer.md
DirectXMan12: Sure, I'll add in a brief overview of the differences.

docs/proposals/old-timer.md
mwielgus: I would suggest to include it in the default deployment. I think that a for 99% deployments a single instance will be enough. It should rather not consume lots of memory. 
I think that anyone who needs to perform large-scale data analysis will do it in some other way (not via this "poor" api).

docs/proposals/old-timer.md
DirectXMan12: They're all specified slightly below (after the definition of `{prefix}`), but if that's not clear, I can more the definitions around a bit.  An example probably wouldn't hurt either ;-)

Hourly 95-percentiles from the last 30 days: `GET /api/v1/historical/namespaces/default/pods/somepod/metrics/cpu-usage/95-perc?start=2016-03-20T10:57:37-04:00&bucket=1h`

docs/proposals/old-timer.md
DirectXMan12: Sounds good to me.

DirectXMan12:
Ok, I've tried to address all of the comments.  I've added in an example aggregation query, and made a few more things more explicit, and rearranged the proposal a bit.

docs/proposals/old-timer.md
bryk: LGTM

docs/proposals/old-timer.md
bryk: All right, sounds good.

docs/proposals/old-timer.md
bryk: > Otherwise, how will we know if "everything" is in memory?

@mwielgus does heapster have this info?

docs/proposals/old-timer.md
bryk: All right, thanks. Sounds good.

docs/proposals/old-timer.md
DirectXMan12: Yeah, if it's running in the same process with Heapster, it could just use the info from Heapster.

bryk:
LGTM from Dashboard UI perspective

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,212 @@
+# Heapster Oldtimer
+
+## Overview
+
+Prior to the Heapster refactor, the Heapster model presented aggregations of
+metrics over certain time periods (the last hour and day).  Post-refactor, the
+concern of presenting an interface for historical metrics was to be split into
+a separate Heapster component: Oldtimer.
+
+Oldtimer will run as part of the main Heapster executable, and will present
```
mwielgus: Oldtimer should rather run as a separate executable/containert in the main Heapster pod.

DirectXMan12: ah, ok, I think I misunderstood you above.

DirectXMan12: That would effectively mean that it would have to be on a different port.  Why not just allow it to run in the same executable, if you're going to be running it as part of the same pod anyway, so that `https://heapster.kube-system/api/v1/model` is for the model, while `https://heapster.kube-system/api/v1/historical` is for oldtimer?

DirectXMan12:
Had to tweak the paths a bit so that go-restful could work with them (it turns out go-restful doesn't like wildcards in the middle of a path, and it's probably not the best idea as far as ambiguity is concerned, anyway)

docs/proposals/old-timer.md
```diff
@@ -0,0 +1,212 @@
...
+    Items []MetricAggregationResult
+}
+```
+
+### Aggregations
+
+Several different aggregations will be supported.  Aggregations should be
+performed in the metrics sink.  If more aggregations later become supported
+across all metrics sinks, the list can be expanded (and the API version
+should probably be bumped, since the supported aggregations should be part of
```
piosz: Please note that we actually don't have versioned API in Heapster. This a kind of promise that we will try to support it in some backward-compatible way but there is no guarantee. If you really want to have a versioned API it should be a part of Resource Metrics API https://github.com/kubernetes/kubernetes/pull/24253

DirectXMan12: I'll tweak the wording here.  I figured since there was an API version in the path, the intention was that you'd have a versioned API at some point.

piosz: There will be another versioned API of Kubernetes standards (proposal linked in previous comment). Especially if you want model/historical api to be fully versioned they should go through the path: alpha -> beta -> stable. We don't want it here to allow higher velocity at cost of possible instability.

piosz:
Let's imagine the situation where we created a pod with some name, it was deleted and then let's say a week ago we created another pod with the same name which is a totally different pod. What will be the semantic of historical metrics in this case? Do you plan to verify also pod-uid somehow?

DirectXMan12:
> What will be the semantic of historical metrics in this case? Do you plan to verify also pod-uid somehow?

Good point.  We probably want `/pod-ids/{pod_id}` and `/pod-ids/{pod_id}/containers/{container}` API endpoints.  Keeping the namespace-pod API endpoint as well might be nice, though, since it mirrors the model API.

piosz:
But operating on pod uids is a bad user experience.

I'm ok with saying (possibly in the first version): if you have two pods with the same name historically the data will be mixed. Better approach is to return data only from the newest pod with the given name. For example I created on Monday pod with name `my-name` and then killed it on Tuesday. I created also a pod with the same name on Friday and it is still running. When querying for historical data for the whole week I'd like to get only metrics for the second pod, so there won't be any data for Mon-Thu. WDYT?

DirectXMan12:
> Better approach is to return data only from the newest pod with the given name

This gets a bit complicated, since effectively we have to then know the appropriate pod uid for a pod name *before* making a query, so we have to make two queries against the database.  That might not be so bad, but it does adds another round trip for every query.

If we're ok with the extra round trip, the optimal setup would probably be that you can refer to pod UID to get a specific pod, or refer to pod name to get the newest pod with that name.

DirectXMan12:
I've added in the pod-id-based endpoints, and added a note about the pod-name-based endpoints.  I think there are some cases for certain backends in which we can get away with only one trip to the backend, but there are several cases where two will be required.

For instance, to get raw metrics from Hawkular, you *need* to know the pod id, AFAICT (@mwringe et all might be able to correct me on this).  For aggregations, I'm not sure how Hawkular deals with a query which could return multiple time series, but I suspect we'll need it there as well.

DirectXMan12:
I've also clarified a couple points for posterity's sake:

- Oldtimer is intended to work only with data written in the Heapster storage schema (i.e. data originally written by Heapster)
- Oldtimer is not mean to be a general-purpose query tool that maps arbitrary queries to the sink query language -- it's just supposed to support a basic set of aggregations on single time series.

cc @fabxc 

I think at this point the proposal is pretty much ready for merge.  @mwielgus anything else you want changed?

mwielgus:
LGTM

mwringe:
> For instance, to get raw metrics from Hawkular, you need to know the pod id, AFAICT (@mwringe et all might be able to correct me on this). For aggregations, I'm not sure how Hawkular deals with a query which could return multiple time series, but I suspect we'll need it there as well.

Hmm, you shouldn't need to know the pod id to get the raw data, you should be able to do a query based on any of the labels. But there may have been an issue preventing this in the past.

Can you get in touch with Stefan Negrea (he is stefan_n in #hawkular on freenode) to answer your questions?

PR #1129, opened 2016-04-14
Proposal: Push Metrics

DirectXMan12:
This is a proposal to support a push model (in addition to the current
pull model) for metrics in Heapster.  It is targeted at pushing custom
metrics, and supplements the metrics collected by the pull model.

### DIFF
--- a/docs/proposals/push-metrics.md
+++ b/docs/proposals/push-metrics.md
@@ -1,0 +1,233 @@
+# Heapster Push Metrics
+
+## Overview and Motivation
+
+Currently, Heapster supports pulling metrics from kubelet, and defines an
+interface for pulling from other sources.  However, in certain cases, it is
+more useful to be able to have services push metrics into Heapster, instead
+of having Heapster pull metrics.
+
+For instance, supporting push metrics makes it easy for cluster admins to add
+custom metrics into Heapster with relatively minimal effort: they can simply
+write a program or script which collects and processes the information, and
+then add a recurring Job or cron job which pushes the metrics.  This also
+enables existing tooling designed with the push model in mind to be adapted to
+provide metrics through Heapster.
+
+### Target Audience and Metrics
+
+Like the existing custom metrics pull mechanism, the metrics pushed through the
+push mechanism are intended to be those metrics that are useful for consumption
+by system components, such as metrics intended for use with autoscaling.  The
+custom metrics pushed through the push mechanisms are still intended to follow
+the overall guidelines for Heapster custom metrics (keep the number of
+different metric names relatively limited, etc).
+
+The current custom metrics mechanism can only be used to collect custom metrics
+which describe the producing pod.  This proposal is designed to provide a
+method of collecting metrics which describe multiple resources (other pods,
+services, etc) across the cluster.  Such producers are generally add-on cluster
+infrastructure components deployed by the cluster admins.  The push mechanisms
+are not, in general, intended for use by arbitrary cluster users (although the
+metrics transfered would probably apply to the users' pods).  It is up to
+cluster admins to decide which applications are permitted to use the push
+mechanims (see the "Authentication" section below for more information on how
+cluster admins can control access).
+
+For these reasons, the push metrics mechanism is not designed to replace the
+existing pull mechanism for custom metrics, but instead to support additional
+use cases not already supported.
+
+## API
+
+### Authentication, Segregation, and Flow Control
+
+Producers will be authenticated similarly to the way Heapster currently
+authenticates clients: producers will present a certificate signed by a CA, and
+Heapster will be configurable to only allow certain names to push metrics, or
+to allow any certificate signed by the CA.
+
+Additionally, the name presented during authentication will be used as a prefix
+for all metrics added.  This will prevent two different metrics producers from
+accidentally overwriting each other's custom metrics (otherwise, push metrics
+will be stored and retrieved identically to pull-based custom metrics).
+
+In order to prevent push metrics from overwhelming the Heapster instance, it
+will be possible to limit the total number of custom metrics each producer is
+allowed to add, and the frequency at which producers are allowed to push new
+sets of custom metrics.  By default, no limits will be enforced unless
+explicitly set via command line arguments.
+
+### Paths
+
+To add metrics, metrics producers will `POST` new metrics to
+`/api/v1/push/{format}/{subpath}` in the format specified by `{format}`.  The
+metrics are pushed in bulk -- it is up to the format to support a way to name
+specific metrics, namespaces, pods, and containers (for instance, the
+Prometheus format uses metric names and labels for this purpose).  The
+`{subpath}` option enables different formats to have format specific sub-paths.
+
+### Metrics Format
+
+The underlying design can support multiple format "backends".  The initial
+backend, detailed here, will be based on the Prometheus format, and will be
+available at `/api/v1/push/prometheus/`,
+`/api/v1/push/prometheus/metrics/job/{producer_name}`, or
+`/api/v1/push/promethus/metrics/jobs/{producer_name}`.  If either of the latter
+two paths are used, `{producer_name}` should match the name used when
+authenticating with Heapster.
+
+Both the Prometheus text and protobuf formats will be supported.
+
+The metric name specified on each metric line will be used as the custom metric
+name in Heapster (except prefixed as discussed in the "Authentication"
+section).  The following labels will be used to determine which object a metric
+is associated with:
+
+- For kubernetes-related metrics, the `namespace` label will indicate
+  namespace, the `pod` label will indicate pod, and the `container` label will
+  indicate container.  If only `namespace` is present, then the metric will be
+  considered namespace-level.  If `namespace` and `pod` are present, then the
+  metric will be considered pod-level.  If all three are present, the metric
+  will be considered container-level.
+
+- For kubernetes-related metrics, the `service` label can be used in
+  conjunction with the `namespace` label to indicate a service-level metric.
+  Currently, this will be stored in Heapster as a namespace-level metric
+  prefixed with the service name.
+
+- For non-kubernetes-related metrics, the `node` label will indicate node, and
+  the `container` label will indicate free container.   If only `node` is
+  present, the metric will be considered node-level.  Otherwise, the metric
+  will be considered free-container-level.
+
+If the `node`, `namespace`, `pod`, and `container`, and `service` labels are
+not present in one of the configurations listed above, the metric line is
+invalid and the batch should be rejected.
+
+Any additional labels will be treated the same as labels on existing custom
+metrics (currently multiple custom metrics with the same name, but different
+labels, are ignored, but this seems like an oversight and should probably be
+fixed).
+
+Timestamps will be assigned by using the next Heapster metrics batch timestamp
+after the time at which the metrics are received.  If a timestamp is provided
+as part of the metric line, this may be stored as a separate field for
+posterity, but the "official" timestamp will be that of the assigned batch.
+
+As with normal Prometheus metrics, the `TYPE` line should be used to provide
+the type of the metric.
+
+#### Example
+
+Suppose a producer with the name "http_gatherer" sent the following metrics to
+`/api/v1/push/prometheus`:
+
+```
+# This is a pod-level metric (it might be used for autoscaling)
+# TYPE http_requests_per_second guage
+http_requests_per_minute{namespace="webapp",pod="frontend-server-a-1"} 20
+http_requests_per_minute{namespace="webapp",pod="frontend-server-a-2"} 5
+http_requests_per_minute{namespace="webapp",pod="frontend-server-b-1"} 25
+
+# This is a service-level metric, which will be stored as frontend_hits_total
+# and restapi_hits_total (these might be used for auto-idling)
+# TYPE hits_total counter
+hits_total{namespace="webapp",service="frontend"} 5000
+hits_total{namespace="webapp",service="restapi"} 6000
+```
+
+This would result in the metrics being available at:
+
+```
+/api/v1/model/namespaces/webapp/pods/frontend-server-a-1/metrics/custom/http_gatherer/http_requests_per_minute
+/api/v1/model/namespaces/webapp/pods/frontend-server-a-2/metrics/custom/http_gatherer/http_requests_per_minute
+/api/v1/model/namespaces/webapp/pods/frontend-server-b-1/metrics/custom/http_gatherer/http_requests_per_minute
+/api/v1/model/namespaces/webapp/metrics/custom/http_gatherer/frontend/hits_total
+/api/v1/model/namespaces/webapp/metrics/custom/http_gatherer/restapi/hits_total
+```
+
+## Discussed Alternatives
+
+A number of alternatives came up during the discussion of this proposal.  They
+are discussed briefly below.  Note that most of these alternatives do not deal
+particularly well with a case where metrics need to come from a source that is
+not running as a pod on the cluster.  While it is expected that many of the
+producers will be running as components on the cluster (e.g. as DaemonSets or
+PetSets), it could still be adventageous to support metrics coming from
+components that are not in the form of pods.
+
+### Writing directly into sinks
+
+This alternative would have producers write directly into sinks in the Heapster
+storage schema, and then use a mechanism similar to the Oldtimer API to read
+the metrics back.
+
+This would require every producer to know how to talk to every sink, would make
+configuring the sinks more complicated, and would most likely lead to software
+only being able to talk to one of the sinks supported by Heapster.
+Additionally, you lose the benefits of the Heapster model, and either have to
+adapt the existing Heapster model to fall back to an Oldtimer-like approach, or
+teach all cluster components to be able to read from both the Heapster model
+and Oldtimer simultaneously.
+
+### Reworking the existing cAdvisor-Kubelet-Heapster Pull Mechanism
+
+This alternative would involve reworking the existing pull mechanism to allow
+certain pods to produce metrics that describe other resources besides the
+themselves, as opposed to the current situation, where all custom metrics
+collected via the current pull mechanism are marked as describing the producer
+pod.
+
+This would require a mechanism for indicating to Heapster which pod names were
+allowed to produce metrics that describe other resources, since admins would
+generally want most pods producing metrics to continue to just have metrics
+which describe only the producer pod.  It would also conceptually blend
+together pods producing metrics about themselves versus pods producing metrics
+about others.  Additionally, the current cAdvisor-based custom metrics
+collection is not secured, so all metrics would be available to anyone with
+knowlege of the appropriate port, but this may change in the future.
+
+### Using a new daemon per node to produce metrics
+
+This alternative would involve running a daemon on each node that aggregated
+all the separate custom metrics producers' results together.
+
+It was suggested that an approach similar to the Prometheus Node Exporter
+Textfile Collector could be used, in which sources would write their metrics to
+files in a directly, which would later be read by the collector when polled for
+metrics.  When the producers are containerized, you'd need to use a hostPath
+volume, have the daeamon look for specific emptyDir mounts in containers (and
+use one director per container), or something similar.
+
+Alternatively, a new daemon could be run on each node that was responsible for
+collecting metrics from producers who produce bulk metrics describing other
+resources.
+
+This would still require some sort of auth to limit which pods where allowed to
+do so (while the scoping above prevents collision, cluster admins would most
+likely still want to limit which pods are allowed to post metrics which appear
+in another pod's list of custom metrics).  Unlike the proposal above, admins
+could not simply rely on "whoever is allowed to authenticate" rule, since
+cAdvisor does not check certificates like the normal Heapster auth mechanism.
+
+Additionally, this adds a bit of complexity on the producer's side, since it
+requires continuously serving the metrics (this could be made easier by
+providing a tool like the Prometheus Node Exporter Textfile Collector, which
+just serves up metrics based on text files in a directory).
+
+### Adding an additional standard pull mechanism
+
+This alternative would involve writing a pull mechanim which, for instance, was
+just able to read Prometheus metrics directly.  This would either require
+admins to configure the Heapster instance to know about every custom metrics
+source (and restart Heapster when a new source needed to be added), or would
+require teaching Heapster how to look for an annotation on certain pods to
+determine which pods to query (Heapster would have a list-watch on the pods,
+and look for pods added/removed/changed with the appropriate annotation).
+
+When used in the latter form, this mechanism would still likely require a
+similar auth setup to the one proposed above, in order to allow the admin to
+restrict which pods actually were allowed to produce the metrics.  It also has
+similar restrictions/disadvantages as the "Using a new daemon per node" method
+discussed above.


DirectXMan12:
cc @piosz @mwielgus @fgrzadkowski

piosz:
cc @dchen1107

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+metrics (currently multiple custom metrics with the same name, but different
+labels, are ignored, but this seems like an oversight and should probably be
+fixed).
+
+Metric lines must have timestamps, which will be used as the timestamp
+associated with the metric in Heapster (similarly to how pull metrics use the
+timestamp provided by Kubelet).  As normal, this requires the producer's clock
+to be in sync to yield reasonable results.
+
+## Open Questions
```
vishh: What about flow control, resource isolation across sources?

piosz:
cc @jszczepkowski

docs/proposals/push-metrics.md
DirectXMan12: Metrics pushers can't overwrite/interfere with each other, since each metric is stored as prefixed by pusher's name, as determined during the auth step, unless you had multiple sources that shared the same name but were not otherwise coordinated.  That scenario seems like something that should just be avoided in this case.  I can make that explicit if need be. 

In general, I imagined that the use cases for this are more designed for either cluster-admin-created components, or admin-blessed components, and thus some of this would be a case of "does the admin trust themselves".

Nonetheless, I could be desirable to limit metrics per pusher, pushes per second, etcd.  It would be fairly easy to do so on a global scale, but I could see a case for having per-pusher controls on those variables.  The question becomes how complicated we want to make this.

DirectXMan12:
P.S. I'm open to changing the proposed initially supported format.  I chose the prometheus one since it's probably familiar to people, and it's pretty easy to write from just about any language (including scripts, etc).

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+  the `container` label will indicate free container.   If only `node` is
+  present, the metric will be considered node-level.  Otherwise, the metric
+  will be considered free-container-level.
+
+Any additional labels will be treated the same as labels on existing custom
+metrics (currently multiple custom metrics with the same name, but different
+labels, are ignored, but this seems like an oversight and should probably be
+fixed).
+
+Metric lines must have timestamps, which will be used as the timestamp
```
brian-brazil: Prometheus developer here.

None of the Prometheus client libraries support timestamps and we don't plan on changing that until https://github.com/prometheus/prometheus/issues/398 is resolved. There is no timeframe on that currently.

Instead you should do as our clients do and not specify the timestamp, and leave Heapster to use the current time. This also avoids clock sync issues.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+producers will present a certificate signed by a CA, and Heapster will be
+configurable to only allow certain names to push metrics.
+
+Pushed metrics will then be additionally scoped by producer name, to avoid
+conflicts.
+
+### Metrics Format
+
+The underlying design can support multiple format "backends".  Initially, only
+the prometheus text format will be supported at `/api/v1/push/prometheus`.  The
```
brian-brazil: This path does not follow the Prometheus Pushgateway API, and thus a Prometheus client library can't be used to push to this. If you want to offer this API then it should end with `/metrics/job/jobvalue/label1/value1/label2/value2` etc. which will be supported by the various existing Prometheus client libraries.

The Go client will push protobuf rather than text, so you should choose the parser based on the Content-Type header. The Go client has the required APIs to handle this for you.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+
+Currently, Heapster supports pulling metrics from kubelet, and defines an
+interface for pulling from other sources.  However, in certain cases, it is
+more useful to be able to have services push metrics into Heapster, instead
+of having Heapster pull metrics.
+
+For instance, supporting push metrics makes it easy for cluster admins to add
+custom metrics into Heapster with relatively minimal effort: they can simply
+write a program or script which collects and processes the information, and
+then add a recurring Job or cron job which pushes the metrics (instead of
```
brian-brazil: You may wish to take a look at how the Prometheus Node Exporter Textfile Collector works as an alternate implementation option. It's how we handle machine level metrics that aren't produced at every scrape, which seems to be the scenario here and has worked quite well for us.
Basically it's a directory you can dump prometheus text format files into, and they'll be included in `/metrics` output.

In this case the kubelet could pull in files on scrape, and include them with the rest of it's metrics. Depending on directory layout, file permissions and label handling this could remove the need to do explicit auth, users having to figure out which container/pod label to apply and opens options for non-kubernetes systems on the machine to produce metrics.

For things you need to set only once this would avoid the need to setup a cronjob or depend on the scheduler to run one (and be in a sufficiently working state to be able to schedule). It also avoids the need for rate limiting, as it's only a write to local disk rather than hitting heapster directly.

fgrzadkowski:
Since @mwielgus is busy with designing cluster autoscaling, reassigning to @piosz

fabxc:
From my understanding _the_ purpose of Heapster is to hold metrics that are relevant for autoscaling and scheduling.

This is a fairly recent document stressing that this is the main purpose:
https://github.com/kubernetes/heapster/blob/master/docs/proposals/vision.md#custom-metrics-status

I'm confused why Heapster needs a feature to push service level metrics through its pipeline if it's a) not its purpose and b) not willing to give any guarantees around it or put effort into it.
There are a variety of projects working hard to apply sane semantics and decent APIs for metrics collection – be that push or pull.

For the system-level metrics needed for autoscaling and scheduling it seems very feasible to just pick a single format and either push or pull.
The specified performance target will also certainly suffer from people routing all the app-level metrics (which are the majority) through Heapster (for no reason).

This proposal seems out of scope to me.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+Metrics added through push metrics will be available as normal through the
+Heapster model.
+
+To add metrics, metrics producers will `POST` new metrics to
+`/api/v1/push/{format}` in the format specified by `{format}`.  Producers will
+be authenticated similarly to the way Heapster currently authenticates clients:
+producers will present a certificate signed by a CA, and Heapster will be
+configurable to only allow certain names to push metrics.
+
+Pushed metrics will then be additionally scoped by producer name, to avoid
```
mwielgus: What do you exactly mean by additional scoping? Do you want to include CA owner in the metric name?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Design
+
+### API
+
+Metrics added through push metrics will be available as normal through the
+Heapster model.
+
+To add metrics, metrics producers will `POST` new metrics to
+`/api/v1/push/{format}` in the format specified by `{format}`.  Producers will
```
mwielgus: How will you exactly attach metrics to pod/containers/namespaces/etc? /api/v1/push seeems to look like a generic bulk push.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+### Metrics Format
+
+The underlying design can support multiple format "backends".  Initially, only
+the prometheus text format will be supported at `/api/v1/push/prometheus`.  The
+metric name in the prometheus metrics will be mapped to a custom metric with
+the name `{client_name}/{metric_name}` in Heapster (with the client name coming
+being the same as used in authentication), and the following labels will be
+used to determine which object a metric is associated with:
+
+- For kubernetes-related metrics, the `namespace` label will indicate
```
mwielgus: Could you please include an example of how the push content should look like?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,69 @@
...
+
+Any additional labels will be treated the same as labels on existing custom
+metrics (currently multiple custom metrics with the same name, but different
+labels, are ignored, but this seems like an oversight and should probably be
+fixed).
+
+Metric lines must have timestamps, which will be used as the timestamp
+associated with the metric in Heapster (similarly to how pull metrics use the
+timestamp provided by Kubelet).  As normal, this requires the producer's clock
+to be in sync to yield reasonable results.
```
mwielgus: The current heapster works on batches. Every 30/60sec it scrapes all the metrics from the nodes packs into a "batch" and pushes together down the processing stream. In the batch all metrics  have the same timestamp which is rounded to 30/60sec (although there is also information about the exact scrape time). If you push them often enough (>>=2x heapster scrape time) then this should not matter too much.

docs/proposals/push-metrics.md
DirectXMan12: So, AFAICT, the pull metrics on Heapster use the time provided by the kubelet APIs (one of the Heapster core devs can feel free to correct me here :wink:).  I don't really want to have one part of Heapster using time from the metrics source, and another part using time from Heapster.

mwielgus: Eventually we will use a "rounded" timestamp for the metrics. So for example if you push your metrics at 12:01:40 we will grab it and use in a batch with timestamp 12:02:00 along with metrics scraped by Kubelet at 12:01:51. The original timestamp can also be available somewhere in the data for some custom processing (like converting cumulative to gauge) but the idea in Heapster is that all data/metrics from one batch share the same timestamp for consistency/simplicity.

DirectXMan12: @mwielgus ok, sounds good.  I'll remove the timestamp bit then.  Thanks!

docs/proposals/push-metrics.md
DirectXMan12: Sure, I'll add in an example.

docs/proposals/push-metrics.md
DirectXMan12: Yes, it's a bulk push.  Labels are used to indicate which metrics go where.  I talk about that a bit below.  I can clarify a bit, though.

docs/proposals/push-metrics.md
DirectXMan12: So, while I mainly chose the prometheus format since it's really pretty easy to write from just about any language (including shell scripts), I do *like* the idea of actually supporting the Prometheus push gateway URL as well.  My main issue is that the URL format (`/metrics/job/{jobname}`) doesn't really fit with the rest of the Heapster API (which looks like `/api/v1/model` and so on).  I would not be averse to the Prometheus format handler accepting something like `/api/v1/push/metrics/job/{jobname}`, but I'm not sure if the Prometheus clients would be able to handle it (and thus if it would be worth trying to implement).

docs/proposals/push-metrics.md
DirectXMan12: Kind of -- I don't want anyone to be able to accidentally (or purposefully, for that matter) overwrite someone else's metrics by using the same name.  The current auth framework extracts the common name from the client cert to check against a list of allowed users.  This proposal would have that name be passed on, and all metrics submitted would be prefixed with that.

DirectXMan12:
@fabxc 

Just to clarify, this proposal *is not* about pushing any and all metrics through Heapster.

This proposal *is* about giving cluster admins an easy and flexible path to push additional metrics that they would like to use to for autoscaling (and perhaps even other cluster components) at various levels of the hierarchy (whether it be pushing some sort of special aggregate metric composed of many smaller metrics at a per-pod level, or being able to push namespace-level metrics, and then consume those from cluster components and potentially autoscaling in the future).

> From my understanding the purpose of Heapster is to hold metrics that are relevant for autoscaling and scheduling.

Yes, we are in agreement here.

> This is a fairly recent document stressing that this is the main purpose:
https://github.com/kubernetes/heapster/blob/master/docs/proposals/vision.md#custom-metrics-status

> I'm confused why Heapster needs a feature to push service level metrics through its pipeline if it's a) not its purpose

This is where we start to disagree.  I would argue that custom metrics *are* important for autoscaling.  The above cited document says "The support for custom metrics is focused on auto-scaling and critical functionality monitoring (and potentially scheduling)".  This proposal is still targeting those metrics that would be used for autoscaling.

Additionally, there's nothing that says that custom metrics here are necessarily application-level metrics.  Sure, current custom metrics are application-level, but this proposal makes it easy for cluster admins to add in system metrics specific to their cluster.

> and b) not willing to give any guarantees around it or put effort into it.

I'd like for @mwielgus to chime in here, but my understanding is less that we're not willing to put any effort into it, and more that we're not going to optimize for having 100s of different custom metrics.

> There are a variety of projects working hard to apply sane semantics and decent APIs for metrics collection – be that push or pull.

I don't think that we want/need another component running on top of the cluster that is required for the cluster to function.  If it's metrics-related and a cluster component is going to make use of it (e.g. the HPA controller), it should go through Heapster or a Heapster-related component.

> For the system-level metrics needed for autoscaling and scheduling it seems very feasible to just pick a single format and either push or pull.

> The specified performance target will also certainly suffer from people routing all the app-level metrics (which are the majority) through Heapster (for no reason).

Just to reiterate, I'm not suggesting that all application-level metrics go though Heapster.  The stated goal of custom metrics support has been to support only those custom metrics needed for autoscaling and related things.  This proposal does not seek to change that.

vishh:
@DirectXMan12 

> Just to reiterate, I'm not suggesting that all application-level metrics go though Heapster. The stated goal of custom metrics support has been to support only those custom metrics needed for autoscaling and related things. This proposal does not seek to change that.

How do you envision enforcing this logic? What would prevent users from exposing monitoring specific metrics via this model?

DirectXMan12:
@vissh so, there's really no way to distinguish between the two types of metrics short of asking the producer.  To be fair, with the current pull model custom metrics, there's also technically no way to tell if the user is exposing monitoring-specific metrics, either.

However, we could probably add options for restricting metrics on a per-producer level fairly easily ("which custom metrics are we currently keeping track of with the prefix '$PRODUCER', compare that to the new names that would be added here, check to make sure the end result wouldn't be more than 5"), if the cluster admin want to hand out push certificates without having to worry about someone pushing in way to many metrics.

fabxc:
I'm sure you have the right intentions here. But people will use what is possible and if enough people use it, eventually development will cater to these uses.

I'd consider the file pattern @brian-brazil described, which keeps things pull-based.

Taking this from another angle:
Making a clear separation between your regular monitoring and your metrics for auto-scaling/scheduling to me means that they should not end up in the same storage.
The overhead of storing them twice (given that you'll probably want them in your normal monitoring) would be fairly negligible considering your total amount of monitoring data.

When nothing but Heapster has to read or write these metrics, there suddenly seems to be little reason to not have one backend. The whole concern of different semantics, different feature-sets, and storage guarantees falls away.
Only concern left would be deployment of the backend. But if it's just part of Heapster, this seems like not much of an issue.

vishh:
> Taking this from another angle:
Making a clear separation between your regular monitoring and your metrics
for auto-scaling/scheduling to me means that they should not end up in the
same storage.

This might not be a valid assumption. AFAIK, there are use cases where
service metrics for auto-scaling can be surfaced via metrics monitoring
systems - sysdig, GCM, etc.

mwielgus:
Regarding 100s of custom metrics - this was never the target and most likely won't be. But we are trying follow Kubernetes scalability goals in Heapster, which for 1.2 were 1000 nodes x 30 pods in reasonable CPU/mem. So if your cluster is smaller then you have enough capacity to move some custom metrics around.

docs/proposals/push-metrics.md
DirectXMan12: It looks like the Go client and the Python client can somewhat accidentally support a having `/metrics/job/{jobname}` at a subpath, while the Python client, for instance, does not really support this (the gateway argument is only used for it's host and port, not the path).

Since I'd really like to preserve the base Heapster API path format (`/api/v{version}/{api_name}/...`), I'm not sure it's worth it to try to support the Prometheus push gateway API.  @brian-brazil do you have any thoughts on this?

DirectXMan12:
I've added in an example, reorganized a bit, and made a couple things a bit clearer.

fabxc:
I just want to make clear that I don't have much background in auto-scaling. I'm poking in all these directions because the goals and scope separations seem very blurry to me and I want to better understand the desired and existing separation.

Initially I understood Heapster collected metrics itself because the feedback loop through a monitoring system/TSDB might be too long. That would be an understandable point but it doesn't seem to be the case. Elsewhere 30/60s rounding is mentioned and apparently we want to also read metrics not ingested via Heapster from the monitoring system/TSDB directly.
Naturally, I also want system level metrics in my monitoring system. So why retrieve them via two different system, i.e. why does Heapster need a write-path at all? Making the distinction to not collect the same metrics twice introduces a lot of operational complexity.

Not only are there scope conflicts with people's regular monitoring systems but even directly with cAdvisor, which I've heard also wants to handle application metrics now (different discussion).

My concerns about different semantics of underlying storage still hold.

Something I do know about auto-scaling that it's incredibly hard to get right and every additional complexity in it might may cause unexpected feedback loops. That makes me question whether enabling more and more things with unclear semantics in quick succession is beneficial for the greater cause.

This is a more fundamental discussion and I'm happy to take it elsewhere.

docs/proposals/push-metrics.md
jimmidyson: This seems like a nice approach to me. It would mean no change to the current flow of data into Heapster, no requirement for authentication changes, no flow control to prevent overloading Heapster.

jimmidyson:
@DirectXMan12 @mwielgus @fabxc Perhaps we could have an open hangout (or whatever) to discuss?

DirectXMan12:
@jimmidyson we can discuss at the upcoming SIG Autoscaling meeting, if that works for you and fabxc, or we can schedule a hangout for some other time next week

jimmidyson:
@DirectXMan12 SGTM - where can I find details on that?

fabxc:
Sounds good depending on the time :) That's certainly a better format.

jimmidyson:
@fabxc From https://docs.google.com/document/d/1RvhQAEIrVLHbyNnuaT99-6u9ZUMp7BfkPupT2LAZK7w/edit#heading=h.oh2koj9sbr3x:

> We have bi-weekly meetings, every second Thursday at 11:30 US East Time, starting Oct 15.

I assume that's still correct?

DirectXMan12:
@jimmidyson @fabxc yeah, the next one is this upcoming Thursday (a week from today).

fabxc:
I'll probably still be traveling by that time unfortunately.

DirectXMan12:
I've clarified the "target audience" a bit (the proposal is geared towards cluster components, not arbitrary user applications, and is not intended to replace the existing custom metrics flow for user apps), and also added in a discussion of some of the proposed alternatives, and what I see to be their disadvantages.

@fabxc @jimmidyson please let me know if I've misrepresented any of the alternatives that you proposed.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+provide metrics through Heapster.
+
+### Target Audience and Metrics
+
+Like the existing custom metrics pull mechanism, the metrics pushed through the
+push mechanism are intended to be those metrics that are useful for system
+components, such as metrics inteded for use with autoscaling.  The custom
+metrics pushed through the push mechanisms are still intended to follow the
+overall guidelines for Heapster custom metrics (keep the number of different
+metric names relatively limitted, etc).
```
brian-brazil: limited

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+
+It was suggested that an approach similar to the Prometheus Node Exporter
+Textfile Collector could be used, in which sources would write their metrics to
+files in a directly, which would later be read by the collector when polled for
+metrics.  Using a shared directory like this poses issues when you want to run
+the producers in containers.  Instead, you could instead have producers push to
+the daemon via HTTP, but you then effectively have the push metrics mechanism,
+except running on every node instead of just running as part of Heapster.
+
+Alternatively, a new daemon could be run on each node that was responsible for
```
brian-brazil: My proposal was to use the kubelet for this, so there's no new daemon.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+collecting cluster-scoped metrics via pull.  This would still require some sort
+of auth to limit which pods where allowed to produce cluster-scoped metrics.
+Additionally, it makes adding metrics at a less-frequent rate more complicated
```
brian-brazil: I fail to see how this is the case, you just write them less often - it's the same as what you're proposing in this regard.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+### Using a new daemon per node to produce metrics
+
+This alternative would involve running a daemon on each node that aggregated
+all the separate custom metrics producers' results together.
+
+It was suggested that an approach similar to the Prometheus Node Exporter
+Textfile Collector could be used, in which sources would write their metrics to
+files in a directly, which would later be read by the collector when polled for
+metrics.  Using a shared directory like this poses issues when you want to run
+the producers in containers.  Instead, you could instead have producers push to
```
brian-brazil: I'd presume that you'd be bind mounting this, or creating such a directory per container.

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+
+#### Example
+
+Suppose a producer with the name "http_gatherer" sent the following metrics to
+`/api/v1/push/prometheus`:
+
+```
+# This is a pod-level metric (it might be used for autoscaling)
+# TYPE http_requests_per_second guage
+http_requests_per_minute{namespace="webapp",pod="frontend-server-a-1"} 20
```
brian-brazil: Above you said that this was for cluster-level metrics, however this is an application level metric. Can you give an example of cluster-level metric that you'd want to push?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Reworking the existing cAdvisor-Kubelet-Heapster Pull Mechanism
+
+This alternative would involve reworking the existing pull mechanism to allow
+certain pods to produce the "cluster-scoped" metrics (i.e. metrics scope to
+pods other than the producer), as opposed to the current situation, where all
+custom metrics collected via the current pull mechanism are scoped to the
+producer pod.
+
+This would require a mechanism for indicating to Heapster which pod names were
```
brian-brazil: Would the prefixing approach you're proposing in the main solution not also handle this?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+allowed to produce cluster-scoped metrics, since admins would generally want
+most pods producing metrics to continue to just have producer-pod-scoped
```
brian-brazil: Above you indicated that this design is for cluster-scoped metrics, why is this talking mostly about pod scoped metrics?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+The metric name specified on each metric line will be used as the custom metric
+name in Heapster (except prefixed as discussed in the "Authentication"
+section).  The following labels will be used to determine which object a metric
+is associated with:
+
+- For kubernetes-related metrics, the `namespace` label will indicate
+  namespace, the `pod` label will indicate pod, and the `container` label will
+  indicate container.  If only `namespace` is present, then the metric will be
+  considered namespace-level.  If `namespace` and `pod` are present, then the
+  metric will be considered pod-level.  If all three are present, the metric
```
brian-brazil: If this is for cluster-level metrics, why are you allowing for pod and container labels to be set?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+uses metric names and labels for this purpose).
+
+### Metrics Format
+
+The underlying design can support multiple format "backends".  The initial
+backend, detailed here, will be based on the Prometheus format, and will be
+available at `/api/v1/push/prometheus`.
+
+Initially, only the Prometheus text format will be supported. In the future, we
+may also wish to support the promethus protobuf format as well.
```
brian-brazil: As I've indicated before what you're proposing here is not compatible with the Prometheus ecosystem. You're effectively asking users to hand-produce the Prometheus text format. I don't think is a good idea for either the Kubernetes or Prometheus ecosystems, as it will frustrate everyone involved.

If you want to use our format you should be compatible with our Pushgateways URL pattern, and support both text and protobuf formats.

brian-brazil:
Reading this I'm very unclear what the goal of this is.

The target audience section says this is for cluster-level metrics, and is not for arbitrary cluster users. The rest of the doc, arguments and examples are however for custom pod-level application/user metrics.

I think we need to step back and see what the real world problem is that we're trying to solve here, and then see what the best approach is. This currently looks like putting Heapster on the path to a fully-general custom metrics solution, which at first glance doesn't seem like a proportionate solution to autoscaling and not something you'd want on a critical control path like that.

My straw man proposal would be to have the autoscaler directly scrape what it needs, as the request rate shown in the example is something that the pod will already be exposing in some fashion. This would give us an idea of what the real world usability/auth/maintainability/performance problems are, and then we can design solutions for them rather than prematurely optimising.

docs/proposals/push-metrics.md
DirectXMan12: :heavy_check_mark:

docs/proposals/push-metrics.md
DirectXMan12: As far as I understand it, the node team did not really want Kubelet in charge of collecting metrics.

brian-brazil: That I can appreciate, though it would be appropriate for node-level metrics.

docs/proposals/push-metrics.md
DirectXMan12: Presumably, even with the prefixing, the admin probably still does not want to allow such behavior globally, even if it's just for cosmetic reasons.

brian-brazil: That's the same as the main proposal then. If that's a disadvantage of this approach, then it's also a disadvantage of that.

docs/proposals/push-metrics.md
DirectXMan12: As I noted above in response to your earlier note, I would *like* to support the Prometheus Pushgateway's URL scheme (and I wouldn't mind supporting protobuf either, I just wanted to limit the initial scope).  *However*, I do not think it's a good idea to have one URL that looks completely different from the rest of Heapster's APIs, so it would have to be something like `/api/v1/push/prometheus/metrics/jobs/...`, and AFAICT, a couple of the Prometheus clients seem to support a URL prefix, while others do not (IIRC the Ruby client looked like it did not), and the specification in the Pushgateway repo does not seem to allow it either.

Ergo, I wasn't sure if it was worth it to include it at a subpath, so I left it out barring further comments on the matter (see https://github.com/kubernetes/heapster/pull/1129#discussion_r60298969).

I'm certainly willing to add it in as a subpath, and I'm also willing to change the format (for instance, we could do a JSON format instead).

docs/proposals/push-metrics.md
DirectXMan12: The very next set of metrics is an example of a service-level metric.  I said above that this is for *cluster components* -- sources producing metrics tied to other pods/services/namespaces/etc.

Currently, we have a method for individual pods to expose information about *themselves* (the cAdvisor-Kubelet-Heapster-Pull chain) This proposal would allow exposing metrics about other entities (pods, services, namespaces, etc) in the cluster.

docs/proposals/push-metrics.md
DirectXMan12: When I mentioned "cluster-scoped" above, I mean metrics not describing the producer itself (as I note on line 171).  Those could be pod-level, node-level, etc.

docs/proposals/push-metrics.md
DirectXMan12: I meant more of "you don't just have the option to push whenever you're ready, instead, you have to have a daemon always running".  I'll reword a bit.

docs/proposals/push-metrics.md
DirectXMan12: Ah, ok, I misunderstood what you were suggesting.  If you didn't want to use a `hostDir`, you'd have to have the daemon looking for a particular emptyDir, or something of that sort.  Is that what you had in mind?  That seems more complicated to me.

docs/proposals/push-metrics.md
brian-brazil: The stated aim of this design is cluster level metrics, service level metrics are not cluster level. Cluster level for me means things about Kubernetes and other related infrastructure of the compute cluster in and of itself.

Similarly user pods and services are not cluster components. Cluster components are things like the kubelet and cadvisor. Pods are (usually) not components when we're talking a system like Kubernetes.

If what you want to handle are per-pod or per-service user application metrics, then please make that clear throughout the document.

DirectXMan12:
> The target audience section says this is for cluster-level metrics, and is not for arbitrary cluster users. The rest of the doc, arguments and examples are however for custom pod-level application/user metrics.

As I noted above, it's for *cluster components*.  Those components could be providing pod-level metrics, namespace-level metric, node-level metrics, etc for a bunch of different pods, namespaces, nodes, etc.

> This currently looks like putting Heapster on the path to a fully-general custom metrics solution

The intention is that the metrics are only those used for autoscaling and other cluster component requirements. This proposal does nothing to change what kind of metrics Heapster stores/collects, it just provides an avenue for getting them into Heapster when you have one source generating metrics for multiple different pods/namespaces/etc.

> My straw man proposal would be to have the autoscaler directly scrape what it needs, as the request rate shown in the example is something that the pod will already be exposing in some fashion.

The "request rate" in the example above is not coming from the pod hosting the application.  The intention (as I attempted to show by using two different pod names) was that it was coming from some sort of cluster-level cache/loadbalancer/etc which provides functionality for the entire cluster.

docs/proposals/push-metrics.md
brian-brazil: You could do it in a number of ways depending on what you want. You could have one per pod, one per node or one per service.

Service ones are actually a bit messy with this idea, though for autoscaling I'd be looking for pod rather than service-level metrics. Service-level metrics imply that aggregation has occurred, which means there's an extra hop in the control loop - which you want to avoid.

DirectXMan12:
> The "request rate" in the example above is not coming from the pod hosting the application. The intention (as I attempted to show by using two different pod names) was that it was coming from some sort of cluster-level cache/loadbalancer/etc which provides functionality for the entire cluster.

FWIW, the examples of service-level metrics below the example of pod-level metrics probably fit better with this example -- I was simply also trying to show what the label-set for pod-level metrics looks like.

brian-brazil:
> The intention (as I attempted to show by using two different pod names) was that it was coming from some sort of cluster-level cache/loadbalancer/etc which provides functionality for the entire cluster.

That doesn't change my strawman, I think it makes it a bit more sensible as there's far fewer systems that you'd need to integrate with than I'd initially thought.


This raises the question of what you propose to do for application level metrics that aren't going through a loadbalancer or other centralised infrastructure component if this design isn't meant to handle them. For say a batch system you might want resources scaled based on the size of your backlog, which would most likely come from a monitoring system.

docs/proposals/push-metrics.md
DirectXMan12: With the main proposal, since producers already need to authenticate with Heapster, we get it for free -- you could potentially just say "whoever has the certificate that allows them to authenticate is allowed to push".  Since cAdvisor doesn't currently check certificates AFAIK, you could not use the same mechanism with cAdvisor -- you'd either have to teach cAdvisor to check certificates (which seems like tying cAdvisor closer to Kube than it wants to be), or set a list of pod names (or some sort of other identifier) at the Heapster level.

docs/proposals/push-metrics.md
DirectXMan12: I apologize if that was not clear.  I'll reword a bit and make it clearer.

DirectXMan12:
> This raises the question of what you propose to do for application level metrics that aren't going through a loadbalancer or other centralised infrastructure component if this design isn't meant to handle them. For say a batch system you might want resources scaled based on the size of your backlog, which would most likely come from a monitoring system.

Just to make sure I'm understanding your scenario here: both the batch system and the monitoring system are user-controlled (i.e. neither are cluster infrastructure components), and the monitoring system knows the backlog length, and you wish to be able to scale the number of workers based on the backlog length?

If that's the case, that's a bit of a tricky scenario for either the proposal, or any of proposed alternatives, IMO, because it involves just one metric (being produced by the monitoring agent) that needs to affect multiple pods.

I think, in that case, the optimal solution might involve a bit of a tweak to the way the HPA reads custom metrics -- you'd allow the HPA to scale not just on pod-level metrics, but on namespace-level metrics (which is useful anyway).  Then you'd point the HPA scaling the workers' ReplicationController at the namespace-level aggregate corresponding to that custom metric.

DirectXMan12:
@brian-brazil Also re: the pushgateway API: do you have any thoughts on what I said in https://github.com/kubernetes/heapster/pull/1129#discussion_r61651447)?

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,229 @@
...
+
+Producers will be authenticated similarly to the way Heapster currently
+authenticates clients: producers will present a certificate signed by a CA, and
+Heapster will be configurable to only allow certain names to push metrics, or
+to allow any certificate signed by the CA.
+
+Additionally, the name presented during authentication will be used as a prefix
+for all metrics added.  This will prevent two different metrics producers from
+accidentally overwriting each other's custom metrics (otherwise, push metrics
+will be stored and retrieved identially to pull-based custom metrics).
```
ncdc: identically

docs/proposals/push-metrics.md
```diff
@@ -0,0 +1,229 @@
...
+write a program or script which collects and processes the information, and
+then add a recurring Job or cron job which pushes the metrics.  This also
+enables existing tooling designed with the push model in mind to be adapted to
+provide metrics through Heapster.
+
+### Target Audience and Metrics
+
+Like the existing custom metrics pull mechanism, the metrics pushed through the
+push mechanism are intended to be those metrics that are useful for consumption
+by system components, such as metrics inteded for use with autoscaling.  The
```
ncdc: intended

docs/proposals/push-metrics.md
DirectXMan12: :heavy_check_mark:

docs/proposals/push-metrics.md
DirectXMan12: :heavy_check_mark:

DirectXMan12:
Also, just to add another use case in, one of the components that this would be useful for is an idler.  We can set up a pod, daemon, or cron job to push in some metrics gathered about the various pods, services, etc for idling (be they special network activity metrics, some sort of metrics from a load balancer, a computed aggregate metrics, etc) and use Heapster push metrics to store them such that they're associated with the pods, services, etc that they describe.  Then the idler can later read them back (via Oldtimer, for instance) when it needs to make its idling decision.

With push metrics, we can get the relevant metrics into Heapster from different sources (since HTTP requests can be made from just about any language/environment easily), in a way that is not tied to a particular TSDB.  Additionally, since the metrics are being fed and retrieved through Heapster, and not accessed directly from the producer or from the TSDB, the idler can easily be made generic across different metrics, sources, storage (since it does not need to care about the TSDB or the original source -- it just needs be given the metric name and access it at service or RC scope).

DirectXMan12:
Re: the pushgateway compatiblity, there's four options:

1. make the pushgateway-style path available at a subpath (e.g. `/api/v1/push/prometheus/metrics/...`): it looks like the official clients would support doing this, except for the Ruby client, although I'm not sure whether this is "accidental" support, or whether it was on purpose.  This option is nice because it preserves the feel of the Heapster API, while still allowing us to work with some of the pushgateway clients OOTB.  The downside is that it's not clear whether this was intended to be supported.

2. Use the base URL without the pushgateway-style path, and just use the clients to generate the format itself, and just have them submit the request separately.  AFAICT, all the official clients support just generating the metrics into a string, except the Go client, which has libraries for generating the format, but doesn't actually support taking the metrics registered with the client and generating those to a string.  This appears to be better/more intentionally supported amongst the clients, but means that we don't work OOTB with the pushgateway clients.

3. Support 1 & 2, and hope that between the two something works for everyone's language and client of choice.

3. Just use the raw push gateway URL at the root.  This has the benefit of being guaranteed to work with the pushgateway clients, but it looks completely different from the rest of the Heapster APIs, and does not provide a clear way to support different formats in the future, if we so choose.

@brian-brazil @mwielgus @piosz what do you think?

mwielgus:
@DirectXMan12 What is your preferred option? I don't have a super strong opinion on how it should be done as I will not use it personally. Option 3 (1+2) seems to give the wider spectrum than the other although you can start with 1 or 2.

DirectXMan12:
@mwielgus I'm in favor of option 3 as well -- I think it provides the best path forwards in the future.  I'll update the docs.

DirectXMan12:
Ok, I've updated the spec to support option 3.

mwielgus:
LGTM

dchen1107:
cc/ @timstclair This is the proposal I mentioned to you a while back.

jszczepkowski:
@DirectXMan12 @piosz @fgrzadkowski 

We have some serious concerns about proposed authentication of producers. The proposed authentication,  with names of users and certificates, doesn't seem to be aligned with Kubernets authentication. We are not sure it will be useful in general case for the majority of clients.

The simplest option seem to be authentication on namespace level: pods can only push metrics to their own namespace.  Most probably, we will need to use ServiceAccounts/Secrets to verify this.

fgrzadkowski:
Apart from being aligned with k8s (which is important), I think that another problem is that it will work well for very static setups. When we have more dynamic entities (e.g. pods) who will want to push metrics to heapster than I think that static certificate is not very user friendly, unless I'm missing something.

DirectXMan12:
> We have some serious concerns about proposed authentication of producers. The proposed authentication, with names of users and certificates, doesn't seem to be aligned with Kubernets authentication. We are not sure it will be useful in general case for the majority of clients.
> The simplest option seem to be authentication on namespace level: pods can only push metrics to their own namespace.

This will not work, since it defeats the main purpose of the proposal, which custom "system components" to push metrics about all pods, etc.

> Most probably, we will need to use ServiceAccounts/Secrets to verify this.

I'm not opposed to changing the proposed authentication model, but it's currently more or less the same authentication model used by the Heapster model API (I have it have the option of using a different CA for flexibility, but otherwise it's the same).  Is that changing as well?

It might not be a bad idea to put authentication behind an interface, and then just allow multiple different authentication methods -- we could keep the certificate-based system for backwards compat/where it makes sense, drop in a new ServiceAccount based system, and, for example, provide an easy place to allow Heapster to be hooked up to the OpenShift policy system. WDYT?

> Apart from being aligned with k8s (which is important), I think that another problem is that it will work well for very static setups. When we have more dynamic entities (e.g. pods) who will want to push metrics to heapster than I think that static certificate is not very user friendly, unless I'm missing something.

Well, if you don't specify a list of names, and just have it trust anything from the CA, you can just distribute certs signed by the CA (e.g. in a secret).  What exactly did you have in mind for the service accounts, etc?  It seems like this would allow much less control over who had the ability to push metrics, unless you otherwise had a way to restrict which service accounts were allowed to do things.

jszczepkowski:
I agree it may reasonable to keep different authentication schemes: system stats w/o namespace will use CA, while namespace scoped stats will use service account. Still, this topic needs some deeper investigation.

jimmidyson:
> namespace scoped stats will use service account

I thought the discussion on this issue decided that push metrics would be restricted to system/infra stats, hence the applicability of using certificates? When you say namespace scoped stats do you mean application stats?

jszczepkowski:
>  When you say namespace scoped stats do you mean application stats?

Yes, but not only. We may have stats pushed by a pod prefixed with pod name (application stats), and more general stats prefixed only with namespace.

DirectXMan12:
> Yes, but not only. We may have stats pushed by a pod prefixed with pod name (application stats), and more general stats prefixed only with namespace.

I think we need to clarify here: when you say "prefixed with pod name/namespace", do you mean metrics stored internally at the keys "namespace:ns/pod:podname" vs "namespace:ns", or do you mean metrics producers prefixing the metric names with the associated pod or namespace?  Push metrics mainly intended for pushing "bulk" metrics that may cross the namespace from the originating pod (e.g. the pod might be in kube-system, but might describe pods from all namespaces).

jszczepkowski:
> I think we need to clarify here: when you say "prefixed with pod name/namespace", do you mean metrics stored internally at the keys "namespace:ns/pod:podname" vs "namespace:ns", or do you mean metrics producers prefixing the metric names with the associated pod or namespace?

I was thinking about the former (annotations in path): keys with "namespace:ns/pod:podname" and "namespace:ns".

DirectXMan12:
So specifically for push metrics, both of these can be pushed in a single push, so it doesn't make sense to provide different auth mechanisms for different metrics storage scopes, unless I'm misunderstanding you.

DirectXMan12:
Perhaps we should find some time for a quick hangout at some point this week?  It might be more productive than trying to play comment tag.

Also, I think it would be reasonable to proceed with push metrics auth as specified in the documentation, and then overhaul Heapster auth in a separate PR (i.e. I don't think we need to tie auth changes/a new auth interface to push metrics, necessarily).

fgrzadkowski:
I disagree. We should implement a general mechanism that will serve multiple purposes. Your use case - single pusher for the whole cluster - is one of them, but no the only. Before we will actually implement it, we should make sure it will also work in other use-case, such as "I have a pod that should be scaled based metric `my-load` and I'd like to push it somewhere so that it's available to HPA`.

PR #1138, opened 2016-04-22
Set actual InfluxDB field names

titilambert:
In order to fit with InfluxDB concepts (https://docs.influxdata.com/influxdb/v0.9/concepts/key_concepts/), this PR remove the use of field "value" when this doesn't make sense.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -122,8 +122,8 @@
 		},
 		{
 			"ImportPath": "github.com/hawkular/hawkular-client-go/metrics",
-			"Comment": "v0.5.1",
-			"Rev": "6a148493851c66d69b005283eb7932452c322b74"
+			"Comment": "v0.6.0",
+			"Rev": "50f4099dfd739f913ed7d8f33288aeeca338516a"
 		},
 		{
 			"ImportPath": "github.com/imdario/mergo",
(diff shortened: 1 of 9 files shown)


titilambert:
Documentation updated

mwielgus:
Thank you for this contribution. I understand that this is cleaner but unfortunately some users already depend on the current layout and we don't want to break things (again) just because of naming change. So we will not merge the PR, however we can revisit it and release with some bigger refactoring/update. 

cc: @piosz

piosz:
@titilambert please have in mind that we changed the schema in Kubernetes 1.2 just a few weeks ago.

titilambert:
@piosz I'm not sure I understand what you mean by "the schema". Could you give me more context about "the schema" ?

titilambert:
@mwielgus what about add an option to get the new schema with influxdb (?withfieds=true) in the URL.
Something like: ```--sink=influxdb:http://monitoring-influxdb:80/?withfields=true```

piosz:
@titilambert Heapster is a dedicated solution to monitor [Kubernetes](https://github.com/kubernetes/kubernetes) cluster. Its requirements and release process are tightly coupled with Kubernetes release process. In Kubernetes 1.1 Influxdb was in version 0.8, while in Kubernetes 1.2 in version 0.9. Also the format of data in InfluxDB between 0.8 and 0.9 has changed, so we had to change the scheme. It was just a few weeks ago and I'm against making another breaking change right now.

AFAIK in July there should be InfluxDB 1.0. We want to migrate to the new version, so we can add your change then. WDYT?

titilambert:
@piosz Thanks ! I understand now the schema issue. What do you think about `withfields` option ? This could permit smooth transitions waiting breaking change. (Of course default value should be old schema)

PR #1139, opened 2016-04-25
upgrade grafana image version to a working monitor set

king6cong:
Influxdb's data scheme doesn't match grafana's.
This should fix it


piosz:
LGTM. Thanks for the change!

PR #1151, opened 2016-05-09
Added Master Metrics API types

piosz:

### DIFF
--- a/metrics/apis/metrics/v1alpha1/types.go
+++ b/metrics/apis/metrics/v1alpha1/types.go
@@ -1,0 +1,56 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	"k8s.io/kubernetes/pkg/api/v1"
+)
+
+// resource usage metrics of a node.
+type NodeMetrics struct {
+	unversioned.TypeMeta `json:",inline"`
+	v1.ObjectMeta        `json:"metadata,omitempty"`
+
+	// The following fields define time interval from which metrics were
+	// collected from the interval [Timestamp-Window, Timestamp].
+	Timestamp unversioned.Time     `json:"timestamp"`
+	Window    unversioned.Duration `json:"window"`
+
+	// The memory usage is the memory working set.
+	Usage v1.ResourceList `json:"usage"`
+}
+
+// resource usage metrics of a pod.
+type PodMetrics struct {
+	unversioned.TypeMeta `json:",inline"`
+	v1.ObjectMeta        `json:"metadata,omitempty"`
+
+	// The following fields define time interval from which metrics were
+	// collected from the interval [Timestamp-Window, Timestamp].
+	Timestamp unversioned.Time     `json:"timestamp"`
+	Window    unversioned.Duration `json:"window"`
+
+	// Metrics for all containers are collected within the same time window.
+	Containers []ContainerMetrics `json:"containers"`
+}
+
+// resource usage metrics of a container.
+type ContainerMetrics struct {
+	// Container name corresponding to the one from pod.spec.containers.
+	Name string `json:"name"`
+	// The memory usage is the memory working set.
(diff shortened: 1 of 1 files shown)


piosz:
cc @davidopp @mwielgus

@lavalamp as we briefly discussed f2f in MTV few weeks ago metrics api types will be defined in Heapster. Is this still valid agreement?

piosz:
@k8s-bot test this

erictune:
This is the first I have heard that Master Metrics types would be in Heapster.  Can I see a link to the discussion where the change in plans was made?

piosz:
@erictune to make it clear (at least what I understood from sync with CSI team):
1. CSI team is working on 'discovery merger' project which is about to allow other components than apiserver to serve Kubernetes API. There will be a kind of proxy introduced in front of apiserver which will redirect user's queries to appropriate component, so that the user won't notice any difference while talking with the api provider. If you need more details please ask @lavalamp 
2. Heapster will act as 3rd party api server which will handle metrics requests. The types will be defined there to make client generation possible (@caesarxuchao did I understand correct this part?) 
3. For 1.3 we decided with @davidopp and @fgrzadkowski that we don't want to implement those mechanisms and only define first version of the api and make it available in Heapster, that's why I haven't written the implementation proposal.

metrics/apis/metrics/v1alpha1/types.go
```diff
@@ -0,0 +1,56 @@
...
+type NodeMetrics struct {
+	unversioned.TypeMeta `json:",inline"`
+	v1.ObjectMeta        `json:"metadata,omitempty"`
+
+	// The following fields define time interval from which metrics were
+	// collected in the following format [Timestamp-Window, Timestamp].
+	Timestamp unversioned.Time     `json:"timestamp"`
+	Window    unversioned.Duration `json:"window"`
+
+	// The memory usage is the memory working set.
```
lavalamp: What about using a key containing "working set" so this is clear?

lavalamp:
Location looks fine to me. I can't comment as to whether the contents are right or not :)

metrics/apis/metrics/v1alpha1/types.go
piosz: This is one of possibility although we agree in proposal (https://github.com/kubernetes/kubernetes/pull/25279) to use just ResourceMemory constant. We may revisit this decision later since this is just alpha version of the API.

piosz:
friendly ping @davidopp

metrics/apis/metrics/v1alpha1/types.go
```diff
@@ -0,0 +1,56 @@
...
+	"k8s.io/kubernetes/pkg/api/v1"
+)
+
+// resource usage metrics of a node.
+type NodeMetrics struct {
+	unversioned.TypeMeta `json:",inline"`
+	v1.ObjectMeta        `json:"metadata,omitempty"`
+
+	// The following fields define time interval from which metrics were
+	// collected in the following format [Timestamp-Window, Timestamp].
```
davidopp: instead of "in the following format" I would say "from he interval"

metrics/apis/metrics/v1alpha1/types.go
```diff
@@ -0,0 +1,56 @@
...
+	Usage v1.ResourceList `json:"usage"`
+}
+
+// resource usage metrics of a pod.
+type PodMetrics struct {
+	unversioned.TypeMeta `json:",inline"`
+	v1.ObjectMeta        `json:"metadata,omitempty"`
+
+	// The following fields define time interval from which metrics were
+	// collected in the following format [Timestamp-Window, Timestamp].
```
davidopp: instead of "in the following format" I would say "from the interval"

metrics/apis/metrics/v1alpha1/types.go
```diff
@@ -0,0 +1,56 @@
...
+	Timestamp unversioned.Time     `json:"timestamp"`
+	Window    unversioned.Duration `json:"window"`
+
+	// Metrics for all containers are collected within the same time window.
+	Containers []ContainerMetrics `json:"containers"`
+}
+
+// resource usage metrics of a container.
+type ContainerMetrics struct {
+	// Container name corresponding to the one from v1.Pod.Spec.Containers.
```
davidopp: Comments on fields shouldn't reference to Go types, since the comment is turned into swagger documentation and the user doesn't know the Go types. I guess you can say "corresponding to the one from pod.spec.containers" -- at least that is a little closer to the json key names (because it uses lower case and omits the v1)

metrics/apis/metrics/v1alpha1/types.go
piosz: Done

metrics/apis/metrics/v1alpha1/types.go
piosz: Done

metrics/apis/metrics/v1alpha1/types.go
piosz: Done

piosz:
comments applied

davidopp:
LGTM

piosz:
Thanks!

davidopp:
cc/ @vishh

PR #1152, opened 2016-05-10
Implementation of Metrics API operations

piosz:

Initial implementation - there is still a bunch of things to implement, added todos in the code.

<del> should be merged after #1151 </del>

cc @davidopp @fgrzadkowski

### DIFF
--- a/metrics/apis/metrics/handlers.go
+++ b/metrics/apis/metrics/handlers.go
@@ -1,0 +1,191 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


metrics/apis/metrics/handlers.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
mwielgus: 2016 :)

metrics/apis/metrics/handlers.go
```diff
@@ -0,0 +1,187 @@
...
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
```
mwielgus: Please write why we had so crazy path.

metrics/apis/metrics/handlers.go
piosz: Done

metrics/apis/metrics/handlers.go
piosz: Done

mwielgus:
LGTM

PR #1157, opened 2016-05-12
Riemann sink fixes

jsoriano:
Some fixes for riemann sink:
- Workarounds issue with int64 reflection in goryman
- Reuses code for metrics appending
- Avoid setting client to nil while sending events as it causes fatal errors

### DIFF
--- a/metrics/sinks/riemann/driver.go
+++ b/metrics/sinks/riemann/driver.go
@@ -16,6 +16,7 @@ package riemann
 
 import (
 	"net/url"
+	"reflect"
 	"runtime"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


jsoriano:
I signed it!

metrics/sinks/riemann/driver.go
```diff
@@ -188,22 +188,26 @@ func (sink *riemannSink) sendData(dataEvents []riemann_api.Event) {
...
 		}
 	}
 	end := time.Now()
-	glog.V(4).Info("Exported %d data to riemann in %s", len(dataEvents), end.Sub(start))
+	if errors > 0 {
+		glog.V(2).Info("There were errors sending events to Riemman, forcing reconnection")
+		sink.client.Close()
+		sink.client = nil
+	}
+	glog.V(4).Infof("Exported %d events to riemann in %s", len(dataEvents)-errors, end.Sub(start))
```
piosz: Shouldn't be here %v instead of %s?

piosz:
LGTM, just a nit.
Thank you for the contribution!

jsoriano:
@piosz `time` type implements `String()`, so I think it's fine with `%s`, this is how it appears in logs:
```
Exported 4001 events to riemann in 162.355751ms
```

jsoriano:
Thanks! :)

PR #1161, opened 2016-05-16
Added support for labelSelector in metrics API

piosz:
ref https://github.com/kubernetes/kubernetes/issues/23376

PR #1163, opened 2016-05-17
Implemented integration tests for Metrics API

piosz:

as promised in https://github.com/kubernetes/heapster/pull/1161#issuecomment-219452619

### DIFF
--- a/metrics/apis/metrics/handlers.go
+++ b/metrics/apis/metrics/handlers.go
@@ -95,7 +95,12 @@ func (a *Api) nodeMetricsList(request *restful.Request, response *restful.Respon
 
 func (a *Api) nodeMetrics(request *restful.Request, response *restful.Response) {
 	node := request.PathParameter("node-name")
-	response.WriteEntity(a.getNodeMetrics(node))
+	m := a.getNodeMetrics(node)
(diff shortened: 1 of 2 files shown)


metrics/apis/metrics/handlers.go
```diff
@@ -95,7 +95,11 @@ func (a *Api) nodeMetricsList(request *restful.Request, response *restful.Respon
 
 func (a *Api) nodeMetrics(request *restful.Request, response *restful.Response) {
 	node := request.PathParameter("node-name")
-	response.WriteEntity(a.getNodeMetrics(node))
+	m := a.getNodeMetrics(node)
+	if m == nil {
+		response.WriteError(http.StatusNotFound, fmt.Errorf("No metrics for ode %v", node))
```
jszczepkowski: Aren't we missing return here?

piosz: Good catch!

jszczepkowski:
It will be nice to extend the test and check, if all pods/nodes are correctly returned by master API, however, it can be addressed in another PR.

LGTM.

piosz:
Will do. Thanks for the review!

PR #1172, opened 2016-05-23
Introduce Oldtimer (with InfluxDB sink support)

DirectXMan12:
This PR introduces Oldtimer, proposed in [docs/proposals/old-timer.md](https://github.com/kubernetes/heapster/blob/master/docs/proposals/old-timer.md).

Oldtimer is accessible from the `/api/v1/historical`, and mirrors the model API.  It can be enabled by passing `--historical_source` with one of the URIs used in the `--sink` argument.

Currently this PR supports the InfluxDB sink.  Support for the rest of the metrics sinks should be coming along shortly in separate PRs (although if someone more familiar with one of the individual sinks is interested in writing the Oldtimer support for that sink, that would be appreciated).

cc @burmanm @piosz @mwielgus @bryk

### DIFF
--- a/metrics/api/v1/historical_handlers.go
+++ b/metrics/api/v1/historical_handlers.go
@@ -1,0 +1,790 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	restful "github.com/emicklei/go-restful"
+
+	"k8s.io/heapster/metrics/api/v1/types"
+	"k8s.io/heapster/metrics/core"
+	"k8s.io/heapster/metrics/util/metrics"
+)
+
+// HistoricalApi wraps the standard API to overload the fetchers to use the
+// one of the persistent storage sinks instead of the in-memory metrics sink
+type HistoricalApi struct {
+	*Api
+}
+
+// metricsAggregationFetcher represents the capability to fetch aggregations for metrics
+type metricsAggregationFetcher interface {
+	clusterAggregations(request *restful.Request, response *restful.Response)
+	nodeAggregations(request *restful.Request, response *restful.Response)
+	namespaceAggregations(request *restful.Request, response *restful.Response)
+	podAggregations(request *restful.Request, response *restful.Response)
+	podContainerAggregations(request *restful.Request, response *restful.Response)
+	freeContainerAggregations(request *restful.Request, response *restful.Response)
+	podListAggregations(request *restful.Request, response *restful.Response)
+
+	isRunningInKubernetes() bool
+}
+
+// addAggregationRoutes adds routes to a webservice which point to a metricsAggregationFetcher's methods
+func addAggregationRoutes(a metricsAggregationFetcher, ws *restful.WebService) {
+	// The /metrics-aggregated/{aggregations}/{metric-name} endpoint exposes some aggregations for the Cluster entity of the historical API.
+	ws.Route(ws.GET("/metrics-aggregated/{aggregations}/{metric-name:*}").
+		To(metrics.InstrumentRouteFunc("clusterMetrics", a.clusterAggregations)).
+		Doc("Export some cluster-level metric aggregations").
+		Operation("clusterAggregations").
+		Param(ws.PathParameter("aggregations", "A comma-separated list of requested aggregations").DataType("string")).
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metric").DataType("string")).
+		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
+		Writes(types.MetricAggregationResult{}))
+
+	// The /nodes/{node-name}/metrics-aggregated/{aggregations}/{metric-name} endpoint exposes some aggregations for a Node entity of the historical API.
+	// The {node-name} parameter is the hostname of a specific node.
+	ws.Route(ws.GET("/nodes/{node-name}/metrics-aggregated/{aggregations}/{metric-name:*}").
+		To(metrics.InstrumentRouteFunc("nodeMetrics", a.nodeAggregations)).
+		Doc("Export a node-level metric").
+		Operation("nodeAggregations").
+		Param(ws.PathParameter("node-name", "The name of the node to lookup").DataType("string")).
+		Param(ws.PathParameter("aggregations", "A comma-separated list of requested aggregations").DataType("string")).
+		Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+		Param(ws.QueryParameter("start", "Start time for requested metric").DataType("string")).
+		Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
+		Writes(types.MetricAggregationResult{}))
+
+	if a.isRunningInKubernetes() {
+		// The /namespaces/{namespace-name}/metrics-aggregated/{aggregations}/{metric-name} endpoint exposes some aggregations
+		// for a Namespace entity of the historical API.
+		ws.Route(ws.GET("/namespaces/{namespace-name}/metrics-aggregated/{aggregations}/{metric-name:*}").
+			To(metrics.InstrumentRouteFunc("namespaceMetrics", a.namespaceAggregations)).
+			Doc("Export some namespace-level metric aggregations").
+			Operation("namespaceAggregations").
+			Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")).
+			Param(ws.PathParameter("aggregations", "A comma-separated list of requested aggregations").DataType("string")).
+			Param(ws.PathParameter("metric-name", "The name of the requested metric").DataType("string")).
+			Param(ws.QueryParameter("start", "Start time for requested metrics").DataType("string")).
+			Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
+			Writes(types.MetricAggregationResult{}))
+
+		// The /namespaces/{namespace-name}/pods/{pod-name}/metrics-aggregated/{aggregations}/{metric-name} endpoint exposes
+		// some aggregations for a Pod entity of the historical API.
+		ws.Route(ws.GET("/namespaces/{namespace-name}/pods/{pod-name}/metrics-aggregated/{aggregations}/{metric-name:*}").
+			To(metrics.InstrumentRouteFunc("podMetrics", a.podAggregations)).
+			Doc("Export some pod-level metric aggregations").
+			Operation("podAggregations").
+			Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")).
+			Param(ws.PathParameter("pod-name", "The name of the pod to lookup").DataType("string")).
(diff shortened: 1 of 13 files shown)


DirectXMan12:
As far as sinks go, this PR includes InfluxDB, and Hawkular will probably be next (IIRC, @burmanm was interested in helping out there).  After that, I'm assuming GCM is the next priority.  Depending on time constrains, I can then work on OpenTSDB (or Reimann or Elasticsearch, but I haven't investigated those quite yet).

DirectXMan12:
Ok, I've removed the `[WIP]` tags from this, since it was decided that additional sinks will get their own PRs.

metrics/core/historical_types.go
```diff
@@ -0,0 +1,153 @@
...
+
+// HistoricalSource allows for retrieval of historical metrics and aggregations from sinks
+type HistoricalSource interface {
+	// GetMetric retrieves the given metric for one or more objects (specified by metricKeys) of
+	// the same type, within the given time interval
+	GetMetric(metricName string, metricKeys []HistoricalKey, start, end time.Time) (map[HistoricalKey][]TimestampedMetricValue, error)
+
+	// GetAggregation fetches the given aggregations for one or more objects (specified by metricKeys) of
+	// the same type, within the given time interval, calculated over a series of buckets
+	GetAggregation(metricName string, aggregations []string, metricKeys []HistoricalKey, start, end time.Time, bucketSize time.Duration) (map[HistoricalKey][]TimestampedAggregationValue, error)
```
burmanm: I'm not fond of the idea of passing []string type here. Couldn't we make it a type that automatically restricts the available values and makes it easier to read?

Now each sink writer has to know that these strings are actually vars from historical_types.go

Considering that every sink writer has to modify them to their internal naming / functions in any case.

DirectXMan12: Just a type alias around string?  Sure.  I often think that Go could use a proper enum type, though...

piosz:
Please do not merge this until we will cut release 1.1 branch.

metrics/core/historical_types.go
```diff
@@ -0,0 +1,155 @@
...
+
+// HistoricalSource allows for retrieval of historical metrics and aggregations from sinks
+type HistoricalSource interface {
+	// GetMetric retrieves the given metric for one or more objects (specified by metricKeys) of
+	// the same type, within the given time interval
+	GetMetric(metricName string, metricKeys []HistoricalKey, start, end time.Time) (map[HistoricalKey][]TimestampedMetricValue, error)
+
+	// GetAggregation fetches the given aggregations for one or more objects (specified by metricKeys) of
+	// the same type, within the given time interval, calculated over a series of buckets
+	GetAggregation(metricName string, aggregations []AggregationType, metricKeys []HistoricalKey, start, end time.Time, bucketSize time.Duration) (map[HistoricalKey][]TimestampedAggregationValue, error)
```
burmanm: This question applies to all of these methods. Are there parameters which can be assumed to be passed on with invalid value? Such as end time 0 or start time 0 (or both), bucketSize 0? I see there's checks for these in the example influx implementation, but nothing in these interface definitions.

And if there are, how should they be handled? I can make a lot of assumptions (and derive something from the influx example), but it would be nice to have these documented to make it easier to implement proper HistoricalSource.

metrics/core/historical_types.go
```diff
@@ -0,0 +1,155 @@
...
+type AggregationValue struct {
+	Count *uint64
+
+	Aggregations map[AggregationType]MetricValue
+}
+
+// TimestampedAggregationValue is an aggregation value with an associated timestamp
+// and bucket size
+type TimestampedAggregationValue struct {
+	Timestamp  time.Time
```
burmanm: Is this the end time of the bucket, or start time (for Hawkular we return both timestamps to the user) ? Or user definable as long as it's consistent?

metrics/sinks/influxdb/influxdb_test.go
```diff
@@ -164,3 +170,323 @@ func TestCreateInfluxdbSink(t *testing.T) {
...
+		return expected.FloatValue == actual.FloatValue
+	case core.ValueInt64:
+		return expected.IntValue == actual.IntValue
+	default:
+		return expected == actual
+	}
+}
+
+func TestHistoricalInfluxRawMetricsParsing(t *testing.T) {
+	// in order to just test the parsing, we just go directly to the sink type
```
mwielgus: What is and what is not covered by this test?

metrics/api/v1/historical_handlers.go
```diff
@@ -0,0 +1,770 @@
...
+	podAggregations(request *restful.Request, response *restful.Response)
+	podContainerAggregations(request *restful.Request, response *restful.Response)
+	freeContainerAggregations(request *restful.Request, response *restful.Response)
+	podListAggregations(request *restful.Request, response *restful.Response)
+
+	isRunningInKubernetes() bool
+}
+
+// addAggregationRoutes adds routes to a webservice which point to a metricsAggregationFetcher's methods
+func addAggregationRoutes(a metricsAggregationFetcher, ws *restful.WebService) {
```
mwielgus: Are you planning any tests for the handlers?

metrics/core/historical_types.go
DirectXMan12: I'll add in documentation to the interface, good catch.

DirectXMan12: (but yes, they can have zero/empty values, which indicate no bounds for time, and to use only a single bucket for bucket size).

metrics/core/historical_types.go
DirectXMan12: In Influx, it's the start time IIRC, so we should stick with that for consistency.  I'll add a note in.

metrics/sinks/influxdb/influxdb_test.go
DirectXMan12: This tests turning the JSON returned by InfluxDB for a  "query row" (which is actually a set of results) into internal Heapster form (both for the normal queries and the aggregated queries).

metrics/api/v1/historical_handlers.go
DirectXMan12: I can write some if desired.

ncdc:
What's the status on this? @DirectXMan12 is there more code to do from your end? @mwielgus does your team have any other comments? I see that the release-1.1 branch exists, so how close do you think we are to merging this?

metrics/api/v1/historical_handlers.go
mwielgus: Yeah, it would be nice. Can be in a following PR but please do them.

metrics/api/v1/historical_handlers.go
```diff
@@ -0,0 +1,770 @@
...
+			Param(ws.QueryParameter("end", "End time for requested metric").DataType("string")).
+			Writes(types.MetricAggregationResult{}))
+
+		// The /pod-id/{pod-id}/metrics-aggregated/{aggregations}/{metric-name} endpoint exposes
+		// some aggregations for a Pod entity of the historical API.
+		ws.Route(ws.GET("/pod-id/{pod-id}/metrics-aggregated/{aggregations}/{metric-name:*}").
+			To(metrics.InstrumentRouteFunc("podMetrics", a.podAggregations)).
+			Doc("Export some pod-level metric aggregations").
+			Operation("podAggregations").
+			Param(ws.PathParameter("pod-id", "The id of the pod to lookup").DataType("string")).
```
mwielgus: What exactly is pod id. Please document.

metrics/api/v1/historical_handlers.go
```diff
@@ -0,0 +1,770 @@
...
+
+	if a.isRunningInKubernetes() {
+		// The /namespaces/{namespace-name}/metrics-aggregated/{aggregations}/{metric-name} endpoint exposes some aggregations
+		// for a Namespace entity of the historical API.
+		ws.Route(ws.GET("/namespaces/{namespace-name}/metrics-aggregated/{aggregations}/{metric-name:*}").
+			To(metrics.InstrumentRouteFunc("namespaceMetrics", a.namespaceAggregations)).
+			Doc("Export some namespace-level metric aggregations").
+			Operation("namespaceAggregations").
+			Param(ws.PathParameter("namespace-name", "The name of the namespace to lookup").DataType("string")).
+			Param(ws.PathParameter("aggregations", "A comma-separated list of requested aggregations").DataType("string")).
```
mwielgus: what aggregations are supported.

metrics/sinks/influxdb/influxdb_historical.go
```diff
@@ -0,0 +1,550 @@
...
+		return fmt.Errorf("Invalid pod id %q", key.PodId)
+	}
+
+	return nil
+}
+
+// aggregationFunc converts an aggregation name into the equivalent call to an InfluxQL
+// aggregation function
+func (sink *influxdbSink) aggregationFunc(aggregationName core.AggregationType, fieldName string) string {
+	switch aggregationName {
```
mwielgus: what about sql injection. can we test that the field name is sane?

metrics/sinks/influxdb/influxdb_historical.go
```diff
@@ -0,0 +1,550 @@
...
+	// This should have been checked by the API level, so something's seriously wrong here
+	panic(fmt.Sprintf("Unknown aggregation type %q", aggregationName))
+}
+
+// keyToSelector converts a HistoricalKey to a InfluxQL predicate
+func (sink *influxdbSink) keyToSelector(key core.HistoricalKey) string {
+	typeSel := fmt.Sprintf("type = '%s'", key.ObjectType)
+	switch key.ObjectType {
+	case core.MetricSetTypeNode:
+		return fmt.Sprintf("%s AND %s = '%s'", typeSel, core.LabelNodename.Key, key.NodeName)
```
mwielgus: same here, can we escape the parameters?

metrics/heapster.go
```diff
@@ -54,6 +54,7 @@ var (
 	argAllowedUsers     = flag.String("allowed_users", "", "comma-separated list of allowed users")
 	argSources          flags.Uris
 	argSinks            flags.Uris
+	argHistoricalSource = flag.String("historical_source", "", "which source type to use for the historical API (should be exactly the same as one of the sink URIs)")
```
mwielgus: what if left empty?

metrics/api/v1/historical_handlers.go
DirectXMan12: ack, will do.  It's the UID of the pod.

metrics/api/v1/historical_handlers.go
DirectXMan12: The aggregations documented in the proposal.  Do you want that documented elsewhere, or a subpath which just returns that list?

metrics/sinks/influxdb/influxdb_historical.go
DirectXMan12: Key values are checked in `checkSanitizedKey` above, and an error is thrown if anything besides `/^[a-zA-Z0-9_.-]+$/` is found.

metrics/heapster.go
DirectXMan12: It is disabled.  I've added a note.

metrics/sinks/influxdb/influxdb_historical.go
DirectXMan12: `aggregationName` values are checked against a fixed list at the API level (see `getAggregations` in `metrics/api/v1/historical_handlers.go:694`)

DirectXMan12:
Ok, I've added tests for the model handlers.

mwielgus:
Tests fail

DirectXMan12:
Oddly, it passes locally (go version go1.6.1 linux/amd64). Looks like timezones in the timestamps in the test are changing across serialization and deserialization.

mwielgus:
Can you fix it anyhow? I cannot merge a pr that breaks unit test.

DirectXMan12:
@mwielgus yep, just need to figure out what's going on ;-)

DirectXMan12:
Ok, I've pushed one final change, which makes the start time parameter mandatory (and mandatorily non-zero), since different sinks treat a zero time differently, and some sinks have issues with excessively large duration (making an explicit start time mandatory makes sense here).  Thanks to @burmanm for pointing out the issue.

Thanks to @ncdc for tracking down the issue with the times.  It had to do with Go having a special "Local" timezone internally that was different than the actually local timezone (but "equivalent"), so `DeepEqual` considered them different (not entirely sure why it worked in east coast time, though).

DirectXMan12:
@mwielgus should be good to go!

mwielgus:
LGTM

PR #1209, opened 2016-07-06
Hawkular Oldtimer support

burmanm:
Adds the ability of Oldtimer to use Hawkular as the backend.
- It's not in every situation possible to get 1:1 matching against internal API. This is because it's possible to configure Hawkular-Metrics sink with parameters that can't be transformed back with the information available in the requests in the Oldtimer API.
- Updates the Hawkular-Metrics client and current sink with small changes.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1209)

### DIFF
--- a/metrics/sinks/hawkular/driver_historical.go
+++ b/metrics/sinks/hawkular/driver_historical.go
@@ -1,0 +1,433 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hawkular
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/hawkular/hawkular-client-go/metrics"
+	"k8s.io/heapster/metrics/core"
+)
+
+// Historical returns the historical data access interface for this sink
+func (h *hawkularSink) Historical() core.HistoricalSource {
+	return h
+}
+
+// GetMetric retrieves the given metric for one or more objects (specified by metricKeys) of
+// the same type, within the given time interval
+func (h *hawkularSink) GetMetric(metricName string, metricKeys []core.HistoricalKey, start, end time.Time) (map[core.HistoricalKey][]core.TimestampedMetricValue, error) {
+	typ := h.metricNameToHawkularType(metricName)
+
+	resLock := &sync.Mutex{}
+	res := make(map[core.HistoricalKey][]core.TimestampedMetricValue, len(metricKeys))
+
+	errChan := make(chan error, len(metricKeys))
+	wg := &sync.WaitGroup{}
+
+	for _, keyM := range metricKeys {
+		wg.Add(1)
+		go func(key core.HistoricalKey) {
+			defer wg.Done()
+			tags := h.keyToTags(&key)
+			tags[descriptorTag] = metricName
+
+			o := []metrics.Modifier{metrics.Filters(metrics.TagsFilter(tags), metrics.TypeFilter(typ))}
+			if h.isNamespaceTenant() && key.NamespaceName != "" && key.ObjectType != core.MetricSetTypeCluster {
+				o = append(o, metrics.Tenant(key.NamespaceName))
+			}
+
+			// Remove in favor of using ReadRaw with filters to reduce the amount of queries to HWKMETRICS,
+			// but that API is still considered unstable in current release
+			metricDefs, err := h.client.Definitions(o...)
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			if len(metricDefs) > 1 {
+				errChan <- fmt.Errorf("Given metric query (metricName=%q, key=%q) did not result in unique metricId, found %d results", metricName, key, len(metricDefs))
+				return
+			} else if len(metricDefs) < 1 {
+				return
+			}
+
+			filters := make([]metrics.Filter, 0, 2)
+
+			// Without start & endTime, the default is requesting 8 hours to the past
+			if !end.IsZero() {
+				filters = append(filters, metrics.EndTimeFilter(end))
+			}
+
+			if !start.IsZero() {
+				filters = append(filters, metrics.StartTimeFilter(start))
+			}
+
+			datapoints, err := h.client.ReadRaw(metricDefs[0].Type, metricDefs[0].ID, metrics.Filters(filters...))
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			if len(datapoints) > 0 {
+				metricValues := make([]core.TimestampedMetricValue, 0, len(datapoints))
+
+				for _, datapoint := range datapoints {
+					value, err := h.datapointToMetricValue(datapoint, &typ)
+					if err != nil {
+						errChan <- err
+						return
+					}
+					metricValues = append(metricValues, value)
+				}
+
+				resLock.Lock()
+				res[key] = metricValues
+				resLock.Unlock()
+			}
+
+		}(keyM)
+	}
+	wg.Wait()
+	close(errChan)
+
+	// Check for errors first
+	for e := range errChan {
+		return nil, e
+	}
+
+	return res, nil
+}
+
+// GetAggregation fetches the given aggregations for one or more objects (specified by metricKeys) of
+// the same type, within the given time interval, calculated over a series of buckets
+func (h *hawkularSink) GetAggregation(metricName string, aggregations []core.AggregationType, metricKeys []core.HistoricalKey, start, end time.Time, bucketSize time.Duration) (map[core.HistoricalKey][]core.TimestampedAggregationValue, error) {
+	typ := h.metricNameToHawkularType(metricName)
+
+	resLock := &sync.Mutex{}
+	res := make(map[core.HistoricalKey][]core.TimestampedAggregationValue, len(metricKeys))
+
+	errChan := make(chan error, len(metricKeys))
+	wg := &sync.WaitGroup{}
+
+	// While we could read everything in a single query to Hawkular-Metrics, we could not reliably parse the results back to HistoricalKey
+	for _, keyM := range metricKeys {
+		wg.Add(1)
+		go func(key core.HistoricalKey) {
+			defer wg.Done()
+			tags := h.keyToTags(&key)
+			tags[descriptorTag] = metricName
+
+			o := make([]metrics.Modifier, 0, 2)
+			if h.isNamespaceTenant() && key.NamespaceName != "" && key.ObjectType != core.MetricSetTypeCluster {
+				o = append(o, metrics.Tenant(key.NamespaceName))
+			}
+
+			filters := make([]metrics.Filter, 0, 5)
+			filters = append(filters, metrics.TagsFilter(tags))
+
+			if bucketSize != 0 {
+				filters = append(filters, metrics.BucketsDurationFilter(bucketSize))
+			} else {
+				// No bucketSize defined, lets request a single bucket over all the data
+				filters = append(filters, metrics.BucketsFilter(1))
+			}
+
+			// Without start & endTime, the default is requesting 8 hours to the past
+			if !end.IsZero() {
+				filters = append(filters, metrics.EndTimeFilter(end))
+			}
+
+			if !start.IsZero() {
+				filters = append(filters, metrics.StartTimeFilter(start))
+			}
+
+			// No need to request percentile 50 as that equals median
+			filters = append(filters, metrics.PercentilesFilter([]float64{95.0, 99.0}))
+
+			o = append(o, metrics.Filters(filters...))
+			buckets, err := h.client.ReadBuckets(typ, o...)
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			if len(buckets) > 0 {
+				aggregationValues := make([]core.TimestampedAggregationValue, 0, len(buckets))
+
+				for _, bucketPoint := range buckets {
+					value, err := h.bucketPointToAggregationValue(bucketPoint, &typ, aggregations, bucketSize)
+					if err != nil {
+						errChan <- err
+						return
+					}
+
+					aggregationValues = append(aggregationValues, value)
+				}
+				resLock.Lock()
+				res[key] = aggregationValues
+				resLock.Unlock()
+			}
+
+		}(keyM)
+	}
+	wg.Wait()
+	close(errChan)
+
+	// Check for errors first
+	for e := range errChan {
+		return nil, e
+	}
+
+	return res, nil
+}
+
+// GetMetricNames retrieves the available metric names for the given object
+func (h *hawkularSink) GetMetricNames(metricKey core.HistoricalKey) ([]string, error) {
+	tags := h.keyToTags(&metricKey)
+	tags[descriptorTag] = "*"
+
+	var r map[string][]string
+	var err error
+
+	if h.isNamespaceTenant() && metricKey.NamespaceName != "" {
+		r, err = h.client.TagValues(tags, metrics.Tenant(metricKey.NamespaceName))
+	} else {
+		r, err = h.client.TagValues(tags)
+	}
+
+	if r != nil {
+		return r[descriptorTag], err
+	}
+	return nil, err
+}
+
+// GetNodes retrieves the list of nodes in the cluster
+func (h *hawkularSink) GetNodes() ([]string, error) {
+	tags := make(map[string]string)
+	tags[core.LabelHostname.Key] = "*"
+	tags[core.LabelMetricSetType.Key] = core.MetricSetTypeNode
+	r, err := h.client.TagValues(tags) // Assume these are stored in the system tenant
+	if r != nil {
+		return r[core.LabelHostname.Key], err
+	}
+	return nil, err
+}
+
+// GetPodsFromNamespace retrieves the list of pods in a given namespace
+func (h *hawkularSink) GetPodsFromNamespace(namespace string) ([]string, error) {
+	tags := make(map[string]string)
+	tags[core.LabelPodName.Key] = "*"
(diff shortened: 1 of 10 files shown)


metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+			tags[descriptorTag] = metricName
+
+			o := []metrics.Modifier{metrics.Filters(metrics.TagsFilter(tags), metrics.TypeFilter(typ))}
+			if h.isNamespaceTenant() && key.NamespaceName != "" && key.ObjectType != core.MetricSetTypeCluster {
+				o = append(o, metrics.Tenant(key.NamespaceName))
+			}
+
+			// Remove in favor of using ReadRaw with filters to reduce the amount of queries to HWKMETRICS,
+			// but that API is still considered unstable in current release
+			mds, err := h.client.Definitions(o...)
```
DirectXMan12: Do we need to get definitions?  Isn't the name that the driver stores metrics under effectively a function of the metric key (or is the goal here to support changing the naming scheme in the future)?

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+			if len(mds) > 1 {
+				errChan <- fmt.Errorf("Given metric query (metricName=%q, key=%q) did not result in unique metricId, found %d results", metricName, key, len(mds))
+				return
+			} else if len(mds) < 1 {
+				return
+			}
+
+			filters := make([]metrics.Filter, 0, 2)
+
+			// Without start & endTime, the default is requesting 8 hours to the past
```
DirectXMan12: This concerns me.  We should decide what zero means across all drivers (8 hours, since the epoch, etc) and always set it correctly, or decide to prevent unspecified start times at the API level.  I don't want different drivers having different functionality when we can avoid it.

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+				filters = append(filters, metrics.StartTimeFilter(start))
+			}
+
+			dps, err := h.client.ReadRaw(mds[0].Type, mds[0].ID, metrics.Filters(filters...))
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			if len(dps) > 0 {
```
DirectXMan12: Nit: please avoid all-acronym variable names like `dps` and `mds` when you can, since it generally makes code much harder to read at a glance (in this case, you could use something like `metricDefs` and `datapoints` instead).

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+		// }
+	}
+	tmv.MetricValue = mv
+
+	return tmv, nil
+}
+
+func (h *hawkularSink) metricNameToHawkularType(metricName string) metrics.MetricType {
+	for _, metric := range core.AllMetrics {
+		if metric.Name == metricName {
```
DirectXMan12: This... is suboptimal.  We should probably just compose a map based on `AllMetrics`  in `core` (e.g. `core.AllMetricsByName`) or have a function, since looking up metric descriptors by name is a useful thing to be able to do and having to loop over all metrics every time is not the best.

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+			return heapsterTypeToHawkularType(metric.Type)
+		}
+	}
+	return metrics.Gauge
```
DirectXMan12: How do custom metrics factor in here?  If, in the future, we have a way to have custom metrics of different types (e.g. push metrics), how would that be dealt with here?

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+		} else {
+			tags[core.LabelNamespaceName.Key] = metricKey.NamespaceName
+			tags[core.LabelPodName.Key] = metricKey.PodName
+		}
+	case core.MetricSetTypeNamespace:
+		tags[core.LabelNamespaceName.Key] = metricKey.NamespaceName
+	case core.MetricSetTypeNode:
+		tags[core.LabelNodename.Key] = metricKey.NodeName
+	case core.MetricSetTypeCluster:
+		// System tenant?
```
DirectXMan12: can you elaborate on what the question is here?

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+	tags := make(map[string]string)
+	tags[labelKey] = "*"
+	r, err := h.client.TagValues(tags, o...)
+	if r != nil {
+		return r[labelKey], err
+	}
+	return nil, err
+}
+
+func (h *hawkularSink) isNamespaceTenant() bool {
```
DirectXMan12: can you add docs to this method to explain what it does/why it's necessary, please?

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+
+	for _, keyM := range metricKeys {
+		wg.Add(1)
+		go func(key core.HistoricalKey) {
+			defer wg.Done()
+			tags := h.keyToTags(&key)
+			tags[descriptorTag] = metricName
+
+			o := []metrics.Modifier{metrics.Filters(metrics.TagsFilter(tags), metrics.TypeFilter(typ))}
+			if h.isNamespaceTenant() && key.NamespaceName != "" && key.ObjectType != core.MetricSetTypeCluster {
```
DirectXMan12: Please drop a comment here about what `isNamespaceTenant` is doing here
(something like `// if we're also separating namespaces by tenant, make sure to also filter on the tenant name`, if I'm reading this correctly)

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,432 @@
...
+func (h *hawkularSink) GetNamespaces() ([]string, error) {
+	// Fetch tenants, if the labelToTenantId is set (and includes namespace label)
+	if h.isNamespaceTenant() {
+		tds, err := h.client.Tenants()
+		if err != nil {
+			return nil, err
+		}
+		ns := make([]string, 0, len(tds))
+		for _, td := range tds {
+			ns = append(ns, td.ID)
```
DirectXMan12: what happens here if we have some tenants that don't represent namespaces (e.g. this Hawkular setup is being shared with other metrics sources), or is such a setup not generally allowed?

DirectXMan12:
In general, it looks like a good start.  I've got some questions about a couple of things, although a couple might be due to not knowing the specifics of the Hawkular Go API.

metrics/sinks/hawkular/driver_historical.go
burmanm: Currently such setup is not supported and we do not have any metadata associated to filter out non-Heapster tenants at this point.

metrics/sinks/hawkular/driver_historical.go
burmanm: Yes, I'll add (it's an Openshift specific behavior to store each namespace on separate tenant)

metrics/sinks/hawkular/driver_historical.go
burmanm: This is included in several places, so I'll add to the method of isNamespaceTenant a longer description if that's alright. It's not actually filtering as by default all the requests go to a default tenant. So it's required to know which tenant has the data to find the metric (requesting from wrong tenant would return wrong data or no data - each metricId+type is unique inside a tenant, so different tenants can have same metricIds)

metrics/sinks/hawkular/driver_historical.go
burmanm: An extra question mark ;). It's just a reminder that "TypeCluster" must always target the default tenant in Openshift (that's where the cluster level information is stored in)

metrics/sinks/hawkular/driver_historical.go
burmanm: I assume push metrics have a type as well, being either Gauges or Cumulatives? Gauge datatype allows richer possiblities as it uses float64 instead of int64, so that's why I've set it as default when nothing else matches. As long as the matching is consistent it shouldn't matter much.

metrics/sinks/hawkular/driver_historical.go
burmanm: Agreed. The 8 hours is the default in Hawkular when user gives no range restrictions, but I'd much rather have some application specific restriction here. Should it be added to the API level that if both are empty, the API would set them with some default? Instead of doing the same filtering on each historicalSource?

metrics/sinks/hawkular/driver_historical.go
burmanm: Technically it is, but with some confusing methods (I have no idea on the history of why), if you look at https://github.com/kubernetes/heapster/blob/master/metrics/sinks/hawkular/client.go#L102 that one could be still resolved back in most cases, however there's for some reason the following addition at some point in the past:

https://github.com/kubernetes/heapster/blob/master/metrics/sinks/hawkular/client.go#L154

And here we don't know if there was a ResourceID set or not. I don't personally like to use metricIds for anything as they provide no structure at all (and should I have my wish, they would be random strings..).

However, after 0.18.0 of HWKMETRICS is out, this Definitions() fetching can be removed and the tags are used with ReadRaw to find the metrics directly - reducing the amount of calls. I just didn't want to use the non-stable API in this integration. Equal to the way GetAggregations is doing the work (and e I could probably reuse a lot more code).

metrics/sinks/hawkular/driver_historical.go
burmanm: I'll change (the acronyms are from returned data types).

metrics/sinks/hawkular/driver_historical.go
burmanm: Yep, agreed. Will you add that in the Oldtimer PR or will I add something to create a map in the start of the HawkularSink and use that instead of looping everytime?

metrics/sinks/hawkular/driver_historical.go
DirectXMan12: Ack, that makes sense.

metrics/sinks/hawkular/driver_historical.go
DirectXMan12: Yeah, I think that will probably avoid the most issues here.  I'll update my original PR.

metrics/sinks/hawkular/driver_historical.go
DirectXMan12: I'll add it to the Oldtimer PR.  I see it being useful across different sinks, etc.

metrics/sinks/hawkular/driver_historical.go
DirectXMan12: So, currently in my push metrics prototype, I just use one type (gauge, IIRC), but that's mainly because the Prometheus ingester API doesn't expose the type annotation in the right places.

Eventually, I'd like the be able to respect the annotations on the pushed metrics, so that gauges are stored as a gauge, and cumulative metrics are stored as cumulative.  I'll have to think a bit about how to resolve that.

metrics/sinks/hawkular/driver_historical.go
DirectXMan12: actually, can you just add it in a separate commit in this PR, but in the same place that core.AllMetrics is?  That way there's some context, and not just a change in the Oldtimer PR that does not appear to have justification.

metrics/sinks/hawkular/driver_historical.go
DirectXMan12: The Oldtimer PR as it went in requires start time (and end time is actually set at the API level to be `time.Now()`), so you should probably amend this comment to say "... but this is prevented at the API level"

DirectXMan12:
Just did another quick pass, and I think this mostly looks good.  The only thing I'm worried about here is assuming the gauge type for unknown metrics -- the push metrics PR allows for custom metrics of both the cumulative and gauge type, so assuming gauge for custom metrics will miss some metrics.  Is there any way to search both gauge and counter metrics in Hawkular?  Alternatively, perhaps we could put some sort of marker in the name (e.g. anything ending in `_rate` is always a gauge, etc)?  WDYT?

burmanm:
It is possible to search for both types, but since the type is part of the unique key, one could have "cpu/usage_rate" "GAUGE" and "cpu/usage_rate" "COUNTER", thus it would be quite difficult to find out which one is the correct one.

I'm not sure what you mean with "miss some metrics", as long as we store it with same assumption. And at this point we store unknown metricType as Gauge also (so transformation happens two ways in identical way). 

Also, Gauge type supports the same operations, such as rate calculation on the server side (but does not have understanding for something like counter reset obviously) so one is not missing any functional behavior.

DirectXMan12:
>  one could have "cpu/usage_rate" "GAUGE" and "cpu/usage_rate" "COUNTER", thus it would be quite difficult to find out which one is the correct one.

Internally, Heapster keys on metric name only (https://github.com/kubernetes/heapster/blob/master/metrics/core/types.go#L123), so data fetched via Oldtimer, which was written originally by Heapster, should not have issues in that regard.

 > I'm not sure what you mean with "miss some metrics", as long as we store it with same assumption. And at this point we store unknown metricType as Gauge also (so transformation happens two ways in identical way). 

I've updated the push metrics PR to properly read which type (cumulative vs gauge) a metric is marked as, and store in Heapster accordingly.  Unless the Hawkular sink implementation ignores the `MetrcType` field (https://github.com/kubernetes/heapster/blob/master/metrics/core/types.go#L90), you're going to have some custom metrics stored as gauge and some stored as cumulative.  Ergo, if you assume all custom metrics are gauges, and you only search for gauges, you're going to miss cumulative custom metrics.

burmanm:
No, Hawkular does not ignore MetricType, it uses it. Could the push metrics expose those defined metrics in some way, so that we could use that information here to search for correct type also?

DirectXMan12:
So, push metrics are transient -- there's no guarantee that the pushed metrics waiting to be scraped have the same names as the ones stored in Hawkular.

However, I just remembered seeing something: doesn't the Hawkular sink keep metrics definitions around: https://github.com/kubernetes/heapster/blob/master/metrics/sinks/hawkular/types.go#L52

Could that information be used here?

(also, it occurs to me, is there going to be a problem with the Hawkular sink for push metrics, since not all push metric names are know at startup time?)

burmanm:
Yes, the map of registered metrics can be used here, but there's the issue of backtracking the correct one, as the metrics.Definition includes a generated id. registerLabeledIfNecessary calls the idName() to generate the identifier.

However, it also uses the heapsterTypeToHawkularType so the typing should be the exact same as what we're trying to find here. Because of this, I'm finding it quite difficult to think of a situation where we would end up with different result on different iterations..

Although you strikeout the pushmetrics/startup time, I'll clarify here if someone wonders. Hawkular-Metrics Registers only models in the startup, not actual metricDefinitions (those are stored as we process metrics). However, I must read your push metrics to see if there's something that needs working (to avoid unnecessary calls / registerations / etc)

DirectXMan12:
The push metrics PR is #1217.

DirectXMan12:
@burmanm did you get a chance to take a look at the push metrics PR?  I'd like to get this in soon.

burmanm:
@DirectXMan12 Hi, yes I did.. I couldn't find the type as an issue, however there's one thing that caught my eye:

``// TODO: core.LabelPodNamespace?``

This is quite important label for us, as it's used in Openshift by default as the namespace divider currently:

https://github.com/openshift/origin-metrics/blob/master/deployer/templates/heapster.yaml#L83

Without the information, things will be written to the system tenant and that's a problem and not just for Oldtimer API.

DirectXMan12:
ack, I'll add that in.  I think I left it out because I wasn't sure if it was redundant (b/c there's already `namespace_name` as a different label).

DirectXMan12:
@burmanm fixed

piosz:
@DirectXMan12 please grant LGTM when you think it's ready to merge. Then I'll take a look.

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,433 @@
...
+				}
+			}
+		}
+
+		mv := core.MetricValue{
+			MetricType: core.MetricGauge,
+			ValueType:  core.ValueFloat,
+			FloatValue: float32(val),
+		}
+		f, err := metrics.ConvertToFloat64(val)
```
DirectXMan12: missed this on my first pass through.  What's the point of calling this function?  We've already established that `val` converts to a `float32` directly above, so why call this function?

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,433 @@
...
+
+	tmv := core.TimestampedMetricValue{
+		Timestamp: dp.Timestamp,
+	}
+
+	switch *mt {
+	case metrics.Counter:
+		mv.ValueType = core.ValueInt64
+		if v, ok := dp.Value.(int64); ok {
+			mv.IntValue = int64(v)
```
DirectXMan12: this cast is redundant

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,433 @@
...
+	case metrics.Gauge:
+		mv.ValueType = core.ValueFloat
+		f, err := metrics.ConvertToFloat64(dp.Value)
+		if err != nil {
+			return tmv, err
+		}
+		mv.FloatValue = float32(f)
+		// if v, ok := dp.Value.(float64); ok {
+
+		// }
```
DirectXMan12: remove this comment, please

metrics/sinks/hawkular/driver_historical.go
```diff
@@ -0,0 +1,433 @@
...
+
+			// Remove in favor of using ReadRaw with filters to reduce the amount of queries to HWKMETRICS,
+			// but that API is still considered unstable in current release
+			metricDefs, err := h.client.Definitions(o...)
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			if len(metricDefs) > 1 {
```
DirectXMan12: It didn't occur to me until just now, but this may be problematic when you have multiple pods with the same name, but that were alive during different time periods.  They could have different UIDs (meaning different entries in Hawkular), but since they have the same name, they'll all be returned here.  Since the definition query doesn't restrict to "time that the pod is alive" I think you may need to just perform multiple queries, *and then* figure out if there were multiple pods alive during that time period and error out.

burmanm:
Rebased from master and fixed the issues @DirectXMan12 mentioned

piosz:
closing as obsolete

PR #1210, opened 2016-07-06
Remove quoting around --sink argument

igalic:
after two days of debugging why we're not getting any metrics, we tried removing the quotes around the --sink argument and it started working.

### DIFF
--- a/deploy/kube-config/kafka/heapster-controller.yaml
+++ b/deploy/kube-config/kafka/heapster-controller.yaml
@@ -25,4 +25,4 @@ spec:
         command:
         - /heapster
         - --source=kubernetes:https://kubernetes.default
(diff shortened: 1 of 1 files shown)


burmanm:
Using without quotes causes your shell to interpret "&" as background job and probably not what you want.

igalic:
that's not our experience, it appears as though something else is already doing the quoting for us before it hits the shell

jimmidyson:
This isn't running a shell, this is executing the command explicitly as pid 1. Could you share the logs? I'd expect the quotes to not be required in this case tbh.

igalic:
showing logs is kinda a ootstrapping issue :D

mwielgus:
Please sign the CLA.

igalic:
I signed it!

mwielgus:
LGTM

PR #1212, opened 2016-07-07
move processors init out of main func

liubog2008:
I want to customize processors, so I move the init out of main func and create a processors factory, just like sources.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1212)

### DIFF
--- a/metrics/heapster.go
+++ b/metrics/heapster.go
@@ -31,7 +31,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/heapster/common/flags"
 	kube_config "k8s.io/heapster/common/kubernetes"
-	"k8s.io/heapster/metrics/core"
 	"k8s.io/heapster/metrics/manager"
 	"k8s.io/heapster/metrics/processors"
 	"k8s.io/heapster/metrics/sinks"
@@ -54,6 +53,7 @@ var (
 	argAllowedUsers     = flag.String("allowed_users", "", "comma-separated list of allowed users")
 	argSources          flags.Uris
 	argSinks            flags.Uris
+	argProcessors       = flag.String("processors", "kubernetes", "processors for heapster")
 )
 
 func main() {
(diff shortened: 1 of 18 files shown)


mwielgus:
Why do you need this change?

BTW, please sign the CLA.

liubog2008:
I signed it!

liubog2008:
I want to add some labels and make data more useful for us, so that I don't need to process data again after getting data from influxdb.

metrics/heapster.go
```diff
@@ -54,6 +53,7 @@ var (
 	argAllowedUsers     = flag.String("allowed_users", "", "comma-separated list of allowed users")
 	argSources          flags.Uris
 	argSinks            flags.Uris
+	argProcessors       = flag.String("processors", "kubernetes", "processors for heapster")
```
mwielgus: Actually, lets name the default set "default".  As Heapster works only in kubernetes name "kubernetes" doesn't mean anything.

metrics/processors/kubeprocessors/cluster_aggregator.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package processors
+package kubeprocessors
```
mwielgus: default

metrics/processors/kubeprocessors/pod_based_enricher.go
```diff
@@ -196,7 +201,19 @@ func intValue(value int64) core.MetricValue {
 	}
 }
 
-func NewPodBasedEnricher(podLister *cache.StoreToPodLister) (*PodBasedEnricher, error) {
+func NewPodBasedEnricher(url *url.URL) (*PodBasedEnricher, error) {
```
mwielgus: We pass pod listener for a reason. In large clusters with thousands of pods podsListeners are quite big (they contain a cache of ALL pods in the system with all metadata) and having 2 copies of them would be a big waste of memory.

liubog2008: Ok, It seems the biggest problem in my pr.  I'll try to create a public cache repository for getting cache store or indexer.

metrics/heapster.go
liubog2008: OK, I'll change it... In fact, I just imitate what Sources do.

piosz:
@liubog2008 This PR is inactive for a longer period of time. If you plan to work on this, please reopen.

PR #1213, opened 2016-07-07
Heapster Auth Revamp

DirectXMan12:
This proposal details an interface which would allow for flexible
authentication and authorization interfaces.  This is useful for
controlling push metrics, as well as for restricting access to the
heapster model in a more fine-grained fashion than the current
"anyone succesfully authenticated is authorized".

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1213)

### DIFF
--- a/docs/proposals/auth.md
+++ b/docs/proposals/auth.md
@@ -1,0 +1,357 @@
+Heapster Authentication and Authorization
+=========================================
+
+Current State
+-------------
+
+Heapster authentication is based on checking client certificates against a CA.
+Authorization is either just a static list of names, or non-existant.  In
+either case, anyone authenticated is authorized to perform any operation.  With
+push metrics, the proposed plan was to allow adding a second CA, to allow
+separating authorization for model and Oldtimer queries from authorization to
+push metrics.  However, this is suboptimal, since it also separates
+authentication.
+
+Requirements
+------------
+
+Heapster should be able to separate authentication from authorization, and
+the authorization interface should be flexible enough to support authorization
+implementations that are able to make the following distinctions:
+
+- containing queries to a particular namespace: we should be able to only allow
+  pods to query for metrics within their own namespace.
+
+- separating querying from pushing: we should allow certain users to push, and
+  certain users to query, but not necessarily tie the two together
+
+- allowing different push scope permissions: we should be able to specify that
+  some users are only allowed to push metrics for the pods within their own
+  namespace (or given sufficient information in the future, possibly their own
+  RC), while others should be allowed to push for any object in the cluster.
+
+Proposed Solution
+-----------------
+
+Heapster will adopt an authentication and authorization model based around
+querying the Kubernetes model.  Authentication will support either client
+certificate and tokens using the Kubernetes API server tokenreview
+endpoint. Authorization will be done using the Kubernetes API server
+subjectaccessreviews endpoint.
+
+### Authentication ###
+
+Authentication will support both client certificates and tokens.  Client
+certificates will be supported by checking against a configurable CA (this
+will generally be the cluster CA that gets injected into the Heapster
+pod), while tokens will be checked against the Kubernetes API server using
+a `TokenReview`.
+
+The configuration will look as such:
+
+- `--authn-ca=/path/to/ca.crt`
+- `--auth-apiserver=https://$KUBE`
+
+### Authorization ###
+
+Authorization will work by matching Heapster concepts to Kubernetes RBAC
+concepts, and then performing `SubjectAccessReview` requests against the
+API server configured in `--auth-apiserver`.
+
+The requests will use an API group of "heapster.k8s.io", and resources
+will be referred to as "resourcename.resourcegroup" (e.g.
+"pod.legacy.k8s.io").
+
+Note that we control access to resources, but not to individual metrics on
+resources (so someone with access to metrics on pod "foo" has access to
+all metrics about "foo", and there's no way to allow only access to, say,
+"cpu/usage" on "foo").
+
+#### Historical Queries #####
+
+**Verb**: get / **Subresource**: historical-metrics
+
+Namespace and Pod requests map to their respective resources.  Requests to
+get container metrics would map a request to get metrics on the
+corresponding pod.  Requests for node and system container metrics would
+simply map to cluster permissions to get historical-metrics on nodes.
+
+Getting by Pod UID would require permissions to get historical metrics on
+pods in any namespace, since we cannot tell namespace from the pod UID.
+
+Geting pods by a list would require making multiple requests, although for
+the sake of performance, it may be desirable to just merge them into
+`namespace=$NS, resource=pod, name=""` request (and, similarly to below
+with push metrics, not fall back on individual requests if the review
+fails in order to avoid added stress on Heapster and the cluster).
+
+##### Examples #####
+
+```
+GET /api/v1/historical/namespaces/somens/pods/somepod/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens
+        resource: "pod.legacy.k8s.io"
+        subresource: "historical-metrics"
+        name: somepod
+    user: $USERNAME
+```
+
+```
+GET /api/v1/historical/namespaces/somens/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens
+        resource: "namespace.legacy.k8s.io"
+        subresource: "historical-metrics"
+        name: somens
+    user: $USERNAME
+```
+
+```
+GET /api/v1/historical/namespaces/somens/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens
+        resource: "namespace.legacy.k8s.io"
+        subresource: "historical-metrics"
+        name: somens
+    user: $USERNAME
+
+```
+
+```
+GET /api/v1/historical/pod-id/ABCD-EFGH-IJKL-MNOP/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        resource: "pod.legacy.k8s.io"
+        namespace: "*"
+        subresource: "historical-metrics"
+    user: $USERNAME
+```
+
+```
+GET /api/v1/historical/nodes/somenode/containers/foobard/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        resource: "node.legacy.k8s.io"
+        subresource: "historical-metrics"
+        name: somenode
+    user: $USERNAME
+
+```
+
+#### Model Queries ####
+
+**Verb**: get / **Subresource**: metrics
+
+These function more or less identically to historical queries, except that
+they don't have to worry about Pod UIDs.
+
+##### Examples #####
+
+```
+GET /api/v1/model/namespaces/somens/pods/somepod/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens
+        resource: "pod.legacy.k8s.io"
+        subresource: "metrics"
+        name: somepod
+    user: $USERNAME
+```
+
+#### Push Creation ####
+
+**Verb**: create / **Subresource**: metrics, unprefixed-metrics
+
+Push metrics has a concept of prefixed vs unprefixed metrics.  Prefixed
+metrics work as long as you're pushing metrics describe a particular
+namespace or objects therein, and all pushed metrics become prefixed by
+the described namespace's name.  Unprefixed metrics can have any prefix,
+and are thus a more privileged operation, and are intended to be used when
+pushing metrics describing multiple namespaces.  Prefixed an unprefixed
+metrics will be represented by different endpoints in the push metrics
+API.
+
+The metrics subresource indicates that a user is allowed to push prefixed
+metrics. The unprefixed-metrics subresource indicates that a user is
+allowed to push unprefixed metrics. Core system metrics can never be
+overriden, however.
+
+In order to determine the resources to use for push metrics, the push
+handler will first determine the full set of resources involved, and will
+then aggregate according to the following rules:
+
+- a request containing only metrics for a single object (e.g. pods,
+  services, etc) will turn into a query against that resource (containers
+  in a pod are considered equivalent to the pod here).
+
+- a request involving multiple objects of a given type (e.g. multiple
+  pods) in a namespace will yield a review for `namespace=$NS,
+  resource=$RES, name=""` for that resource.
+
+- metrics describing a namespace will yield a review against that
+  namespace.
+
+- metrics describing nodes or system containers on a node will yield
+  a review against each node in question.
+
+Additionally it may be valuable to simply detect when a large number of
+namespaces are in play, and combine that all into a single cluster-wide
+review against `namespace="*"` (for cases when a "system" component is
+pushing metrics against all namespaces).  The system will not retry
+against the resources individually if this review fails (in order to
+prevent DoSing the cluster or Heapster via many reviews from a single
+push).
+
+##### Examples #####
+
+```
+http_requests{namespace="somens",pod="somepod"} 8675
+gopher_requests{namespace="somens",pod="somepod"} 2
+
+---
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: create
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens
+        resource: "pod.legacy.k8s.io"
+        subresource: metrics
+        name: somepod
+```
+
+```
+gopher_requests{namespace="somens1",pod="somepod1"} 2
+gopher_requests{namespace="somens1",pod="somepod2"} 3
+gopher_requests{namespace="somens2",pod="somepod1"} 4
+gopher_requests{namespace="somens2",pod="somepod2"} 5
+
+http_requests{namespace="somens1",service="somesvc"} 8675
+
+---
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: create
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens1
+        resource: "pod.legacy.k8s.io"
+        subresource: metrics
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: create
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens2
+        resource: "pod.legacy.k8s.io"
+        subresource: metrics
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: create
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens1
+        resource: "service.legacy.k8s.io"
+        subresource: metrics
+```
+
+```
+queue_length{namespace="someapp1"} 10
+
+---
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: create
+        group: "heapster.k8s.io"
+        version: "v1"
+        namespace: somens
+        resource: "namespace.legacy.k8s.io"
+        subresource: metrics
+        name: someapp1
+```
+
+```
+bits_flipped{node="node1",container="somecont1"} 101
+bits_flipped{node="node1",container="somecont2"} 110
+
+
+---
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: create
+        group: "heapster.k8s.io"
+        version: "v1"
+        resource: "node.legacy.k8s.io"
+        subresource: metrics
+        name: node1
+```
+
+Implementation
+--------------
+
+The authenticator would be inserted as a filter in the go-restful handlers,
+which would then inject authentication information into the request (via
+go-restful request attributes).  Then, each of the different handlers would
+be responsible for fetching the authenticator's user info from the attribute,
+invoking the authorizer, and terminating the request if the authorizer returns
+an error.  For instance, the model handlers can simply construct the needed
+information from the route URL, while the push handler might extract the metric
+names, keys, and prefix, and then call the authorizer with that information.


DirectXMan12:
cc @ncdc @piosz @mwielgus @jszczepkowski

DirectXMan12:
This is a rough draft, but I wanted to get the ideas out there so that debate on the general design could begin.

ncdc:
@pmorie

smarterclayton:
Does this need sign off from sig auth?

ncdc:
Can't hurt. cc @bgrant0607 @kubernetes/sig-auth

erictune:
Seems reasonable.  
Allowing a pod to only push metrics that are for that Pod or Namespace seems pretty useful, and a common thing to want to do.


Some areas I that would be interesting to think about are:

1. should a Container that is pushing metrics use a credential that identifies it as being in a namespace, having a service account, having a certain pod name, or being a particular container of a particular pod?
1. how do users want to aggregate metrics?  Presumably sometimes by Service or ReplicaSet that the pod is part of.  But, then how does we join pod-provided metrics with the Service or ReplicaSet dimensions?  By endpoints?  Or by labels?  If labels, how are those labels validated as being the right labels for the pusher?  
1. If namespace-level authentication is sufficient, then the pushing program could present its service account JWT from which namespace can be inferred.

deads2k:
For general authorization, we've said the boundary is at the namespace level. I think that's probably sufficient since it will contain the damage to a scope where the person affected has the power to fix the situation.  Basically, the project admin shouldn't run pods that mess with his other pods' metrics in the same namespace.

I'm less certain about using the SA credentials, but its probably just a lack of familiarity with set up.  Using the SA tokens would allow heapster (and by side effect anyone who can masquerade as heapster or has edit level access to the heaspter namespace) direct access to credentials that can be used against the API server.  When I set up heapster, is it configured as a tightly controlled component where this sort of interaction is safe?  You'd have to require SSL and the serving cert should be signed in way that allows the heapster callers to validate it.

DirectXMan12:
I chatted with @deads2k earlier to day, and he brought up the issue of the API server HTTP service proxy, which would not work with this proposal out of the box -- the proxy currently strips off authentication information (e.g. the bearer token).

To solve this, from the Heapster end, you'd have to explicitly not authorize the proxy user by itself to do anything (since you have no idea which user the request came from, so you can't properly make decisions).

In order to get the proxy to continue to work properly, we'd probably have to change the proxy itself a bit.  @deads2k and @liggitt  suggested that we could have have the proxy accept credentials in the `Proxy-Authenticate` header (like squid) instead, and simply pass through existing credentials (new clients would have to do a version check to figure out that they were talking to an old server).

liggitt:
> we'd probably have to change the proxy itself a bit

worse than that, it would mean changing the auth layer to authenticate requests to proxy subresources differently than the rest of the API (with fallback to support older clients)

it would also mean changing clients to authenticate requests to proxy subresources differently than the rest of the API (with fallback to support older servers, hoping the older server isn't passing through credentials intended for the API to the backend pods)

deads2k:
> worse than that, it would mean changing the auth layer to authenticate requests to proxy subresources differently than the rest of the API (with fallback to support older clients)

This isn't a particularly difficult thing to do.



> it would also mean changing clients to authenticate requests to proxy subresources differently than the rest of the API (with fallback to support older servers, hoping the older server isn't passing through credentials intended for the API to the backend pods)

Yes, it would.  However, it would allow token based authentication for services through the proxy in a consistent way.

DirectXMan12:
@liggitt the alternative is to just say "if you want to have any sort of authorization rules, mark the service proxy user as untrusted by Heapster, and always go through the service IP/service DNS name directly, or through a Route/Ingress/what have you".  This would mean the HPA would have to go through the service IP (or DNS name) directly (which seems reasonable to me), but I'm not sure what we'd do about kubernetes/kubernetes#28844 (having an `oc` command require a route seems suboptimal).  I feel a bit like we're between a rock and a hard place here -- we're going to have to make some use case more difficult.

@deads2k re heapster setup, Heapster supports serving over HTTPS, so it would be quite possible to have heapster have a verifiable identify (and is recommended, IIRC).

In OpenShift, we do something similar with Hawkular, where Hawkular authenticates and authorizes in a single step by performing a SubjectAccessReview against OpenShift using the user's Authorization header, AFAICT.

Which brings me to a question: we'd eventually probably want to write something similar to what Hawkular does as an option for Heapster authentication/authorization.  In this case, the "authenticate" step would effectively just have to return an empty user, and pass the bearer token to the authorizer.  This could potentially be done via the `GetExtra` method on the `auth.User` interface, but it seems a little hacky, so I was wondering if people had thoughts on that (e.g. we could use an interface that either returned `auth.User` or an indication that the authorizer should handle things).

deads2k:
> In this case, the "authenticate" step would effectively just have to return an empty user, and pass the bearer token to the authorizer.

See https://github.com/kubernetes/kubernetes/pull/28788 for webhook authentication endpoint in the API server and https://github.com/kubernetes/kubernetes/pull/20573 for kube SAR.

DirectXMan12:
@deads2k neat-o!  The webhook auth makes my above question a non-problem in the future, it looks like.

DirectXMan12:
@deads2k did a bit of work, and found an issue with the proxy change idea (see openshift/origin#9826).  So, I think, for the time being, the answer has to basically be "if you want fine-grained authorization, block the system:master-proxy user, and have things connect through the Heapster service IP/Route/Ingress".  If you just want to use the existing setup (client certs, no fine-grained authorization), then you can continue to use the master service proxy with the new interfaces.

Does that seem reasonable to people?

erictune:
IIUC, webhook auth only supports token, and not client cert (which a fair number of people use).

liggitt:
Correct, because client cert info is not transitive. Something similar could be done by verifying the client cert's signature, then remotely checking the client cert's subject's access.

deads2k:
> IIUC, webhook auth only supports token, and not client cert (which a fair number of people use).

Yes, but you could still want to support both: client certs and/or API tokens.

DirectXMan12:
> Something similar could be done by verifying the client cert's signature, then remotely checking the client cert's subject's access.

Yeah, for the client certs, you could just use the current implementation (the right CA gets injected into Heapster, and you just verify the cert, and use the CN).

Supporting both (and using SubjectAccessReview for authz) would look like
` --authn=cert:/path/to/ca.ct --authn=kubetoken:https://kubernetes.default.svc?someoption=something --authz=sar:https://kubernetes.default.svc?someoption=something`

Then, if the request had client certs, it would try verifying the signature on the client certs, and using the CN from the client cert as the user name (using the same code as today).  If the request did not, it would look for the token, check the token against kubernetes, and use the returned user info.

piosz:
@fgrzadkowski

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+authentication.
+
+Requirements
+------------
+
+Heapster should be able to separate authentication from authorization, and
+the authorization interface should be flexible enough to support authorization
+implementations that are able to make the following distinctions:
+
+- containing queries to a particular namespace: we should be able to only allow
```
erictune: Are you trying to grow Heapster into a robust multi-tenant system?  Or is there a narrower motivation for this?

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+  pods to query for metrics within their own namespace.
+
+- separating querying from pushing: we should allow certain users to push, and
+  certain users to query, but not necessarily tie the two together
+
+- allowing different push scope permissions: we should be able to specify that
+  some users are only allowed to push metrics for the pods within their own
+  RC or namespace, while others should be allowed to push for any object in
+  the cluster.
```
erictune: How important is "own RC".  That is a complex thing to track.

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+
+Proposed Solution
+-----------------
+
+Like in Kubernetes, authentication and authorization should be split into
+separate interfaces.  Each would be configured using a URL flag at the command
```
erictune: You say URL.  Do you want to support file-based as well?

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+)
+
+type Authorizer interface {
+    // AuthorizeCreation authorizes creation of metrics with the given prefix
+    // (if any) and the given names at the given keys.
+    AuthorizeCreation(user kauthuser.Info, keys []MetricKey, prefix string, metricNames []string) error 
+
+    // AuthorizeQuery authorizes querying of metrics at the given keys with the
+    // given names from the given source.
+    AuthorizeQuery(user kauthuser.Info, source HeapsterQuerySource, keys []MetricKey, metricNames []string) error
```
erictune: Is there a way to look at Creation and Query as POST and GET or "create" and "get", like in RBAC, so that and RBAC rule could express Heapster authorization, and so that for example, in the future, an RBAC role for a pod's service-account could encompass not only the ability to access the K8s API if needed, but the ability to push metrics?

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+The authorizer may not assume a relationship between keys and metric names --
+it must assume all metric names given may be pushed for all metric keys (see
+below for the rationale for passing metric names).
+
+The authorizer will then take the given information, and return an error if
+the given user is not authorized to perform the given query.
+
+For creation of metrics, `AuthorizeCreation` takes much the same parameters
+as `AuthorizeQuery` (except for the `source` parameter), with the addition
+of the `prefix` parameter.  If the prefix parameter is given, when checking
```
erictune: Why doesn't the caller do the prepending?

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+metrics only for their own namespace, or even for their own pod, while others
+may be able to push metrics for any key in Heapster.
+
+Passing in metric names allows the authorizer to make complex decisions about
+which users are allowed to push all which metrics.  For instance, an authorizer
+could restrict certain users to pushing metrics with a certain prefix, while
+allowing another user to push metrics without a prefix, as long as they were
+not one of the built-in metrics.  On the query side, this could allow
+metric-level restrictions on queries, such as only allowing certain users to
+read a certain set of custom metrics, but not have access to built-in metrics.
```
erictune: I could not discern any concrete use case from the above paragraph.

docs/proposals/auth.md
DirectXMan12: As in placing the configuration in a separate file?  While I'm not necessarily against that, I suspect that most of the auth mechanisms would work fine with just a "URL" configuration (note, this doesn't necessarily mean that the authentication or authorization is based on a remote HTTP call -- Heapster just uses a quasi-URL scheme for specifying configuration of sinks and sources, so this would probably work fine here as well).

docs/proposals/auth.md
erictune: And I think dropping metric-names from the authorization model will make it easier to make it match the RBAC model.

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+the authorizer should treat the request as if the user were requesting access
+to all names prefixed with the prefix given by `prefix`.
+
+To replicate the current client certificate mechanism, an implementation would
+simply accept any request (optionally checking the user name against a list of
+allowed users, as in the current setup).
+
+A system based on OpenShift/Kube authorization might generate a
+`SubjectAccessReview` request against an OpenShift or Kube server (e.g.
+mapping ability to `GET` pods to the ability to query the model).
```
erictune: If just Heapster does this, it is not so bad maybe, but if lots of peripheral systems to this, it makes the meaning of "GET pods" encompass a lot of things, which makes it hard for someone to understand this permission.  

Once you have a fairly general IAM system like RBAC, you might as well express the permissions within the RBAC framework.  And then extend the default roles to have the appropriate permissions.

This also allows you to show a users permissions explicitly in a clear way in a CLI/GUI.  (e.g. when you list "roleFoo" permissions, you see both "GET pods in this namespace" and "GET modelMetrics in this namespace" and "GET historicalMetrics in this namespace".

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+`--authn=clientca:/etc/heapster/ca.crt`.
+
+### Authorization ###
+
+For authorization, a system tailored to Heapster directly would be more useful:
+
+```go
+// This is basically just core.HistoricalKey from the Oldtimer PR,
+// renamed to be more generic
+type MetricKey struct {
```
erictune: Do you really need all this information?  Excessive information makes it hard to cache authorization decisions, and makes it harder to integrate with a general authorization system like RBAC.

erictune:
I like the overall direction and the parts about authentication.  I'd suggest more thought about concrete Authorization use cases, and what is the minimal amount of information that crosses the interface in order to make authorization decisions for those use cases.

docs/proposals/auth.md
erictune: Thanks for clarification.  So, it sounds like you would be able to reuse, for example, the local-file-based token authentication module from Kube, but you would use a pseudo-URL to name that module via a flag, e.g. `--authn=password:localfile:/etc/heapster/passwords.csv`.

Sounds fine.

docs/proposals/auth.md
```diff
@@ -0,0 +1,150 @@
...
+  RC or namespace, while others should be allowed to push for any object in
+  the cluster.
+
+Proposed Solution
+-----------------
+
+Like in Kubernetes, authentication and authorization should be split into
+separate interfaces.  Each would be configured using a URL flag at the command
+line, similarly to how sinks and sources are currently configured.  Adding
+multiple authentication drivers would be supported (for example, to support
```
erictune: Define how it works with multiple:  presumably, each is tried in order, and success if any succeeds, with short circuit?

docs/proposals/auth.md
DirectXMan12: that's just an example.  It's difficult to track in the current setup, but I can envision a future in which we can somehow identify the pod pushing, and I'd like to be able not have to change the interface of the auth mechanism (this is mainly the motivation for including the whole "key" to the metrics, and not just the namespace -- it allows us the option to make more complex decisions like this, should we so choose).

docs/proposals/auth.md
DirectXMan12: Somewhat.  The motivation comes from several related issues:

1. We want to be able to restrict who can push metrics, read recent metrics (via the model), and read historical metrics (via Oldtimer) -- it's desirable to allow different sets of people to have permissions for those things.
2. We *potentially* want to be able to restrict who can push metrics associated with certain "objects" (e.g. you might want some to be able to push metrics for any cluster objects, while other might be only allowed to push metrics in their own namespace).
3. We seen concern over the fact that you cannot restrict reading of metrics to a certain scope for different users.  A mechanism to restrict users to reading things only in their own namespace, for example, would allow Heapster access to be more "safely" granted to different applications and users on the cluster.

docs/proposals/auth.md
DirectXMan12: Ack.

docs/proposals/auth.md
DirectXMan12: The decision is mostly on the "prefix" vs "no prefix" -- it could allow for a extension to push metrics where certain applications are trusted to push metrics without the "custom/" prefix, for example (which could be desirable for certain users).  I think it probably makes sense just to drop down to just having the prefix specified.

docs/proposals/auth.md
DirectXMan12: It's mainly for convenience of the auth mechanism implementation.  A common case would probably be "is the user authorized to push metrics prefixed with the given prefix" (e.g. generally, users are only allowed to push metrics prefixed with their name, but some users might not have that restriction).  If you pass the prefix separately, you can do 1 string compare.  If you don't pass the prefix in, you have to run through the whole list of metrics separately.

docs/proposals/auth.md
DirectXMan12: Partially -- create is POST, but the URL doesn't say anything about the particular metrics, resources, etc involved (and multiple metrics and resources may be involved).

Query is GET, and the URL does say something about the resources involved.

However, that seems like a detail for the particular auth mechanism.  An authorization mechanism like the current one (just using a list of user names) probably doesn't care about GET vs POST, and just wants to know which action (create vs query model vs query historical) is being performed, so it can check the list of users allowed.

docs/proposals/auth.md
DirectXMan12: Exactly.

docs/proposals/auth.md
DirectXMan12: Yeah, that was the plan (that's how the Kube auth works, right?)

docs/proposals/auth.md
erictune: So, you have both kubernetes namespace, and "username prefix" as namespacing mechanisms.  Sorta confusing.

erictune:
I'm not sure what our overall goals are for how heapster fits into a Kube system. 

If heapster is going to remain an optional, replace-able component in a Kube cluster, that just happens to share some authentication code, then I think this proposal is just fine.

However, if heapster is going to look more and more like just another part of the Kube api to the point where the heapster API becomes an expected part of any Kube cluster, then I feel somewhat strongly that the Heapster authz needs to conform more to the Kube RBAC style (hence comments in this PR.).

These other PRs:
 - https://github.com/kubernetes/kubernetes/pull/26858 and   
 - and especially https://github.com/kubernetes/kubernetes/pull/28844
make me think it is more of the latter.

piosz:
@erictune Heapster serves Metrics API and some time in the future (for 1.5?) it will be a kind of API server (details TBC), so the latter is true. So that eventually the authorization stuff in Heapster should be the same as in Kubernetes.

@DirectXMan12 please correct me if I'm wrong: the solution you propose is a temporary solution to make the effort of pushing custom metrics to Heapster possible rather than our final goal.

DirectXMan12:
@piosz @erictune the goal was to propose an interface that could be coded against currently that would make it so that we could quickly reimplement the current system (client certs and user list), and then later implement a system based on RBAC/SAR (or what have you) without having to rewrite the code which uses the interface.

Basically, the goal is to not have to rewrite a bunch of the push metrics code when we implement a more robust authorization system, but to also not have to make *having* the more robust system implemented a requirement to merging push metrics.

docs/proposals/auth.md
DirectXMan12: Note here that not all information is present for all key types -- for pod queries for instance, either `NamespaceName` and `PodName` are present, or `PodID` is present.  `Node` is used for node-level metrics and system-container metrics (which also use `ContainerName`).

DirectXMan12:
> conform more to the Kube RBAC style

Putting aside the "not needing RBAC auth as a prerequisite for merging push metrics" issue, there's also the issue that that there's not necessarily a direct mapping between a single action and a single (verb, resource) pair like in Kubernetes (see below).  IMO, we should make the methods which the various parts of Heapster call into speak in Heapster terminology, and then whatever is behind them can do some conversion into the appropriate Kubernetes terminology.

For instance, while most queries (both model and historical) can be expressed directly as a (verb, resource) pair (i.e. GET, with the subresource representing either model or historical), when you get to lists and pod IDs the mapping is slightly less straightforward (although still pretty clear -- you'd probably just map lists to GET ALL IN NAMESPACE, and since it's not clear what name and namespace a UID belonged to, fetching by UID would probably just be mapped to GET ALL IN CLUSTER).

Once you get the push metrics, the mappings break down further: push metrics are intended to allow bulk submission of metrics, so in a single push you could have: metrics on the namespace, metrics on pods, metrics on containers, and metrics on the cluster.  If all the metrics are in a particular namespace, you might be able to notice that and make a more direct query, and fall back to POST ALL IN CLUSTER for everything else.

The various Heapster components, however, should not have to know anything about these mappings.  They should just be able to say "here are the affected Heapster keys, can I preform creation/model get/historical get" and get a response.

DirectXMan12:
That's not to say that we shouldn't have RBAC mappings for Heapster resources at some point (I think we should have them), I just don't think the internal Heapster components should specifically have to think about the mappings, unless the RBAC system can speak more in terms of Heapster terminology.

DirectXMan12:
@liggitt @deads2k @erictune thoughts?

erictune:
If Heapster is going to serve a Kubernetes-style API soon, as @piosz says, then Heapster will need to adjust its concepts to fit the Kube authorization model.  

If Heapster is going to forever remain a separate, optional, system, then it can keep its own concepts (as @DirectXMan12 suggests in https://github.com/kubernetes/heapster/pull/1213#issuecomment-233412472).  

I haven't seen a proposal on what the future of Heapster is within Kubernetes.  
I honestly don't know what the plan is, but I think there should be a firm decision, involving the Heapster and Kubernetes communities -- a decision on whether heapster is required, and if it serves an API that is required for essential kubernetes APIs to work right.   Once that is decided, we will know what to do about heapster authorization.

lavalamp:
> So that eventually the authorization stuff in Heapster should be the same as in Kubernetes.

Right, we want to expose the two behind a single access point, so they have to have compatible security models.

> Basically, the goal is to not have to rewrite a bunch of the push metrics code when we implement a more robust authorization system, but to also not have to make having the more robust system implemented a requirement to merging push metrics.

This seems like a reasonable goal, however if the timeframe is short, is it necessary? Why not just start out with the kube auth concepts?

erictune:
subjectaccessreview has these fields:

- subject:
   - user making a request, (service account if called from a pod), 
  - the user's group memberships 
- "verb" (create, delete, etc).
- object:
  - namespace of the object
  - name of the object (exact match, no pattern match)
  - kind of the object
  - subresource

Heapster wants to authorize their requests based on this set of attributes:
- NodeName
- Namespace
- PodName
- ContainerName
- PodId string
- QueryType (historical or model)
-  metric name (prefix match)
- ObjectType in {MetricSetTypePodContainer, MetricSetTypePod, core.MetricSetTypeNamespace}

How can we fit the heapster attributes into the subjectaccessreview object?

@lavalamp @ericchiang @deads2k

erictune:
For the authentication, is the idea to set up a completely separate CA for heapster, and a separate distribution path for heapster-CA-signed-client-certs, which is parallel to but independent of the service-account and service-account-token framework?

deads2k:
> How can we fit the heapster attributes into the subjectaccessreview object?

In openshift we've used synthetic access checks where we would model those concepts something like this:

 1. historical query -> verb: get
 1. model query -> verb: get, subresource: model
 1. creation -> verb: create

Assuming you model the things you're getting and updating as resources, that allows clean ACL for your verbs

I'm less sure about the resources
 1. namespace name -> this seems to map directly to our existing namespace concept, so in the attributes it would be namespace=namespace.Name
 1. PodName/PodID - I don't understand the distinction between these.  We key based on PodName now.  Are we worried about re-using a name and an old credential having rights to update it?  Drawing an ACL boundary on namespace seems to make sense, so this distinction seems to matter less.  I would expect apiGroup: heapster, resource: pod, name: name
 1. ContainerName - I'm not sure I see the need to acl down to this level.  Why would I want to subdivide my access based on containers inside pods?
 1 NodeName - apigroup: heapster, resource: nodes ?

DirectXMan12:
> namespace name -> this seems to map directly to our existing namespace concept, so in the attributes it would be namespace=namespace.Name

Yeah, that's probably the easiest one ;-). 

> PodName/PodID - I don't understand the distinction between these. We key based on PodName now. Are we worried about re-using a name and an old credential having rights to update it? Drawing an ACL boundary on namespace seems to make sense, so this distinction seems to matter less. I would expect apiGroup: heapster, resource: pod, name: name

PodID is the Pod's UUID.  It's used in the historical API to allow users to disambiguate between several historical pods with the same name.  Queries with UUIDs have no associated namespace/name information, so I'd imaging they'd have to be a more privileged operation (e.g. `GET POD METRICS in namespace "*"`)

> ContainerName - I'm not sure I see the need to acl down to this level. Why would I want to subdivide my access based on containers inside pods? 1 NodeName - apigroup: heapster, resource: nodes ?

That information was mainly intended to allow ACLing on system containers, which are identified by node name and pod name.  In general, I'd imagine that you'd probably just want system container metrics to be a yes/no switch, although I'm not entirely certain how you'd represent that.  The field is mostly there for completeness -- pods are ACLed on NAMESPACE and POD NAME, while the "corresponding" attributes for system containers seemed to be NODE and CONTAINER NAME.

The biggest wrinkle that I see is push metrics -- from what I've seen of the ACL system (and it's entirely possible I've missed something), we basically need to decide how we want to break up the requests for push metrics.  How do we deal with multi-namespace requests (do we make multiple requests, do we just say any number of namespaces greater than one requires permissions in to push in any namespace (which seems like it could yield confusing results)), what does the prefix map to, etc?

deads2k:
Ok, so if we model resources like this:
 1. group=heapsterGroup, namespace=normal, resource=pod, name=podname, subresource=model, for most requests, then we're able to use the get/create/update verbs like we'd expect
 1. group=heapsterGroup, namespace="", resource=pod, name="", for cases where you're using pod UID, that would make it a privileged operation
 1. group=heapsterGroup, namespace="", resource=container, name=name, for cases where you're using container without a pod, that would make it a privileged operation

Is there a case where one group of people can see containers on some nodes and not others?

deads2k:
> How do we deal with multi-namespace requests (do we make multiple requests, do we just say any number of namespaces greater than one requires permissions in to push in any namespace (which seems like it could yield confusing results)), what does the prefix map to, etc?

In openshift, when an operation requires multiple permissions, we check for permission to perform each action.  For example tagging an image in another namespace requires permission to tag in ns1 and permission to pull the image from ns2.



> what does the prefix map to, etc?

Hmm... we need to subdivide at that level of granularity?  I could be done via subresource, but that would be unpleasant for all concerned.

DirectXMan12:
> Hmm... we need to subdivide at that level of granularity? I could be done via subresource, but that would be unpleasant for all concerned.

I imagine in most cases it would be something like "is this user allowed to push with no prefix, is this user allowed to push with a custom prefix (maybe), or is this user allowed to push with a prefix equivalent to their user name" (so, effectively, one of two or three options).

DirectXMan12:
> Is there a case where one group of people can see containers on some nodes and not others?

Is there a case where a node might only be allowed to see resources for its own node?  That's probably more fine-grained than you want, right?  Otherwise, I think that looks reasonable.

DirectXMan12:
> For example tagging an image in another namespace requires permission to tag in ns1 and permission to pull the image from ns2.

That might be a *lot* of requests, conceivably.  Is it a reasonable operation to first check for "ANY NAMESPACE" permissions on more that one request, and then if you don't have that, make the individual requests (but have a cap of some sort)?

deads2k:
> That might be a lot of requests, conceivably. Is it a reasonable operation to first check for "ANY NAMESPACE" permissions on more that one request, and then if you don't have that, make the individual requests (but have a cap of some sort)?

Eh, you could, but generally you're running an LRU with a TTL to keep from making remote calls.  Since kube and openshift now agree on the Authorize interface, I think you can just lift our layers.

DirectXMan12:
@deads2k @erictune I've updated the PR to just work with Kube SAR and TokenReviews+client certs.  Does that look closer to what you're looking for?

docs/proposals/auth.md
erictune: Number 3 sounds like "yes" to my question about multi-tenancy.

Say you add authorization by namespace.
Then, the next thing that happens is someone sends lots of high-cpu queries and makes the system slow for everyone else.  Or sends to many metrics, and  uses up all the storage.  IF heapster was not designed to be multitenant from the start, and I don't think it was, this is going to be whack-a-mole for a long time. 

The alternative is to run a heapster per namespace.  This solves both authz and qos problems.  If you have a free tier where you can't afford to run heapster per namespace, you can use a shared instance, but restrict access to it.

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+concepts, and then performing `SubjectAccessReview` requests against the
+API server configured in `--auth-apiserver`.
+
+Concepts will map as such:
+
+#### Historical Queries #####
+
+**Verb**: get / **Subresource**: historical-metrics
+
+Namespace and Pod requests map to their respective resources.  Requests to
```
erictune: Let me test my understanding.   A request for $PODNAME metrics in namespace $NS would do a  SAR like this:
```yaml
kind: subjectAccessReview
apiVersion: authorization.k8s.io/v1beta
spec:
  resourceAttributes:
    namespace: $NS
    verb: get
    group: ""
    version: "v1"
    subresource: "historical-metrics"
    resource: "pod"
    name: $PODNAME
  subject:
    user: directxman12
```

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+get container metrics would map a request to get metrics on the
+corresponding pod.  Requests for node and system container metrics would
+simply map to cluster permissions to get historical-metrics on nodes.
+
+Getting by Pod UID would require permissions to get historical metrics on
+pods in any namespace, since we cannot tell namespace from the pod UID.
```
erictune: Can you not extend heapster to remember the namespace when it stores stats by pod UID?
hmm.  I guess that should be the namespace UID?

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+
+Geting pods by a list would require making multiple requests, although for
+the sake of performance, it may be desirable to just merge them into
+`namespace=$NS, resource=pod, name=""` request.
```
erictune: Would you fall back to individual requests if that failed, or just reject?

docs/proposals/auth.md
DirectXMan12: > IF heapster was not designed to be multitenant from the start, and I don't think it was, this is going to be whack-a-mole for a long time. 

> The alternative is to run a heapster per namespace. This solves both authz and qos problems. If you have a free tier where you can't afford to run heapster per namespace, you can use a shared instance, but restrict access to it.

Does the Heapster multitenancy story need work?  Definitely.  Does that mean we shouldn't work towards it?  No. 

It's not just a "free-tier" problem -- if you have a lot of paying users, or you're running a company wide cluster with a lot of different consumers, etc, it's going to be a problem.  With Heapster-per-namespace as the auth solution, you're basically say "if you're running a small cluster, you get fine grained auth, otherwise ¯\\_(ツ)_/¯".

If you have a bunch of namespaces and heapster-per-namespace, get problems of:

- load (kubelet has to serve *many* more HTTP requests, Heapster sinks have to be able to handle *many* more writers)
- extra work for Heapster consumers (pushers now have to divide their metrics by namespace, discover the Heapster in each namespace, and make a request to each Heapster, and the HPA now has to know how to, and be permitted to, talk to many different Heapsters)
- user-friendliness (there's now a RC in every namespace not under each user's control), 
- admin-friendliness (you now need a component to auto-provision each namespace with a Heapster, provide secrets to talk to the backend, etc)

It also entirely sidesteps the problem of making auth easier for the user -- plugging into something like the Kube RBAC system makes it easier to manage your permissions, and easier for users to auth -- you don't have to have a side-channel for distributing certificates, etc.

docs/proposals/auth.md
DirectXMan12: Yeah, that's more or less what I was envisioning.  I can add a couple of examples to the proposal.

docs/proposals/auth.md
DirectXMan12: When you're querying historical metrics, there's no guarantee that you'll have that info (you may have been restarted, etc).  It's possible for model metrics, but the model doesn't support UID queries.

docs/proposals/auth.md
DirectXMan12: probably just reject -- we don't want someone to be able to DoS the cluster and or Heapster like that in a case with a *lot* of namespaces.

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+**Verb**: get / **Subresource**: metrics
+
+These function more or less identically to historical queries, except that
+they don't have to worry about Pod UIDs.
+
+#### Push Creation ####
+
+**Verb**: create / **Subresource**: metrics, unprefixed-metrics
+
+The metrics subresource indicates that a user is allowed to push metrics
```
erictune: Say I have an nginx pod, and it wants to push some metrics about itself, such as number of requests handled since starting.   Say that the pod has a credential with username  `$USERNAME`.

Is the recommended pattern:

1. the `nginx` pushes metrics prefixed with that username, like `$USERNAME-total-requests`.  Also, $USERNAME need to have a role binding which gives it `create metrics` permission. Or is it...
2.  the `nginx` pod  pushes unprefixed metrics, such as `total-requests`.  Also  $USERNAME needs to have a role binding which gives it `create unprefixed-metrics` permission.

docs/proposals/auth.md
deads2k: > The alternative is to run a heapster per namespace. This solves both authz and qos problems. If you have a free tier where you can't afford to run heapster per namespace, you can use a shared instance, but restrict access to it.

I think that heapster per-namespace is going to have similar requirements about separating readers and writers.  The kube `TokenAccessReview`, `SubjectAccessReview` and Impersonation features were intended to enable use-cases like this one.  I think this will be good experience for us as we try to move those APIs out of the beta stage.

Also, this feature wouldn't preclude per-namespace heapster scenarios if we want to try them out.

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+- allowing different push scope permissions: we should be able to specify that
+  some users are only allowed to push metrics for the pods within their own
+  namespace (or given sufficient information in the future, possibly their own
+  RC), while others should be allowed to push for any object in the cluster.
+
+Proposed Solution
+-----------------
+
+Heapster will adopt an authentication and authorization model based around
+querying the Kubernetes model.  Authentication will support either client
```
deads2k: You'll have to terminate client certs yourself or support the impersonation headers from wherever they are terminated.

docs/proposals/auth.md
deads2k: > Yeah, that's more or less what I was envisioning. I can add a couple of examples to the proposal.

Please add one.

docs/proposals/auth.md
deads2k: > When you're querying historical metrics, there's no guarantee that you'll have that info (you may have been restarted, etc). It's possible for model metrics, but the model doesn't support UID queries.

Are there some cases where you will have the information?  Is it difficult to use it when you have it?

docs/proposals/auth.md
deads2k: > probably just reject -- we don't want someone to be able to DoS the cluster and or Heapster like that in a case with a lot of namespaces.

That sounds reasonable to start, but you should clarify it.

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+These function more or less identically to historical queries, except that
+they don't have to worry about Pod UIDs.
+
+#### Push Creation ####
+
+**Verb**: create / **Subresource**: metrics, unprefixed-metrics
+
+The metrics subresource indicates that a user is allowed to push metrics
+not prefixed with their username.  The unprefixed-metrics subresource
+indicates that a user is allowed to push metrics with no prefix.  Core
```
deads2k: So if `metrics` == push metrics not prefixed with username and `unprefixed-metrics` == push metrics with no prefix, is there a permission that says, "push metrics that are prefixed with my username"?

Also, how are you going to deconflict user "billy" pushing "bob-foo" and user "billy-bob" pushing "foo"?

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+- `--authn-ca=/path/to/ca.crt`
+- `--auth-apiserver=https://$KUBE`
+
+### Authorization ###
+
+Authorization will work by matching Heapster concepts to Kubernetes RBAC
+concepts, and then performing `SubjectAccessReview` requests against the
+API server configured in `--auth-apiserver`.
+
+Concepts will map as such:
```
deads2k: All these queries need to be setting an `APIGroup` of something under `heapster.k8s.io`

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+
+- a request containing only metrics for a single object (e.g. pods,
+  services, etc) will turn into a query against that resource (containers
+  in a pod are considered equivalent to the pod here).
+
+- a request involving multiple objects of a given type (e.g. multiple
+  pods) in a namespace will yield a review for `namespace=$NS,
+  resource=$RES, name=""` for that resource.
+
+- metrics describing a namespace will yield a review against that
```
deads2k: nodes and namespaces will be named?  It would be helpful to have one example each here too.

docs/proposals/auth.md
DirectXMan12: The *intended* pattern is/was `nginx` pushes against `metrics` with `total-requests`.  Heapster checks if you're permitted to push unprefixed metrics, and if not, sees that you're permitted to push prefixed metrics, and thus prepends `custom/$USERNAME/`.  Also, $USERNAME needs to have permissions allowing it to `create pod/metrics`

Alternatively, "trusted system component X" may be allowed to have `unprefixed-metrics` permissions.  "trusted system component X" pushes metrics, Heapster sees the permission for `unprefixed-metrics`, and uses the permissions as-is.  The username of "trusted system component X" needs to have permissions on to `create pod/unprefixed-metrics`.

Unfortunately, this requires two requests unless you cache the result of prefixed vs unprefixed, and assume uniformity among prefixed vs unprefixed across different resources.  I'm open to suggestions here (up to and including removing unprefixed metrics, but at the very least not have the "custom/" prefix for things intended to be cluster system components is nice...)

docs/proposals/auth.md
```diff
@@ -0,0 +1,128 @@
...
+namespaces are in play, and combine that all into a single cluster-wide
+review against `namespace="*"` (for cases when a "system" component is
+pushing metrics against all namespaces).
+
+Implementation
+--------------
+
+The authenticator would be inserted as a filter in the go-restful handlers,
+which would then inject authentication information into the request (via
+go-restful request attributes).  Then, each of the different handlers would
```
deads2k: In core kube, we actually parse the URL separately to determine the authorization attributes.  It has the advantage of being *very* obvious that you're handling every incoming request.  I think you should strongly consider doing something similar.

docs/proposals/auth.md
DirectXMan12: Yeah, the client certs stuff will require having the CA available, as I noted below, unless I'm misunderstanding you.

docs/proposals/auth.md
DirectXMan12: That won't work when Heapster gains the ability to support metrics on objects other than pods.  How do you distinguish between querying for metrics on an object in two different API groups if you can't pay attention to the SubjectAccessReview's API group?

docs/proposals/auth.md
deads2k: > Yeah, the client certs stuff will require having the CA available, as I noted below, unless I'm misunderstanding you.

Yeah, I got there after I left a comment here.  Impersonation headers may still be useful and it seems like you may have an intersection with https://github.com/kubernetes/kubernetes/pull/29714.

docs/proposals/auth.md
DirectXMan12: That would be great, except that it doesn't work for push metrics.  It could work for model and historical queries, though.

docs/proposals/auth.md
deads2k: > The intended pattern is/was nginx pushes against metrics with total-requests. Heapster checks if you're permitted to push unprefixed metrics, and if not, sees that you're permitted to push prefixed metrics, and thus prepends custom/$USERNAME/. Also, $USERNAME needs to have permissions allowing it to create pod/metrics

I don't think that pattern is a good idea.  It seems like writing to different metrics based on your current level of access will produce confusion.  Seems like you should make them say what they mean.

docs/proposals/auth.md
DirectXMan12: I replied to a similar question by @erictune above, but I'll add in some more examples.

docs/proposals/auth.md
deads2k: > That won't work when Heapster gains the ability to support metrics on objects other than pods. How do you distinguish between querying for metrics on an object in two different API groups if you can't pay attention to the SubjectAccessReview's API group?

What about `apiGroup=heapster.k8s.io`, `resource=pods.legacy.k8s.io`?  You cannot assume that there isn't an actual `pods/metrics` endpoint that needs to be ACLed in a different group.

DirectXMan12: > You cannot assume that there isn't an actual pods/metrics endpoint that needs to be ACLed in a different group.

Ah, fair point. 

> apiGroup=heapster.k8s.io, resource=pods.legacy.k8s.io

Yeah, that would work.

docs/proposals/auth.md
DirectXMan12: The push metrics implementation PR hasn't actually merged yet, so I can still change the semantics a bit there.  I'll figure something out.

docs/proposals/auth.md
DirectXMan12: It's possible (although it requires keeping a list of all pods ever seen), but it seems like it makes it less confusing to users just to blanket say "pod uid access is more privileged" vs saying "pod uid access is sometimes more privileged".

docs/proposals/auth.md
DirectXMan12: Also, apparently I had an extraneous not in that description.

docs/proposals/auth.md
DirectXMan12: ✔️

DirectXMan12:
I've added some examples and hopefully clarified a couple of points.

RE: prefixed vs unprefixed metrics, I'm imaging the metrics will look the same (i.e. submitted metrics will never have a username prefix), but the URL may end up differing slightly, although that will have to be a change to the push metrics PR.

docs/proposals/auth.md
deads2k: > It's possible (although it requires keeping a list of all pods ever seen), but it seems like it makes it less confusing to users just to blanket say "pod uid access is more privileged" vs saying "pod uid access is sometimes more privileged".

I can buy that.

docs/proposals/auth.md
```diff
@@ -0,0 +1,340 @@
...
+GET /api/v1/historical/namespaces/somens/metrics/cpu/usage_rate
+
+apiVersion: authorization/v1beta1
+kind: SubjectAccessReview
+spec:
+    resourceAttributes:
+        verb: get
+        group: "heapster.k8s.io"
+        version: "v1"
+        resource: "namespace.legacy.k8s.io"
```
deads2k: to match normal authz rules, this should have the namespace and the name set to `somens`

docs/proposals/auth.md
```diff
@@ -0,0 +1,340 @@
...
+Getting by Pod UID would require permissions to get historical metrics on
+pods in any namespace, since we cannot tell namespace from the pod UID.
+
+Geting pods by a list would require making multiple requests, although for
+the sake of performance, it may be desirable to just merge them into
+`namespace=$NS, resource=pod, name=""` request (and, similarly to below
+with push metrics, not fall back on individual requests if the review
+fails in order to avoid added stress on Heapster and the cluster).
+
+##### Examples #####
```
deads2k: These help.  Make a top level note that we're controlling access to metrics on resource X, but not subdividing ACL on individual metrics on resource X.  I think this reasonable, but its worthy of note.

DirectXMan12: ack.

DirectXMan12:
After some lengthy discussion with @deads2k (thanks!), I've tweaked the section about prefixed and unprefixed metrics, and will be updating the push metrics API PR correspondingly.

DirectXMan12:
closing in light of the new monitoring vision

PR #1228, opened 2016-07-26
Add object lists to metrics api

### LINKED ISSUE #1219: Introduce object lists to metrics api
We should have `PodMetricsList` and `NodeMetricsList` instead of returning a slice of `PodMetrics` or `NodeMetrics`.

mksalawa:
fix #1219


piosz:
LGTM. Thanks for the change!

Test failures unrelated.

PR #1229, opened 2016-07-27
Allow for labeled metric queries

DirectXMan12:
This commit allows for querying for information (both from the model and
from the Oldtimer) about labeled metrics.  This is done by passing the
`labels` query parameter when querying for metric data.  The value of
the `labels` parameter should be a comma-separated list of `key:value` pairs.
The label set provided must be an exact match for the set of "extra" labels
on the target metric.


DirectXMan12:
cc @piosz @mwielgus 

cc @burmanm depending on if/when this gets merged, you might have to implement this for #1209

DirectXMan12:
While there are not currently many labeled metrics pushed through Heapster (although, IIRC, there are a few), there may be more in the future (e.g. per-port network stats on pods and services), and push metrics also opens up a new avenue for adding more labeled metrics.

DirectXMan12:
@k8s-bot test this

mwielgus:
LGTM

PR #1231, opened 2016-07-27
Add support for metrics of all pods

### LINKED ISSUE #1223: Add support for pod metrics from all namespaces in Metrics API
We should add support for handler `apis/metrics/v1alpha1/pods` in https://github.com/kubernetes/heapster/blob/master/metrics/apis/metrics/handlers.go#L52

mksalawa:
fix #1223


piosz:
LGTM

PR #1236, opened 2016-07-27
Reduce when some of our messages get logged.

mwringe:
There are a bunch of messages that we are currently logging which could be moved to not being logged by default. This will prevent the logs from growing too much in size and will allow people to more easily parse through the logs to determine if something went wrong.

The logs about how many metrics were gathered have been moved to V(1) level instead of the default logging level.

If we don't have node level metrics, we shouldn't be filling up the logs with warning messages that we can't aggregate metrics for the non-existent metric. This has also been moved to V(1).


mwringe:
FYI, it looks like the test failure is not related to this commit.

mwielgus:
LGTM

Issue #1246, opened 2016-08-07
Add node labels to metrics with node type


andyxning:
AFAIK, currently we only show labels for metrics with type pod and pod_container. Metrics with type node is not supplied with node labels. However, this is especially useful in some situation. Below is one example. 

Kubernetes labels is designed for user defined usage. So, we can leverage label to do some mark. 

If we run too  clusters in one datacenter. we use the defacto monitoring stack(cAdvisor + heapster + InfluxDB + Grafana). we don't want to make two monitors for both clusters. However, we also want to make a distinguish between two clusters for cluster info which is aggregated from node. 

In order to make it, we use node label to mark the cluster info for one node, such as "cluster:c1" and "clster:c2". However, we can not get node labels from InfluxDB. :-(

andyxning:
Since the PR #1247 has been merged in master branch. This issue is closed. :)

PR #1247, opened 2016-08-07
add node labels for metrics with type node

andyxning:
Add node labels for metrics with type node. For more info see #1246


mwielgus:
lgtm

PR #1252, opened 2016-08-09
Bumped Kubernetes deps

piosz:


piosz:
cc @mksalawa

DirectXMan12:
Ok, so, I think I figured out why this is failing -- there are a couple of `returns` missing after calls to `WriteError`.  I suspect that something changed in go-restful that caused it to previous ignore all writes after `WriteError`, but the new version allows for writes after `WriteError`.  I've got a patch coded up (it's just a couple lines).  I'll post it in a couple minutes.

piosz:
@DirectXMan12 many thanks for the help!

DirectXMan12:
@piosz NP.  The PR above should fix it.  Sorry about that ;-)

piosz:
@k8s-bot test this

mwielgus:
LGTM

PR #1254, opened 2016-08-09
Add missing returns to the Historical API handlers

DirectXMan12:
The historical API handlers were missing return statements after
`response.WriteError` calls in a couple of places.  This does not
prevent it from functioning currently, but with #1252, it causes
issues.


piosz:
LGTM

PR #1267, opened 2016-08-22
fix error value for metrics of type node

### LINKED ISSUE #1248: Error value for type node memory metrics
according to the [storage schema](https://github.com/kubernetes/heapster/blob/master/docs/storage-schema.md), all metrics info gathered from all k8s is about running pods and system containers. 

However, for node memory metrics, `memory/limit` and `memory/request` with type `node` are aggregated from all pods which is then aggregated from all containers running in pods. For more info please see [node_aggregator](https://github.com/kubernetes/heapster/blob/master/metrics/processors/node_aggregator.go#L31-L54) and [aggregator metrics definition](https://github.com/kubernetes/heapster/blob/master/metrics/heapster.go#L104-L163)

However, for other memory metrics with type node, such as 
* `memory/major_page_faults `
* `memory/major_page_faults_rate `
* `memory/page_faults `
* `memory/page_faults_rate`
* `memory/usage`
* `memory/working_set `

uses the value gained from [runc](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/memory.go) which is actually uses memory cgroups. currently, these memory metrics with type node uses the top level memory cgroup values from 
* `memory.stat.pgfault`
* `memory.stat.pgmajfault`
* `memory.usage_in_bytes`

This strategy is wrong, i think. we should aggregate these memory metrics just like `memory/limit` and `memory/request` with aggregating from all pods.

**More Info**:
according to [cgroups hierarchy spec](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt) sec 5.2

```
# status considering hierarchy (see memory.use_hierarchy settings)
     
 hierarchical_memory_limit - # of bytes of memory limit with regard to hierarchy  
 			under which the memory cgroup is  
 hierarchical_memsw_limit - # of bytes of memory+swap limit with regard to  
 			hierarchy under which memory cgroup is.  
   
 total_<counter>		- # hierarchical version of <counter>, which in  
 			addition to the cgroup's own value includes the  
 			sum of all hierarchical children's values of  
 			<counter>, i.e. total_cache  
```

### **UPDATE**
----
The same as 
* `cpu/usage`
* `network/rx`
* `network/rx_rate`
* `network/rx_errors`
* `network/rx_errors_rate`
* `network/tx_errors`
* `network/rx_errors_rate`
* `network/tx`
* `network/tx_rate`
These metrics with type node will also need to aggregating value from all pods. :) 

If we want node level `cpu/usage` or `memory/usage` we should name them as `cpu/node_usage` or `memory/node_usage`

andyxning:
fix #1248 

this PR make metrics for type node **aggregated from pods running on it**.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1267)

### DIFF
--- a/metrics/core/metrics.go
+++ b/metrics/core/metrics.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	cadvisor "github.com/google/cadvisor/info/v1"
+	"k8s.io/heapster/metrics/util/metrics"
 )
 
 const (
@@ -28,6 +29,8 @@ const (
 var StandardMetrics = []Metric{
 	MetricUptime,
 	MetricCpuUsage,
+	MetricMemoryCache,
+	MetricMemoryRSS,
 	MetricMemoryUsage,
 	MetricMemoryWorkingSet,
 	MetricMemoryPageFaults,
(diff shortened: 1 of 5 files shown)


piosz:
closing as explained in #1248

PR #1273, opened 2016-08-30
Remove unneeded line from Makefile

gmarek:
Fix kubernetes/test-infra#471

cc @mksalawa @mwielgus @piosz

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1273)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ deps:
 	which godep || go get github.com/tools/godep
 
 build: clean deps
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go install ./...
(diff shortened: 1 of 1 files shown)


piosz:
LGTM, thanks Marek

Makefile
```diff
@@ -11,7 +11,6 @@ deps:
 	which godep || go get github.com/tools/godep
 
 build: clean deps
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go install ./...
```
ixdy: I think we still want this. now there's nothing verifying that everything in the repo builds.

gmarek: Why do we care if *everything* builds? Isn't it enough that the binary builds?

PR #1296, opened 2016-09-16
[WIP] Heapster API server with custom handlers

mksalawa:
The idea of this PR is to make Heapster use the genericapiserver library, using the custom handlers for all needed paths.

The next step would be registering the metrics as resources.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1296)

PR #1294, opened 2016-09-16
WIP Implement API server in Heapster

mksalawa:
- Basic implementation of the apiserver
- serving NodeMetrics and PodMetrics
- added /healthz endpoint
- **mocked etcd** - will be done after refactoring the main function

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1294)

PR #1301, opened 2016-09-19
Heapster API server init

mksalawa:
Add an empty API server exposed on a separate port.

Also, changed flag "port" to "heapster-port" to avoid conflict with the generic API server flag.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1301)


metrics/options/options.go
```diff
@@ -36,14 +38,19 @@ type HeapsterRunOptions struct {
...
 }
 
 func (h *HeapsterRunOptions) AddFlags(fs *pflag.FlagSet) {
+	h.ServerRunOptions.AddUniversalFlags(pflag.CommandLine)
+
 	fs.Var(&h.Sources, "source", "source(s) to watch")
 	fs.Var(&h.Sinks, "sink", "external sink(s) that receive data")
 	fs.DurationVar(&h.MetricResolution, "metric_resolution", 60*time.Second, "The resolution at which heapster will retain metrics.")
-	fs.IntVar(&h.Port, "port", 8082, "port to listen to")
+	fs.IntVar(&h.Port, "heapster-port", 8082, "port to listen to")
```
piosz: Please find a better name for the flag and extend the description.

piosz:
Looks reasonable. @nikhiljindal could you please take a look? Our goal is to refactor Heapster to reuse genericapiserver library. This is the first step which creates apiserver without any resource.

DirectXMan12:
What's the relationship between this, #1296 (which looks to be an extension of this), and #1294?

metrics/cmd/heapster-apiserver/app/server.go
```diff
@@ -0,0 +1,149 @@
...
+		OIDCUsernameClaim: s.OIDCUsernameClaim,
+		OIDCGroupsClaim:   s.OIDCGroupsClaim,
+		KeystoneURL:       s.KeystoneURL,
+	})
+	if err != nil {
+		glog.Fatalf("Invalid Authentication Config: %v", err)
+	}
+
+	authorizationModeNames := strings.Split(s.AuthorizationMode, ",")
+	authorizationConfig := genericauthorizer.AuthorizationConfig{
```
DirectXMan12: Should this have separate authorization, or just bounce off of the main API server use SARs, etc?

cc @deads2k

deads2k:
> Should this have separate authorization, or just bounce off of the main API server use SARs, etc?

Separate authentication powered by the webhook authorizer.

DirectXMan12:
> Separate authentication powered by the webhook authorizer.

I think we're agreeing here -- my point was that you don't necessarily need/want the other auth mechanisms -- you just want the webhook one.

deads2k:
> I think we're agreeing here -- my point was that you don't necessarily need/want the other auth mechanisms -- you just want the webhook one.

Yeah, your launch command probably doesn't need other options.

mksalawa:
@DirectXMan12 about the other two PRs - they were the trials of different approaches to splitting this task into smaller parts. We ended up with this as the best solution, so those will be either closed or updated after this one is finished.

metrics/options/options.go
```diff
@@ -36,14 +38,20 @@ type HeapsterRunOptions struct {
...
 
 func (h *HeapsterRunOptions) AddFlags(fs *pflag.FlagSet) {
+	h.ServerRunOptions.AddUniversalFlags(pflag.CommandLine)
+
 	fs.Var(&h.Sources, "source", "source(s) to watch")
 	fs.Var(&h.Sinks, "sink", "external sink(s) that receive data")
 	fs.DurationVar(&h.MetricResolution, "metric_resolution", 60*time.Second, "The resolution at which heapster will retain metrics.")
-	fs.IntVar(&h.Port, "port", 8082, "port to listen to")
+	// TODO: Revise this flag before Heapster v1.3 and Kubernetes v1.5
+	fs.IntVar(&h.Port, "heapster-port", 8082, "port used by the Heapster-specific APIs")
```
nikhiljindal: Why do we want to expose the heapster specific APIs on a different port?

Changing the meaning of an existing flag doesnt seem right to me.
If I was setting --port=8090 in my config and was expecting to see a heapster specific API at that port, then this change will break me.

nikhiljindal:
Didnt look too closely at the first Godeps commit.
The second commit mostly LGTM, apart from a comment regarding the port flag.

nikhiljindal:
Also, what about tests?
Not sure what testing do you have around the heapster API server.

metrics/options/options.go
piosz: We are not sure how to handle auth, so intially we would like to have two ports: for the native heapster api and metrics api server via api server. Eventually native api should be server on the same port as an addtional handlers in genericapiserver library.

I agree, but this is because there is already deprecated flag in genericapiserver with the same name and we have a conflict. Eventually (before releasing next Heapster version) we want to either remove the flag from genericapiserver or move it to apiserver and then revert this change here.

piosz:
@DirectXMan12 @deads2k we are not sure how the current auth model in Heapster works, so to not break anything we are starting a new http server (using genericapiserver) library on a different port which uses exactly the same auth model that federated-apiserver uses. We will reach you to figure out the auth issues.

piosz:
@nikhiljindal thanks for the review.

@mksalawa could you please flag-gate this change?

mksalawa:
@nikhiljindal as for the tests - there are no resources registered yet, so once we have them, we'll reuse the existing tests for the Metrics API on a different endpoint. Would that be ok? As for now, we would actually check only if the endpoint is reachable.

piosz:
LGTM

piosz:
Tests on Travis passed, but Travis hangs forever.

DirectXMan12:
Please file a bug or follow-up PR to discuss and/or fix the auth as per my comments and @deads2k's comments.

PR #1304, opened 2016-09-21
Register metrics API group with NodeMetrics

mksalawa:
Metrics storage logic copied from existing handlers.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1304)

### DIFF
--- a/metrics/apis/metrics/install/install.go
+++ b/metrics/apis/metrics/install/install.go
@@ -88,7 +88,9 @@ func enableVersions(externalVersions []unversioned.GroupVersion) error {
 func newRESTMapper(externalVersions []unversioned.GroupVersion) meta.RESTMapper {
 	// the list of kinds that are scoped at the root of the api hierarchy
 	// if a kind is not enumerated here, it is assumed to have a namespace scope
-	rootScoped := sets.NewString()
+	rootScoped := sets.NewString(
+		"NodeMetrics",
+	)
 
 	ignoredKinds := sets.NewString()
 
--- a/metrics/apis/metrics/install/install.go
+++ b/metrics/apis/metrics/install/install.go
@@ -1,0 +1,127 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


piosz:
@wojtek-t could you please take a look?

piosz:
LGTM from me, @nikhiljindal could you please take a look?

metrics/apis/metrics/install/install.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+const importPrefix = "k8s.io/heapster/metrics/apis/metrics"
+
+var accessor = meta.NewAccessor()
+
+// availableVersions lists all known external versions for this group from most preferred to least preferred
+var availableVersions = []unversioned.GroupVersion{v1alpha1.SchemeGroupVersion}
+
+func init() {
```
nikhiljindal: FYI, @deads2k and @lavalamp have changed this install code at many places.
You can look at https://github.com/kubernetes/kubernetes/blob/cf7301f16c036363c4fdcb5d4d0c867720214598/pkg/apis/extensions/install/install.go for an example.

I think @lavalamp wanted to fix it some more before updating all install.go, so this should be fine. But wanted to let you know that we will need to update it soon.

nikhiljindal:
Just one suggestion.
Also, again no tests? Will be good to add some tests :)

metrics/apis/metrics/install/install.go
```diff
@@ -0,0 +1,129 @@
...
```
deads2k: Take a look at the other install.go files under `/apis` in authentication, authorization, batch.  They're much smaller.

mksalawa:
@nikhiljindal thanks for the review! The new version looks much better. I would rather update it in the follow-up PR, as it would require godeps update I guess.

As for the tests, we talked about it with @piosz and from what I understand, as the connection to the API server is secured, we need the generated client first to get there in a way kube API is accessed, am I right?

metrics/apis/metrics/install/install.go
piosz: AFAIU your objection is similar to https://github.com/kubernetes/heapster/pull/1304#discussion_r80764878. @mksalawa will fix this in a follow up PR.

piosz:
LGTM,
Thanks @mksalawa for the change!
Thanks @nikhiljindal @deads2k for the review.

piosz:
Travis time-outed.

nikhiljindal:
@mksalawa Yes you will need that for the complete e2e test. But in the meantime you can add unit tests for the code you are adding.
You can also take a look at https://github.com/kubernetes/kubernetes/blob/master/test/integration/examples/apiserver_test.go, which tests an apiserver without enabling auth. It uses simple CURL commands to verify that apiserver comes up fine and expected endpoints exist on the server.

PR #1318, opened 2016-09-28
Remove service-cluster-ip-range flag from configs not using apiserver

mksalawa:
This flag was needed before the "api-server" flag was introduced. Now it's only obligatory if the API server is enabled.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1318)


piosz:
LGTM

PR #1321, opened 2016-09-29
Unit test for Heapster API server

mksalawa:
Basic test, checking availability of the endpoints.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1321)

### DIFF
--- a/metrics/heapster_test.go
+++ b/metrics/heapster_test.go
@@ -1,0 +1,211 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


piosz:
@nikhiljindal could you please take a look?

metrics/heapster_test.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func TestRunSecureServer(t *testing.T) {
+	serverIP := fmt.Sprintf("https://localhost:%d", testSecurePort)
+
+	go func() {
+		opt := getServerOptions()
+		opt.InsecurePort = 0
+		opt.SecurePort = testSecurePort
+		opt.CertDirectory = "./tmp"
```
mksalawa: The default value here is "/var/run/kubernetes", which fails with permission denied, so I'm not sure what should be used here. Is there any particular place the certificates should go to?

nikhiljindal: Is it a required param? What if you dont set it at all

nikhiljindal:
Thanks, LGTM assuming the tests are passing

metrics/heapster_test.go
mksalawa: Sorry, it could have been unclear - if it's not set, it defaults to what I mentioned above, "var/run/kubernetes", which causes the problem.

piosz:
LGTM

PR #1324, opened 2016-09-30
Add support for scoped tokens in Monasca Sink

taimir:
In practice most Monasca monitoring setups require an authorization role for the monasca metric collecting agents (hear the heapster monasca sink). Since roles are only associated with scope authorization, this needs to be supported.

The required change is in the `keystone_client.go`, where a scope is added to the token creation request if `TenantID` has been provided by the user. Note that the currently used version of `gophercloud` (OpenStack client) requires that there is no `TenantID` field in the AuthOpts passed to a token creation request, that's why it needs to be set to empty.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1324)

### DIFF
--- a/docs/sink-configuration.md
+++ b/docs/sink-configuration.md
@@ -105,7 +105,7 @@ To use the Monasca sink add the following flag:
 	--sink=monasca:[?<OPTIONS>]
 
 The available options are listed below, and some of them are mandatory. You need to provide access to the Identity service of OpenStack (keystone).
-Currently, only authorization through `username` / `userID` + `password` / `APIKey` is supported.
+Currently, only authorization through `username` / `userID` + `password` / `APIKey` is supported. If the agent access (for sending metrics) to Monasca is restricted to a role, please specify the corresponding `tenant-id` for automatic scoped authorization.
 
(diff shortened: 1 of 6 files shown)


piosz:
@k8s-bot ok to test

piosz:
I'm fine with the test, but we need someone who understand the Monasca to take a look. Do you have any suggestion?

piosz:
cc @cheld

taimir:
@witekest maybe you can have a look

taimir:
Just for reference: the change is necessary because of the way `gophercloud/rackspace` processes the AuthOpts when connecting to keystone. See [here](https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L48) (`TenantID` not allowed in opt) and [here](https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L163) for how one can achieve scoped auth.

I believe this library was revisited a couple of weeks ago, so it might be worth to consider bumping to the newest version: https://github.com/gophercloud/gophercloud

cheld:
manually tested and code reviewed. lgtm

piosz:
Thanks @taimir for the contribution. Thanks @cheld for the review.

Issue #1316, opened 2016-09-27
Unable to view Grafana 


owenmorgan:
From uncommenting the "type: LoadBalancer" from "grapfana-service.yaml" and running

$ kubectl create -f deploy/kube-config/influxdb/

Everything loads ok, and I'm able to see metrics on the Dashboard. But hitting the ELB thats created for Grafana, I see the below.

![image](https://cloud.githubusercontent.com/assets/10656664/18864970/c8142a88-8490-11e6-96bb-9b8464483b21.png)

access log on the pod shows.

> 2016-09-27T07:58:40.336230046Z 2016/09/27 07:58:40 [I] Completed 10.2.96.0 - "GET /api/v1/proxy/namespaces/kube-system/services/monitoring-grafana/public/app/app.ca0ab6f9.js HTTP/1.1" 404 Not Found 23 bytes in 403us

aegixx:
Same issue here.

For the record, these configuration files are fairly out-of-date - though I'm not sure if that's the issue.

I'm hooking up Chronograf instead of Grafana, but I'll let you know if I have any success.

LaurentDumont:
I was able to use the grafana service with a NodePort endpoint instead.

```
apiVersion: v1
kind: Service
metadata:
  labels:
    kubernetes.io/name: monitoring-grafana
  name: monitoring-grafana
  namespace: kube-system
spec:
  type: NodePort
  ports:
  - port: 3000
    name: http
    targetPort: 3000
    nodePort: 30005
  selector:
    name: influxGrafana
```

DirectXMan12:
> I'm not quite sure what is the status of the Heapster project as a part of the Kubernetes ensemble of products though

Currently, it's required for the Horizontal Pod Autoscaler to function, and otherwise is useful for collecting metrics and placing those into sinks (InfluxDB being one of them).  We're currently re-evaluating the monitoring pipeline, however.

As for the state of the grafana configs, they are most likely out of date.  Grafana is not really a part of Heapster (those configs are mostly just provided as an example of how to set up the InfluxDB sink).  If you have updates that work, we'd be willing to accept them in pull-request form.

aegixx:
I do have an update for grafana + a working configuration for chronograf - I'll post a PR shortly

aegixx:
I've updated the configurations in #1326, but this isn't going to fix the Grafana display problem - however, Chronograf works.  Though I haven't reviewed the actual metrics being collected, so #1320 may still be relevant.

[...]

PR #1326, opened 2016-10-01
Update influxdb configuration (update grafana) 

aegixx:
- Update influxdb & grafana images
- Update grafana configuration files

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1326)


deploy/kube-config/influxdb/chronograf-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: chronograf
+  namespace: kube-system
+spec:
+  # In a production setup, we recommend accessing Grafana through an external Loadbalancer
+  # or through a public IP.
+  type: LoadBalancer
+  # type: NodePort
+  ports:
+  - port: 80
+    targetPort: 10000
```
LaurentDumont: We could add a nodePort: XXXXX as a comment just to make the entire process a bit clearer for the average user.

deploy/kube-config/influxdb/grafana-service.yaml
```diff
@@ -2,16 +2,17 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
-    kubernetes.io/cluster-service: 'true'
```
LaurentDumont: Why remove the cluster label from the Grafana service? I know that it might cause issue for a service that isn't part of a Kubelet manifest but it could be indicated to the user.

aegixx: This is just me being a newbie to Kubernetes :-) - I didn't realize these had any relevance beyond the deployment -- I will add both of these labels back.

DirectXMan12:
cc @mwielgus

deploy/kube-config/influxdb/chronograf-service.yaml
aegixx: nodePort is not required, so wouldn't it be more confusing to explicitly specify one?

LaurentDumont: Yeah, that makes more sense. I was used to specifying the nodePort so I dont have to look up which dynamic port was assigned.

deploy/kube-config/influxdb/grafana-service.yaml
kris-nova: @LaurentDumont - wondering if we can get more information about the issue this *might* cause? We are noticing some unexpected behavior when having this label defined.

LaurentDumont: @kris-nova As far as I know, if you manually start a service as "cluster-wide" and start it AND it is missing from /etc/kubernetes/addons, the addons.sh service is going to kill the service. I was using kube-deploy to start the cluster which means that I was using a fixed list of cluster-wide services.

Adding it to /etc/kubernetes/addons should create everything you need automagically.

docs/influxdb.md
```diff
@@ -1,30 +1,39 @@
...
-Grafana is set up to auto-populate nodes and pods using templates.
 
 ## Troubleshooting guide
 1. If the Grafana service is not accessible, chances are it might not be running. Use `kubectl.sh` to verify that the `heapster` and `influxdb & grafana` pods are alive.
 
-	kubectl get pods
+	kubectl --namespace kube-system get pods
 
-	kubectl get services
+	kubectl --namespace get services
```
mssola: It should be `kubectl --namespace kube-system get services`, no ?

aegixx:
CLA signed

docs/influxdb.md
aegixx: Thank you - fixed!

deploy/kube-config/influxdb/grafana-service.yaml
aegixx: Do you think it is confusing to assume everyone is deploying this as a cluster add-on?  If they don't take that additional step - their service will be shutdown.  On the other hand, if they **should** be deployed this way, should _grafana_, _chronograf_, and _influxdb_ also be flagged with `cluster-service` as well?

LaurentDumont: @aegixx Well, I guess that the "best-practice" approach would be to deploy it as a cluster-service. It handles update/restart/shutdown as opposed to a manual deployment. As long as it's clear in the docs (we can work on that part!) I feel like a cluster service is a best setup.

In the old configs, influxdb and grafana are marked as cluster services. Since Chronograf is essentially a Grafana replacement, it should also be marked as one.

aegixx: Made all services `cluster-service` and added notes on running them independently.

aegixx:
I believe everything has been handled and this should be ready for merging.  Looks like the CI test is failing due to `Your test run exceeded 120 minutes` - which I don't think I can do much about :-)

piosz:
@k8s-bot ok to test

piosz:
@aegixx Please rebase

piosz:
IIUC this PR adds Chronograf UI. I have no idea what is it and why do we need it, especially in the default setup. Who will maintain this? Could you please elaborate a bit?

aegixx:
@piosz Neither Grafana or Chronograf are part of the project officially - so I think these integrations are primarily introduction points depending on which stack is consuming your metrics.  I'm fine removing Chronograf in favor of Grafana (based on my own usage, Chronograf is simpler, but inferior to what Grafana provides)  We just need to ensure Grafana functions as intended based on this PR (which I'll validate prior to my rebase)

aegixx:
Removed Chronograf and confirmed Grafana functions as intended.  This PR is now focused on getting Grafana up and running only.

piosz:
@aegixx as I wrote in https://github.com/kubernetes/heapster/issues/1316#issuecomment-252897423 I'm against having Grafana with public IP in the default setup. How about just adding a comment that if you are changing service type to `LoadBalancer` you should also update `GF_SERVER_ROOT_URL` arg in Grafana pod spec?

There is already Grafana RC specification. I'm ok with migrating it to Deployment but please remove RC in this case.

Also please follow the same patterns as in https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/cluster-monitoring/influxdb (label names, etc.).

aegixx:
With the removal of Chronograf, I can vastly simplify this PR, going to close and reopen a new one with a cleaner history.

PR #1335, opened 2016-10-12
[WIP] vendor: Replace Godep with glide

### LINKED ISSUE #1314: Move to glide
Kubernetes will eventually migrate to use Glide instead of godep (which is kinda deprecated for vendoring): https://github.com/kubernetes/kubernetes/pull/24202

Can we start and move this project to glide?

@thockin @piosz @DirectXMan12

nhlfr:
Fixes #1314

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1335)


piosz:
Did you bump any deps? I'd expect this change to be transparent in term of added/removed files.

DirectXMan12:
Perhaps glide isn't doing the filtering based on used vs unused files, like Godep does?  Do we need glide-vc?

AlmogBaku:
@DirectXMan12 @piosz @nhlfr as I wrote in #1314 there is no need to commit `vendor` directory at all.. that's one of the great benefits of glide.. specify semver versions for deps, and just do `glide install` and voila!

Makefile
```diff
@@ -8,25 +8,26 @@ SUPPORTED_KUBE_VERSIONS = "1.3.6"
 TEST_NAMESPACE = heapster-e2e-tests
 
 deps:
-	which godep || go get github.com/tools/godep
+	which glide || go get github.com/Masterminds/glide
+	glide update
```
AlmogBaku: should be `glide install`

nhlfr:
That's right. I'm going to remove the `vendor` dir today.

AlmogBaku (changes requested):
IMO, from my experience there is no need to commit `vendor` directory at all.

AlmogBaku:
@nhlfr can you fix the PR? it looks like the tests failed

Makefile
```diff
@@ -7,26 +7,39 @@ FLAGS =
...
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o eventer k8s.io/heapster/events
 
 sanitize:
 	hooks/check_boilerplate.sh
 	hooks/check_gofmt.sh
 	hooks/run_vet.sh
 
 test-unit: clean deps sanitize build
-	GOOS=linux GOARCH=amd64 godep go test --test.short -race ./... $(FLAGS)
+	GOOS=linux GOARCH=amd64 go test --test.short -race ./... $(FLAGS)
```
AlmogBaku: should be
```
GOOS=linux GOARCH=amd64 go test --test.short -v -race `glide novendor` $(FLAGS)
```

This will prevent testing the vendored libs.

Makefile
```diff
@@ -7,26 +7,39 @@ FLAGS =
 SUPPORTED_KUBE_VERSIONS = "1.3.6"
 TEST_NAMESPACE = heapster-e2e-tests
 
-deps:
-	which godep || go get github.com/tools/godep
+verify-glide-installation:
+	which glide || go get github.com/Masterminds/glide
+	which glide-vc || go get github.com/sgotti/glide-vc
```
AlmogBaku: why is that needed if we are not going to commit the `vendor` dir?

AlmogBaku (changes requested):

nhlfr:
Looks like glide is using raw hg to fetch a repo from bitbucket...

AlmogBaku:
if im not mistaken so as godep... so why usage of hg natively here doesnt work?

nhlfr:
Probably `go get` and godep don't use `hg` as an executed command - that's my assumption, I cannot imagine any other scenario.

However, I'll try to think about and propose a change to k8s infra.

AlmogBaku:
looks very strange... since im pretty sure that godep required me to install hg..
can we re-run it again? or at least just fix the integration test?

AlmogBaku:
ping @nhlfr

piosz:
@luxas

luxas:
I would **not** switch to Glide in any k8s official projects before we have done it for core (if we really want that, I'm not sure). Actually I'm in favor for committing the vendor dir into the project, because once you've downloaded the project with git, you don't have to fetch it with glide in scripts or manually.

cc @thockin for thoughts on glide

This will not at least make the next release, and I don't think there is any argument for doing so either.

thockin:
Glide still has shortcomings and is sort of in stasis until the Go
vendoring workgroup produces results.  I would not swicth yet.  Godep is
inscrutable, but we largely understand a way to use it that works.

piosz:
As I wrote in https://github.com/kubernetes/heapster/issues/1314#issuecomment-253418553, I think Heapster should follow Kubernetes standards. According to what @thockin we want to stay with godeps for a while. Once Kubernetes is migrated we will revisit this decision. Closing the PR in the meantime.

PR #1340, opened 2016-10-15
Add troubleshooting information regarding using heapster and autoscaler

agsola:
Many users install heapster to get autoscaling working.
Maybe it's worth adding an additional point for those users to clarify the need of setting CPU request in order to get it working, since the user is able to see the metrics but unable to get it working in the autoscaler.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1340)


agsola:
I signed it!

piosz:
@k8s-bot ok to test

piosz:
This is unrelated to influxdb. How about putting this into `heapster/docs/debugging.md`?

agsola:
I suggested that file as it includes the common troubleshooting guide, and would be the most common place to look for.
Maybe it can be added to debugging and link it from docs/influxdb.md

piosz:
@agsola I'm fine with linking to `docs/debbuging.md` from `docs/influxdb.md`

AlmogBaku:
This is not relevant for InfluxDB.. and afaik HPA is not depends on any sink...

piosz:
Agree with @AlmogBaku

PR #1357, opened 2016-10-27
Modifying Makefile to add support for s390x System

Ryan-Dmello:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1357)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,13 +6,14 @@ FLAGS =
 
 SUPPORTED_KUBE_VERSIONS = "1.3.6"
 TEST_NAMESPACE = heapster-e2e-tests
+ARCH=$(shell uname -i)
 
 deps:
 	which godep || go get github.com/tools/godep
 
 build: clean deps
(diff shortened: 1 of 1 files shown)


DirectXMan12:
The concept looks good.  Does the race detector not work on s390x?

Ryan-Dmello:
Currently we faced some issue while building heapster for s390x. So after some investigation came across this bug which says -race detector flag is only supported for amd64

here is the link related to that bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831724

Ryan-Dmello:
Hi,

wanted to know whether there is a concept of corporate CLA for cla/linuxfoundation. Any pointers to signing this cla will be very helpful.

Thanks,
Ryan D'Mello

piosz:
@Ryan-Dmello see https://github.com/kubernetes/kubernetes/blob/master/CONTRIBUTING.md#contributor-license-agreements

piosz:
closing in favor of #1387

PR #1364, opened 2016-11-05
add version info for heapster/eventer

### LINKED ISSUE #714: Embed version via ldflags
Right now the version is hard-coded & not updated properly - see https://github.com/kubernetes/heapster/blob/master/version/version.go#L22. This should be changed to add the version at compile time via `ldflags`.

andyxning:
Currently, Heapster supports print version info with `--version` command line option. It utilizes [the version package from Kubernetes](https://github.com/kubernetes/kubernetes/tree/master/pkg/version) and the output is useless when i just run `heapster --version`. 

IMO, Kubernetes is used to distributed with binary, we need the info about `GitVersion`, `GitCommit`, `BuildDate`, `GoVersion` and so on to determine which go version it is built and when it is built cause it is build by others. Heapster is mainly distributed with source code. So, i think what we mainly need is the version and git commit hash value.

Version info is not hard coded in the version package in Heapster. To be more like a gopher, we can use the `-ldflags` to set the version info when compile. :)

Also, currently, `Eventer` does not support print version info with `--version`.

This PR will remove the use of version package from Kubernetes, and write the simple  process about printing the version and git commit hash info with `-ldflags`.

fix #714

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1364)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 all: build
 
-TAG = v1.2.0
 PREFIX = gcr.io/google_containers
 FLAGS = 
+
+VERSION = v1.2.0
+GIT_COMMIT := `git rev-parse --short HEAD`
 
 SUPPORTED_KUBE_VERSIONS = "1.4.6"
 TEST_NAMESPACE = heapster-e2e-tests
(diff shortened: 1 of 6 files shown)


andyxning:
i signed it

andyxning:
@piosz @AlmogBaku

AlmogBaku (approved):
LGTM

Makefile
```diff
@@ -4,15 +4,18 @@ TAG = v1.2.0
 PREFIX = gcr.io/google_containers
 FLAGS = 
 
+VERSION = 1.2.0
```
piosz: We have now TAG and VERSION which are supposed to be the same. Could you please use just one of them?

andyxning: done

piosz:
@k8s-bot ok to test

Makefile
```diff
@@ -1,18 +1,20 @@
 all: build
 
-TAG = v1.2.0
 PREFIX = gcr.io/google_containers
 FLAGS = 
 
+VERSION = 1.2.0
```
piosz: nit: please change to v1.2.0 to follow the current pattern

piosz:
LGTM, just a nit

Makefile
andyxning: Done

piosz: I can't see it. Did you commit your change?

piosz:
@andyxning something went wrong with rebase. Now you have 100+ files modified in this PR

andyxning:
@piosz Fixed. PTAL.

piosz:
LGTM, thanks!

PR #1378, opened 2016-11-11
added url param to force ipv4 or ipv6 mode

kcrawley:
This PR allowed the user to pass a URL parameter with the `source` argument when executing `heapster`, effectively forcing the IP scraper to get the first IP it encounters of the specified type.

Name: `ipMode`
Valid options: `4, v4, ipv4` || `6, 16, v6, ipv6`

```
...
heapster --source=kubernetes:https://foo.default?ipMode=v4 --sink=...
```

Without this parameter, the application should behave as it did before, selecting the first IP it encounters.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1378)

### DIFF
--- a/metrics/sources/kubelet/configs.go
+++ b/metrics/sources/kubelet/configs.go
@@ -57,11 +57,19 @@ func GetKubeConfigs(uri *url.URL) (*kube_client.Config, *kubelet_client.KubeletC
 			return nil, nil, err
 		}
 	}
+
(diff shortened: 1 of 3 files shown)


kcrawley:
Did not realize I had made a change to a vendor library to support this. I'll be revisiting this and finding a way around it.

PR #1313, opened 2016-09-23
Feature/elasticsearch new mapping

AlmogBaku:
fixes #1310, #1278 #1263  #1262:

This PR brings a lot of changes, most of them caused by refactoring the schema .
In order to allow effective usage of the ELK stack, we had to modify the schema so it'll work properly with Kibana(otherwise, it require development of a custom special dashboard! which is insane!)

This PR fixes the following issues:
- Types per "metric family"(cpu/memory/network/fs/etc): each entry includes more than one parameter that related to its family; allowing us to correlate the data(such as "usage VS limit")
  - Uses bulk functionalities of ES
  - Creates daily indexes (allowing easy removal of old data, as Logstash and others do)
  - Create aliases for the indexes

This is a big PR, which bring breaking changes(old schema not valid anymore). However- I have to admit that the old schema is totally unusable, and probably not many users use it if any(but me)

cc @huangyuqi @DirectXMan12 @piosz @thermi

You can see the exported schema here: https://gist.github.com/AlmogBaku/8fe24fd9c61df5a72402a4d25992cb36

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1313)

PR #1379, opened 2016-11-14
[ES] enhance es mapping for event

andyxning:
This PR adds support for adding more info for analyze events in ElasticSearch. This PR adds `EventType`, `EventKind`, `EventReason` and `EventMessage` as tags and thus allows us to query based on them. Such as, what are the warning events in last two hours. When does the cluster emits `FailedScheduling` events.

This PR also adds the **namespace** tag info when the event is of Pod kind.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/heapster/1379)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-sudo: false
-
 language: go
 
 go_import_path: k8s.io/heapster
--- a/common/elasticsearch/mapping.go
+++ b/common/elasticsearch/mapping.go
@@ -141,6 +141,22 @@ const mapping = `{
       "properties": {
         "EventTags": {
           "properties": {
+            "eventType": {
+              "type": "string",
+              "index": "not_analyzed"
+            },
+            "eventKind": {
+              "type": "string",
+              "index": "not_analyzed"
+            },
+            "eventReason": {
+              "type": "string",
+              "index": "not_analyzed"
+            },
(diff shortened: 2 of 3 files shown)


andyxning:
@googlebot i run it correctly in my laptop.

andyxning:
I don't know why the test `TestOneExportInTime` in `k8s.io/heapster/metrics/sinks` will fail. I runs ok in my laptop. :(

andyxning:
it turns out that memory is out and something like `go build testmain: /home/travis/.gimme/versions/go1.6.linux.amd64/pkg/tool/linux_amd64/link: signal: killed` is reported. For more info please see [Travis CI build envrionment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) and [Travis CI FAQ about memory limit](https://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error)

AlmogBaku:
That's an interesting approach, i solved that slight differently in #1313

piosz:
Can anyone review it? @AlmogBaku?

AlmogBaku:
I think it's now redundant after #1313
If you see any reason to implement it this way other than the method i
used, can you elaborate?

piosz:
closing per @AlmogBaku comment

PR #1402, opened 2016-12-06
README: discussion to slack and kubernetes-dev

philips:
I think that IRC channel is deprecated, move people to slack.


DirectXMan12:
LGTM :+1:, but we might want to mention both sig-instrumentation and sig-autoscaling (since many questions are HPA-related)

AlmogBaku:
LGTM

PR #1418, opened 2016-12-15
Make retention policy configurable when using influxdb

### LINKED ISSUE #566: Set default retention policy in InfluxDB
We should set default retention policy in InfluxDB to be 5 days. Any data that needs to persist longer should be moved to a different database using continuous queries and separate retention policies.

Any thoughts @mwielgus @piosz ?

mpvaniersel:
This makes Heapster configure the retention policy used by InfluxDB for the data Heapster stores. This will make it possible to limit Heapster data collection to a certain period of time.  In the previous situation, Heapster data is stored forever until the disk is full, and there is no way to reliably override this retention policy in a kubernetes environment. 

This fixes #566

### DIFF
--- a/metrics/sinks/influxdb/influxdb.go
+++ b/metrics/sinks/influxdb/influxdb.go
@@ -197,7 +197,6 @@ func (sink *influxdbSink) createDatabase() error {
 	if sink.dbExists {
 		return nil
 	}
-
 	q := influxdb.Query{
 		Command: fmt.Sprintf(`CREATE DATABASE %s WITH NAME "default"`, sink.c.DbName),
 	}
@@ -218,7 +217,7 @@ func (sink *influxdbSink) createDatabase() error {
 
 func (sink *influxdbSink) createRetentionPolicy() error {
 	q := influxdb.Query{
-		Command: fmt.Sprintf(`CREATE RETENTION POLICY "default" ON %s DURATION 0d REPLICATION 1 DEFAULT`, sink.c.DbName),
+		Command: fmt.Sprintf(`CREATE RETENTION POLICY "default" ON %s DURATION %s REPLICATION 1 DEFAULT`, sink.c.DbName, sink.c.RetentionPolicy),
 	}
 
 	if resp, err := sink.client.Query(q); err != nil {
--- a/common/influxdb/influxdb.go
+++ b/common/influxdb/influxdb.go
@@ -32,13 +32,14 @@ type InfluxdbClient interface {
 }
 
 type InfluxdbConfig struct {
-	User        string
-	Password    string
-	Secure      bool
-	Host        string
-	DbName      string
-	WithFields  bool
-	InsecureSsl bool
+	User            string
+	Password        string
+	Secure          bool
+	Host            string
+	DbName          string
+	WithFields      bool
+	InsecureSsl     bool
+	RetentionPolicy string
 }
 
 func NewClient(c InfluxdbConfig) (InfluxdbClient, error) {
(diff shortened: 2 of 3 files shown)


mpvaniersel:
I signed the CLA

mpvaniersel:
Not sure why the linuxfoundation CLA needs signing, but I went ahead and signed that one  as well.

DirectXMan12:
I've edited the title a bit, but please fill in a few more details in the PR description.

metrics/sinks/influxdb/influxdb.go
```diff
@@ -197,14 +197,20 @@ func (sink *influxdbSink) createDatabase() error {
...
+	qCreate := influxdb.Query{
 		Command: fmt.Sprintf("CREATE DATABASE %s", sink.c.DbName),
 	}
-	if resp, err := sink.client.Query(q); err != nil {
+	if resp, err := sink.client.Query(qCreate); err != nil {
 		if !(resp != nil && resp.Err != nil && strings.Contains(resp.Err.Error(), "already exists")) {
 			return fmt.Errorf("Database creation failed: %v", err)
 		}
 	}
+	qRetPol := influxdb.Query{
```
DirectXMan12: if no retention policy is set in the sink config, should we still run this query?

DirectXMan12 (commented):
I think this looks good.  I don't think we have to worry too much about escaping issues, since this is a config parameter, and not some input coming from the user.

metrics/sinks/influxdb/influxdb.go
mpvaniersel: Good point. Since there is a default defined of "0", I thought it would not be necessary to check if there is a retention policy set. But it would be easy to wrap it in a "if (sink.c.RetentionPolicy) {}"

mpvaniersel:
@DirectXMan12 Now that I have your attention - right now this code only applies to influxdb as a metrics sink, not when it's used as an events sink. Do you think I should apply these changes to both? I hesitate to do so as I can't test that in my current setup.

piosz:
@mpvaniersel could you please rebase?
@luxas could you please take a look?

metrics/sinks/influxdb/influxdb.go
luxas: Yes, please have this inside an if statement that checks `if RetentionPolicy != "0"`

luxas (commented):
I think @DirectXMan12 is better suited for this, but at a quick look it looks good to do this.

Rebase and comment fix needed though

mpvaniersel:
I rebased. The code is much simpler now as creation of retention policy has also been added by somebody else in the mean time. Now it's just a matter of passing the configuration parameter to createRetentionPolicy.

eric-tucker:
Can we pretty please get any merge conflicts fixed up and merge this?  We really need an easy way to configure the retention policy to something not infinite.

mpvaniersel:
I rebased this a couple of weeks ago, nobody looked at it, and now it's conflicting again.
I would rebase this again IF I can get a promise from a developer that this will be merged in a short time frame. Otherwise it's just a waste of my time.

DirectXMan12 (approved):
This LGTM :+1:.  The only thing I'd worry about is a malformed retention policy argument botching the query, but I don't think that's too big a of problem.

DirectXMan12:
gah, apparently GitHub's web resolve-conflicts tool doesn't quite do what I want it to do (sorry about that extraneous commit).  I'll merge the changes as soon as you rebase.

I'd like to apologize for the delay -- I know long review delays can be frustrating, and I'm sorry I caused one.

mpvaniersel:
I fixed the whitespace issue that was tripping up the CI. Please go ahead and merge!

luxas:
Travis is green :+1:

DirectXMan12:
Thanks for the contribution.  Sorry it took so long :-/

mpvaniersel:
Thanks! Don't worry about it.

PR #1421, opened 2016-12-19
	Convert cumulation to rate.

gadaigadai:
Maybe the word 'cumulaties' is misspelled.


DirectXMan12:
I think the word we're looking for here is "cumulative", as in "convert a cumulative metric to a rate metric"

PR #1427, opened 2016-12-21
Use protobuf instead of jsons

### LINKED ISSUE #1422: Heapster should be using protobufs for communication with main apiserver
Heapster is reading/watching a lot of data from main apiserver (especially in large clusters). To reduce the load both in heapster and in apiserver, we should use protobufs for this communication.

 If I correctly understand the code here in heapster, we are currently using these two binaries:
https://github.com/kubernetes/heapster/blob/e19c9fb0d78695ce02e34afc821be5525f70f1d7/metrics/heapster.go#L55
https://github.com/kubernetes/heapster/blob/e19c9fb0d78695ce02e34afc821be5525f70f1d7/events/eventer.go#L44

It seems to me that to achieve this, we only need to modify this function:
https://github.com/kubernetes/heapster/blob/96d9dbf9ee20c200bb7703a8dff0b5ea7c1d1ba1/common/kubernetes/configs.go#L66

by setting:
```
kubeConfig.ContentType = "application/vnd.kubernetes.protobuf"
```

@gmarek @shyamjvs @piosz @aleksandra-malinowska

aleksandra-malinowska:
Fix #1422

### DIFF
--- a/common/kubernetes/configs.go
+++ b/common/kubernetes/configs.go
@@ -136,5 +136,7 @@ func GetKubeClientConfig(uri *url.URL) (*kube_client.Config, error) {
 		}
 	}
 
+	kubeConfig.ContentType = "application/vnd.kubernetes.protobuf"
+
 	return kubeConfig, nil
 }


mwielgus:
LGTM

PR #1429, opened 2016-12-22
add labels to eventer es mapping

andyxning:
This PR adds `labels` field to eventer elasticsearch mappings. 

We often use `labels` to add some meta info such as the cluster name or service name corresponding to this pod. We need this info to do some filter search.

### DIFF
--- a/common/elasticsearch/mapping.go
+++ b/common/elasticsearch/mapping.go
@@ -311,6 +311,15 @@ var mapping = `{
         "Count": {
           "type": "long"
         },
+        "Labels": {
+          "type": "string",
+          "fields": {
+            "raw": {
+              "type": "string",
+              "index": "not_analyzed"
+            }
+          }
+        },
         "Metadata": {
           "properties": {
             "creationTimestamp": {
--- a/events/sinks/elasticsearch/driver.go
+++ b/events/sinks/elasticsearch/driver.go
@@ -23,6 +23,7 @@ import (
 	esCommon "k8s.io/heapster/common/elasticsearch"
 	event_core "k8s.io/heapster/events/core"
 	"k8s.io/heapster/metrics/core"
+	"k8s.io/heapster/metrics/util"
 	kube_api "k8s.io/kubernetes/pkg/api"
 )
 
@@ -50,6 +51,7 @@ type EsSinkPoint struct {
 	Message                  string
 	Reason                   string
 	Type                     string
+	Labels                   string
 	EventTags                map[string]string
 }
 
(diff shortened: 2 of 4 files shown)


AlmogBaku:
1) Where do you fill the data in the driver? you haven't committed any changes to the eventer driver...

2) Why do you need the RAW data of this field.. it's seems classy just for search...

andyxning:
@AlmogBaku 

>Where do you fill the data in the driver? you haven't committed any changes to the eventer driver...

Just like `name` and `namespace` field under `MetaData`, `labels` field is also contained in [`ObjectMeta`](https://github.com/kubernetes/kubernetes/blob/v1.4.0/pkg/api/types.go#L60-L157)

> Why do you need the RAW data of this field.. it's seems classy just for search...

Yep. I want to search some meta labels with `wildcard` or `regexp`. This is the main purpose to add labels field.

AlmogBaku:
@andyxning I can't see anywhere in the `driver.go` where it commit the labels into this field...
```go
func eventToPoint(event *kube_api.Event, clusterName string) (*EsSinkPoint, error) {
	point := EsSinkPoint{
		FirstOccurrenceTimestamp: event.FirstTimestamp.Time.UTC(),
		LastOccurrenceTimestamp:  event.LastTimestamp.Time.UTC(),
		Message:                  event.Message,
		Reason:                   event.Reason,
		Type:                     event.Type,
		Count:                    event.Count,
		Metadata:                 event.ObjectMeta,
		InvolvedObject:           event.InvolvedObject,
		Source:                   event.Source,
		EventTags: map[string]string{
			"eventID":      string(event.UID),
			"cluster_name": clusterName,
		},
	}
	if event.InvolvedObject.Kind == "Pod" {
		point.EventTags[core.LabelPodId.Key] = string(event.InvolvedObject.UID)
		point.EventTags[core.LabelPodName.Key] = event.InvolvedObject.Name
	}
	point.EventTags[core.LabelHostname.Key] = event.Source.Host
	return &point, nil
}
```

andyxning:
@AlmogBaku 
It's my fault for misconsidering [`Labels`](https://github.com/kubernetes/kubernetes/blob/v1.4.0/pkg/api/types.go#L124-L133) filed in `ObjectMeta` as `string` type. In that situation, eventer will not write event's ObjectMeta labels.

I have updated the PR by making `labels` field in the same level with `Reason` and `Message` cause we need to convert a labels map to a string.

andyxning:
@AlmogBaku I have test this feature in my local environment. It finally turns out that the `ObjectMeta` is contains data about the event itself not the corresponding pod or some other resource  types. The `Labels` is always empty. :(

piosz:
@AlmogBaku could you please accept the invitation to heapster-reviewers?

AlmogBaku:
I don't think this information is available without changing the functionality.. and I don't know for myself how you can dig this information.

I guess that for the time being we can close this issue.

andyxning:
@AlmogBaku 

> I don't think this information is available without changing the functionality.

Agree. To make this work. we need to make a change to kubernetes. I will make it later to see whether kubernetes community can agree on this.

At this moment, we can close this PR. :)

Issue #1410, opened 2016-12-12
Any plans to support Elasticsearch 5


maxott:
It appears that 1.2 can't connect to ES5

```E1212 05:24:11.987800     120 factory.go:75] Failed to create sink: Failed to create ElasticSearch client: no Elasticsearch node available```

It works perfectly fine for ES2. Unfortunately, some of my other components require ES5.

Any current plans to upgrade the ES backend?

AlmogBaku:
Upgrading to ES5 will break the current support of ES2...
I think that  for the time being we need to support both of the sinks version..

Also, It looks like we'll have to replace only the ES driver and not the eventer/heapster sinks, since it's modular.

Currently I don't have the time/resources to build the ansible script to install ES5, so I'll have to wait for ES team to update the script in order to create a new ES Cluster..

rikatz:
@AlmogBaku Why it will break support to ES2? I think it's only something like pointing the lib contained here: https://github.com/kubernetes/heapster/blob/master/common/elasticsearch/elasticsearch.go#L25

To the v5 library, right?

I'll try to change the source code pointing to this new version, and see what happens. We did this earlier with telegraf and worked fine, the only real problem was with fields containing '.' character, because ES 2 does not support them.

rikatz:
Guys, it's pretty simple to support Elastic v5. I did some simple tests here and it worked. I'm going to open a PR late tomorrow, to change the lib version (and some minor changes needed to make this work) and post here.

AlmogBaku:
https://github.com/olivere/elastic/issues/431

AlmogBaku:
@rikatz the discussion on the issue above means that we need to create a different sink for es5; however all the drivers can be the same

rikatz:
@maxott Elasticsearch 5 is already supported in Heapster (as merged in #1607).

Can you please close this issue?

Thanks!

PR #1443, opened 2017-01-04
Es5

rikatz:
This PR adds support to Elasticsearch 5, according to issue #1410

It has been tested on Elasticsearch 2 and on Elasticsearch 5 and works as expected.

There are three major changes:

Changes the referenced library from elastic.v3 to elastic.v5
Adds the context on validation functions (Do(context.Background()), as this is requested by the new library.
Changes the Godeps, removing old vendor and adding the new one.


rikatz:
@DirectXMan12 @AlmogBaku Will wait for CI to run again (had some problems with godep here...).

Hope this PR is better :)

DirectXMan12:
err... what's up with the "add missing files" commit?  Did you not use the `godep` tool to pull in the new dependency?  Also, you can squash the Godeps related commits into a single commit, just as long as the commit which updates the main code is separate from the one that updates the Godeps.json/vendor directory.

rikatz:
@DirectXMan12 you're right. I spent the whole morning trying to use godep, but because of some strange reason (or my fault) I wasn't able to add the deps correctly.

I've tried to godep save ./..., godep update ./..., and everything else and failed miserably! 

Sometimes it truncates the Godeps.json, or it erases the directory. 

With godep restore it tried to download missing libraries to my GOPATH (this is the right behaviour, I think) but broke with some dependencies that doesn't exists anymore (as the riemann library).

So I've made a new program, ran godep on it, copied the array of missing libraries to Godeps.json and the vendor/gopkg.in/olivere to the right place, inside the code.

As you have a lot more experience that I do, can you teach me the right way to do this? 

Thanks!!

rikatz:
@DirectXMan12 this is one of the libs with problem:

github.com/bigdatadev/goryman

rikatz:
@DirectXMan12 So this is exactly the problem, the referenced lib doesn't exists anymore on github.

I've found the same library on other repo (sauravmndl/goryman), the checksums from the files on 'vendor' directory and this repo matches.

Forked this one, changed the in program imports from the old repo to my new one, and commited.

The new library is rikatz/goryman.

Note that this lib seems deprecated (the original was deleted from github).

I'm going to close this PR by now, and open a new one to solve the riemann library problem (the reference, not the code!)

And when this is corrected, I'll open a new PR to this feature (ES5), and Godeps working fine!

AlmogBaku:
This is a proof why we need glide ASAP. Godep makes the development process a nightmare.
#1314

PR #1447, opened 2017-01-05
Document Heapster Maintenance Procedures

DirectXMan12:
This commit adds documentation for Heapster maintenance producers and
sink owners.  This should make it clearer what needs to be done to take
care of Heapster going forward.

Closes #1407

PR #1451, opened 2017-01-09
Docs:  Maintenance: PR number for Graphite Sink

DirectXMan12:
The PR number for the Graphite sink in the maintenance docs
was incorrect.  This fixes it.

See https://github.com/kubernetes/heapster/pull/1447#discussion_r95159533


piosz:
LGTM

PR #1452, opened 2017-01-10
Always --pull in docker build to ensure recent base images

ixdy:
@timstclair 

x-ref https://github.com/kubernetes/kubernetes/pull/39695

The riemann image is using debian:8.1 explicitly, which is pretty old (June 2015). Should we update it? I'm not sure who uses it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ container:
 	cp deploy/docker/Dockerfile $(TEMP_DIR)
 	cd $(TEMP_DIR) && sed -i "s|BASEIMAGE|$(BASEIMAGE)|g" Dockerfile
 
-	docker build -t $(PREFIX)/heapster-$(ARCH):$(VERSION) $(TEMP_DIR)
+	docker build --pull -t $(PREFIX)/heapster-$(ARCH):$(VERSION) $(TEMP_DIR)
 ifneq ($(OVERRIDE_IMAGE_NAME),)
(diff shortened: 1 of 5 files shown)


luxas:
LGTM, but I do not have write access.

I think we should a) update or b) remove the `riemann` and `kafka` images

Probably the latter; since the kafka image has not been updated in [more than a year](https://github.com/kubernetes/heapster/commits/master/kafka) and does likely not work.
cc @huangyuqi

Even worse, the `riemann` image has not been updated for more than one year and a half. cc @jfoy 

I propose to remove these images, these implementations seem to be a better fit outside of the main heapster repo, since the k8s team is not maintaining those images.

@mwielgus @piosz I can send a such PR if we decide that. I strongly think we should remove them for the v1.3 release. If someone comes and updates everything in time for an eventual v1.4 that might be fine if he/she commits to maintaining them as well.

DirectXMan12:
We're removing the riemann image, so I'm going to close this.

DirectXMan12:
(see #1455)

luxas:
@ixdy Please update this to include only the heapster, influxdb and grafana change.

@DirectXMan12 There were other changes as well, not only for the riemann image. Please reopen.

ixdy:
don't have write access here, so I can't reopen this PR. created #1456 instead.

PR #1455, opened 2017-01-11
Remove the outdated riemann and kafka images and examples

luxas:
> I think we should a) update or b) remove the riemann and kafka images

> Probably the latter; since the kafka image has not been updated in more than a year and does likely not work.
cc @huangyuqi

> Even worse, the riemann image has not been updated for more than one year and a half. cc @jfoy

> I propose to remove these images, these implementations seem to be a better fit outside of the main heapster repo, since the k8s team is not maintaining those images.

> @mwielgus @piosz I can send a such PR if we decide that. I strongly think we should remove them for the v1.3 release. If someone comes and updates everything in time for an eventual v1.4 that might be fine if he/she commits to maintaining them as well.

I talked to @piosz and he was ok with the idea. Basically, too much time has gone without anyone willing to maintain these, so therefore we're removing them. We might be willing to add them back once we have a person that wants to work on it. And the code is preserved in earlier releases, commits so even though we remove this now, git has it in memory of course for future reference.


piosz:
LGTM

cc @DirectXMan12

PR #1468, opened 2017-01-16
Made authFile override logic not fallback to in-cluster config for standalone mode

shyamjvs:
Changed the function `GetKubeClientConfig`  to use `NewNonInteractiveClientConfig` instead of `NewNonInteractiveDeferredLoadingClientConfig` as we do not want to fallback to inClusterConfig if authFile is specified, for heapster standalone mode. This is required because heapster could be running on an overlay cluster and we do not want to fallback to inClusterConfig which would make heapster talk to the underlay apiserver instead of the overlay one.
(Inspired from this example https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/bootstrap.go#L149)

cc @piosz

### DIFF
--- a/common/kubernetes/configs.go
+++ b/common/kubernetes/configs.go
@@ -104,9 +104,19 @@ func GetKubeClientConfig(uri *url.URL) (*kube_client.Config, error) {
 		}
 
 		if authFile != "" {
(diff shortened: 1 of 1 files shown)


piosz:
LGTM

PR #1486, opened 2017-01-26
abstract the GetNodeLister() function

tianshapjq:
there are duplicated codes spreads in the source code, and I abstract into the util.GetNodeLister().
Is this ok?


DirectXMan12 (approved):
LGTM :+1:

In the future, these should probably be passed the lister instead.

Issue #1478, opened 2017-01-23
Problem with pushing multi-arch images to gcr.io 


piosz:
When I'm trying to push Heapster images after support for multiple architectures was introduced in #1387 I can see the following issue:
```
./manifest-tool push from-args --platforms linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x --template gcr.io/google_containers/heapster-ARCH:v1.3.0-beta.1  --target gcr.io/google_containers/heapster:v1.3.0-beta.1
INFO[0000] Retrieving digests of images...              
INFO[0002] Image "gcr.io/google_containers/heapster-amd64:v1.3.0-beta.1" is digest sha256:<hash>; size: 949 
INFO[0004] Image "gcr.io/google_containers/heapster-arm:v1.3.0-beta.1" is digest sha256:<hash>; size: 950 
INFO[0006] Image "gcr.io/google_containers/heapster-arm64:v1.3.0-beta.1" is digest sha256:<hash>; size: 950 
INFO[0008] Image "gcr.io/google_containers/heapster-ppc64le:v1.3.0-beta.1" is digest sha256:<hash>; size: 950 
INFO[0010] Image "gcr.io/google_containers/heapster-s390x:v1.3.0-beta.1" is digest sha256:<hash>; size: 950 
FATA[0030] Registry push unsuccessful: response 400: 400 Bad Request 
make: *** [push] Error 1
```

It seems that images for specific architectures was pushed but there is no generic image.

luxas:
I've contacted some gcr.io folks, I'm waiting for an answer from them. 
It works fine on docker hub and _should work_ on gcr as well, but doesn't

Meanwhile, heapster-grafana-amd64:v4.0.2 is available

whereisaaron:
Another limitation of gcr.io is it doesn't have any human way to find tags. You can use `docker search` to find images, but then you need to using the API to find tags, e.g. `curl https://gcr.io/v2/google-containers/heapster-influxdb-amd64/tags/list`, these appear to be the current tags.

```
gcr.io/google_containers/heapster-influxdb-amd64:v1.1.1
gcr.io/google_containers/heapster-grafana-amd64:v4.0.2
gcr.io/google_containers/heapster-amd64:v1.3.0-beta.1
```

piosz:
ping @luxas

luxas:
Sent a PR: #1502

luxas:
[...]

PR #1490, opened 2017-01-27
Update grafana image to a valid/working name:tag

rarneson:
Their is currently no image gcr.io/google_containers/heapster-grafana:v4.0.2. The container registry has only one image there, and that is tagged at v2.6.0-2. I was able to pull and apply to my cluster with this image.

### DIFF
--- a/deploy/kube-config/influxdb/grafana-deployment.yaml
+++ b/deploy/kube-config/influxdb/grafana-deployment.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: grafana
-        image: gcr.io/google_containers/heapster-grafana:v4.0.2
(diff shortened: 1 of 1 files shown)


rarneson:
I signed it!

rarneson:
My commit was originally made with a different user.email value in my git config. I updated it and pushed the updates. It should now match my CLA signatures.

DirectXMan12:
cc @piosz

mortensteenrasmussen:
Related to PR #1496

piosz:
Grafana is available in version `v4.0.2` but you need specify your architecture explicitly due to a problem with multi-arch release tool. See #1478. Sorry for the confusion.

PR #1496, opened 2017-02-02
Update influx yaml to default to amd64 images and update docs

### LINKED ISSUE #1495: InfluxDB yaml files doesn't work out of the box
When trying to set up heapster using the influxdb docs, the default setup doesn't work. The images specified doesn't exist on gcr.io (they only exist on specific architectures, eg. `amd64`, `arm`, `arm64` etc.

I'm proposing to add a default value using the `-amd64` image in the YAML files since that exists, and I assume most people is running amd64. 
Also if you _don't_ run amd64, you're generally more aware of it and is used to change images to suit your arch.

I'd like to submit a PR updating the defaults in the  `deploy/kube-config/influxdb` YAML files, as well as update the docs mentioning that amd64 is the default (so people can easily spot if they need to make changes to the YAML files.

This particular annoyance caught me off guard a week or two ago, and today I saw another person experiencing the same issue on slack.

mortensteenrasmussen:
Fixes #1495

### DIFF
--- a/deploy/kube-config/influxdb/grafana-deployment.yaml
+++ b/deploy/kube-config/influxdb/grafana-deployment.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: grafana
-        image: gcr.io/google_containers/heapster-grafana:v4.0.2
+        image: gcr.io/google_containers/heapster-grafana-amd64:v4.0.2
         ports:
           - containerPort: 3000
             protocol: TCP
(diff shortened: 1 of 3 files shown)


mortensteenrasmussen:
I signed it!

DirectXMan12:
@piosz there's been some upheaval with the images lately.  WDYT?

mortensteenrasmussen:
FWIW the images I changed in the YAML seems only to be released in a -$ARCH, meaning the image listed in the YAML doesn't even exist. 

That coupled with the poor searchability of gcr.io images/tags makes for a rather unfortunate first impression (IMHO).

DirectXMan12:
> That coupled with the poor searchability of gcr.io images/tags makes for a rather unfortunate first impression (IMHO)

Not in disagreement here, just wanted to check and make sure this matches the intended image setup ;-)

piosz:
LGTM, thanks!

PR #1502, opened 2017-02-06
Disabled manifest lists for now so we can release v1.3

luxas:
Although GCR [claims](https://cloud.google.com/container-registry/release-notes) they support manifest lists, they do not, at least not fully. The "fat manifest" aka manifest list which makes the docker registry point to the right layers depending on the host platform doesn't work at the moment on gcr.io.

While we're waiting for that, we have to use `-ARCH` suffixed images like other k8s images.
@piosz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,12 +82,21 @@ ifneq ($(OVERRIDE_IMAGE_NAME),)
 endif
 	rm -rf $(TEMP_DIR)
 
-push: ./manifest-tool gcr-login $(addprefix sub-push-,$(ALL_ARCHITECTURES))
-	./manifest-tool push from-args --platforms $(ML_PLATFORMS) --template $(PREFIX)/heapster-ARCH:$(VERSION) --target $(PREFIX)/heapster:$(VERSION)
+do-push:
+	docker push $(PREFIX)/heapster-$(ARCH):$(VERSION)
+ifeq ($(ARCH),amd64)
+# TODO: Remove this and push the manifest list as soon as it's working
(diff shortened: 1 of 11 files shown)


mortensteenrasmussen:
Related to #1496 (and others)

Makefile
```diff
@@ -82,12 +82,21 @@ ifneq ($(OVERRIDE_IMAGE_NAME),)
 endif
 	rm -rf $(TEMP_DIR)
 
-push: ./manifest-tool gcr-login $(addprefix sub-push-,$(ALL_ARCHITECTURES))
-	./manifest-tool push from-args --platforms $(ML_PLATFORMS) --template $(PREFIX)/heapster-ARCH:$(VERSION) --target $(PREFIX)/heapster:$(VERSION)
+do-push:
+	docker push $(PREFIX)/heapster-$(ARCH):$(VERSION)
+ifeq ($(ARCH),amd64)
```
piosz: IIUC you push amd64 image as `heapster:<version>` without explicit specifying architecture. Why do you need an architecture in Heapster manifest then?

Also why don't you have the same thing for grafana/influx?

piosz (changes requested):

Makefile
luxas: I did this for backwards compability since earlier versions are released as `heapster:${version}`

Grafana and influxdb completely changed image names, so it's safe to require the `-amd64` suffix there until gcr.io starts working as it should.

piosz (approved):

vielmetti:
This should be revisited now that GCR supports fat manifests.

luxas:
@vielmetti Please send a PR to fix this or talk to @mkumatag

PR #1508, opened 2017-02-08
Log error from influxdb

### LINKED ISSUE #1376: influxdb metrics and events sinks should log errors
If a write fails in influxdb we get nothing in heapster logs and can guess the error using tcpdump and logs in influxdb. 

The problem is this:
```
if _, err := sink.client.Write(bp); err != nil {
		if strings.Contains(err.Error(), dbNotFoundError) {
			sink.resetConnection()
		} else if _, _, err := sink.client.Ping(); err != nil {
			glog.Errorf("InfluxDB ping failed: %v", err)
			sink.resetConnection()
		}
	}

```

The error returned from the Write is never logged :(

I could submit PR if that helps :)

jonaz:
Resolves #1376

### DIFF
--- a/events/sinks/influxdb/influxdb.go
+++ b/events/sinks/influxdb/influxdb.go
@@ -155,6 +155,7 @@ func (sink *influxdbSink) sendData(dataPoints []influxdb.Point) {
 
 	start := time.Now()
 	if _, err := sink.client.Write(bp); err != nil {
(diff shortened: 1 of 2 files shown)


piosz:
LGTM, thanks! please sign the CLA.

jonaz:
@piosz  cla/linuxfoundation signed. How do i trigger re-check here?

piosz:
It's ok right now

PR #1512, opened 2017-02-09
the --source should be enclosed by double quotes

tianshapjq:
I'm not sure if this proposal is correct. 
In my test, if I place more than 1 query parameter in the --source flag, such as --source=kubernetes:http://localhost:8080?inClusterConfig=false&useServiceAccount=false --sink=log, the command will not recognize the following --sink flag, and will not start the Metric Sink. But if using --source="kubernetes:http://localhost:8080?inClusterConfig=false&useServiceAccount=false" --sink=log, the command will start initializing the Metric Sink. So I think it's better to enclose the content of --source by double quotes.
Anyone test this before?

### DIFF
--- a/docs/source-configuration.md
+++ b/docs/source-configuration.md
@@ -13,17 +13,17 @@ Kubernetes clusters.
 ### Kubernetes
 To use the kubernetes source add the following flag:
 
-	--source=kubernetes:<KUBERNETES_MASTER>[?<KUBERNETES_OPTIONS>]
+	--source="kubernetes:<KUBERNETES_MASTER>[?<KUBERNETES_OPTIONS>]"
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
This shouldn't matter if you're launching in a pod, I think.  What did your YAML look like?

tianshapjq:
@DirectXMan12 I just ran by the go-builded file for testing, and realized that may happen in a pod. I will check this in a running pod. BTW, where are the log files in a pod for Heapster? I couldn't find them in a pod :(

DirectXMan12:
they should be in the pod's standard output -- just `kubectl logs $HEAPSTER_POD_NAME`

tianshapjq:
@DirectXMan12 yeah, it worked fine in a pod. I will close this pr, thanks.

PR #1514, opened 2017-02-09
Rebase heapster image on scratch & don't run as root

timstclair:
As part of the effort behind https://github.com/kubernetes/kubernetes/issues/40248, I'm cutting out unnecessary dependencies from system containers by rebasing them on scratch (for containers without any external dependencies). In addition, heapster does not require any elevated privileges, so run as the `nobody` user instead of root.

For debugging scratch containers, see https://github.com/kubernetes/contrib/pull/2372

/cc @ixdy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,22 +29,6 @@ TEST_NAMESPACE=heapster-e2e-tests
 
 HEAPSTER_LDFLAGS=-w -X k8s.io/heapster/version.HeapsterVersion=$(VERSION) -X k8s.io/heapster/version.GitCommit=$(GIT_COMMIT)
 
-ifeq ($(ARCH),amd64)
-	BASEIMAGE?=busybox
-endif
-ifeq ($(ARCH),arm)
-	BASEIMAGE?=armhf/busybox
-endif
-ifeq ($(ARCH),arm64)
-	BASEIMAGE?=aarch64/busybox
-endif
-ifeq ($(ARCH),ppc64le)
-	BASEIMAGE?=ppc64le/busybox
-endif
-ifeq ($(ARCH),s390x)
-	BASEIMAGE?=s390x/busybox
-endif
-
 fmt:
 	find . -type f -name "*.go" | grep -v "./vendor*" | xargs gofmt -s -w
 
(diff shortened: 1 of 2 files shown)


deploy/docker/Dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM BASEIMAGE
+FROM scratch
 
 COPY heapster eventer /
 COPY ca-certificates.crt /etc/ssl/certs/
 
+#   nobody:nobody
+USER 65534:65534
```
luxas: Is this known to work universally?

Makefile
```diff
@@ -75,8 +75,6 @@ container:
 		&& GOARCH=$(ARCH) CGO_ENABLED=0 go build -ldflags \"$(HEAPSTER_LDFLAGS)\" -o /build/eventer k8s.io/heapster/events"
 
 	cp deploy/docker/Dockerfile $(TEMP_DIR)
-	cd $(TEMP_DIR) && sed -i -e "s|BASEIMAGE|$(BASEIMAGE)|g" Dockerfile
```
luxas: You can remove the above definitions of `BASEIMAGE` as well I think

timstclair: Thanks, done.

deploy/docker/Dockerfile
timstclair: Note: if you put a username:group here instead, it uses the mapping in /etc/passwd, but since this is from scratch, there is no mapping there. The alternative would be to make our own, but that seems unnecessary.

timstclair:
Friendly ping. I would like to get this in 1.6

luxas:
This has been LGTM since you fixed the nits.
Haven't tried it out manually though, which I assume you have...

DirectXMan12:
Does anybody have any last minute complaints, or are we good to merge this?

timstclair:
I'm rerunning the k8s e2e's with the most recent build if you want to hold off - I'll report my results here when it's done.

DirectXMan12:
SGTM

timstclair:
Tests pass. Good to go?

piosz:
LGTM, thanks!

PR #1515, opened 2017-02-10
pod-container-level rest APIs are not available before v1.2.0

tianshapjq:
In my test, I deployed heapster v1.2.0 and found that the pod-container-level rest apis were not supported, such as "/api/v1/model/namespaces/{namespace-name}/pods/{pod-name}/containers/" just returned 404 not found.
So I propose deviding the specifications about containers into pod-container-level and free-container-level, explaining the pod-container-level only supported since V1.3.0.

### DIFF
--- a/docs/model.md
+++ b/docs/model.md
@@ -54,11 +54,11 @@ pairs for the requested namespace-level metric, within the time range specified 
 `/api/v1/model/namespaces/{namespace-name}/pods/{pod-name}/metrics/{metric-name}?start=X&end=Y`: Returns a set of (Timestamp, Value) 
 pairs for the requested pod-level metric, within the time range specified by `start` and `end`. 
 
-### Container-level Metrics
+### Container-level Metrics 
 Container metrics and stats are accessible for both containers that belong to
 pods, as well as for free containers running in each node.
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
They actually were supported before then, IIRC, there was just period where they incorrectly got dropped.

DirectXMan12:
You should note that instead.

tianshapjq:
@DirectXMan12 So is this a bug in V1.2.0?

DirectXMan12:
yeah

tianshapjq:
I found that only the `/api/v1/model/namespaces/{namespace-name}/pods/{pod-name}/containers/` api is not supported, the others worked fine. Someone mistaked deleting the _/containers_ api in the source code. I note this in the doc. 
Do we need to make a new release like V1.2.1 to fix this?

DirectXMan12:
I think this looks good now.  Not sure if we really need v1.2.1 -- I think just fixing it in v1.3.0 is find.
WDYT @piosz ?

piosz:
I agree with @DirectXMan12. This will be avaialble in 1.3.0 and is not important enough to make a patch release.

piosz:
LGTM, thanks!

PR #1519, opened 2017-02-12
Image pull errors, removing amd64 fixes it

ChrisBuchholz:
The image `gcr.io/google_containers/heapster-amd64:v1.3.0-beta` doesn't seem to exist and I get an `ErrImagePull` when using the manifest. Removing the `-amd64` specification seems to work, although a better fix might be to actually create the image with `-amd64` if there is other images available than amd64 ones.


ChrisBuchholz:
I signed it!

piosz:
Closing as obsolete. Please use  `gcr.io/google_containers/heapster-amd64:v1.3.0-beta.1`

PR #1525, opened 2017-02-15
fixing elasticsearch and adding tests

### LINKED ISSUE #1211: Elasticsearch sink not working
I have setup an elk stack in its own namespace, aka `es`. And there is a service called: `elasticsearch.es.svc.cluster.local`.

For the sink option I simply use: `--sink=elasticsearch:http://elasticsearch.es.svc.cluster.local:9200". This does not work and i suppose it has to do with: 

```
    // Set the URL endpoints of the ES's nodes. Notice that when sniffing is
    // enabled, these URLs are used to initially sniff the cluster on startup.
    if len(opts["nodes"]) < 1 {
        return nil, fmt.Errorf("There is no node assigned for connecting ES cluster")
    }
```

In case I and the `?nodes...` param, then I get:

```
factory.go:74] Failed to create sink: failed to create ElasticSearch client: no Elasticsearch node available
```

I have also tried to play with the sniff, timeouts etc options, so I'm stuck for the time being and the elasticsearch sink does not work

(Question: the nodes paramater seems to be mandatory? According to the `if len(opts["nodes"]) < 1`? Why is that?)

astropuffin:
fixes [1211](https://github.com/kubernetes/heapster/issues/1211) and adds a test so it doesn't happen again.

Root cause:
Opaque would work if it was set, but uri.Parse does not actually set Opaque.

### DIFF
--- a/common/elasticsearch/elasticsearch.go
+++ b/common/elasticsearch/elasticsearch.go
@@ -127,8 +127,8 @@ func CreateElasticSearchService(uri *url.URL) (*ElasticSearchService, error) {
 	var startupFns []elastic.ClientOptionFunc
 	if len(opts["nodes"]) > 0 {
 		startupFns = append(startupFns, elastic.SetURL(opts["nodes"]...))
-	} else if uri.Opaque != "" {
-		startupFns = append(startupFns, elastic.SetURL(uri.Opaque))
+	} else if uri.Scheme != "" && uri.Host != "" {
+		startupFns = append(startupFns, elastic.SetURL(uri.Scheme+"://"+uri.Host))
 	} else {
 		return nil, fmt.Errorf("There is no node assigned for connecting ES cluster")
 	}
--- a/common/elasticsearch/elasticsearch_test.go
+++ b/common/elasticsearch/elasticsearch_test.go
@@ -96,3 +96,53 @@ func TestCreateElasticSearchServiceForDefaultClusterName(t *testing.T) {
 		t.Fatalf("cluster name is not equal. Expected: %s, Got: %s", ESClusterName, esSvc.ClusterName)
 	}
 }
+
+func TestCreateElasticSearchServiceSingleDnsEntrypoint(t *testing.T) {
+	clusterName := "sandbox"
+	esURI := fmt.Sprintf("https://foo.com:9200?"+
+		"esUserName=test&esUserSecret=password&maxRetries=10&startupHealthcheckTimeout=30&"+
+		"sniff=false&healthCheck=false&cluster_name=%s", clusterName)
+
+	url, err := url.Parse(esURI)
+	if err != nil {
+		t.Fatalf("Error when parsing URL: %s", err.Error())
+	}
+
+	esSvc, err := CreateElasticSearchService(url)
+	if err != nil {
+		t.Fatalf("Error when creating config: %s", err.Error())
+	}
+
+	expectedClient, err := elastic.NewClient(
+		elastic.SetURL("https://foo.com:9200"),
+		elastic.SetBasicAuth("test", "password"),
+		elastic.SetMaxRetries(10),
+		elastic.SetHealthcheckTimeoutStartup(30*time.Second),
+		elastic.SetSniff(false), elastic.SetHealthcheck(false))
+
+	if err != nil {
+		t.Fatalf("Error when creating client: %s", err.Error())
+	}
+
+	actualClientRefl := reflect.ValueOf(esSvc.EsClient).Elem()
+	expectedClientRefl := reflect.ValueOf(expectedClient).Elem()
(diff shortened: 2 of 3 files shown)


astropuffin:
Linux Foundation LCA is stuck in limbo for tech reasons. Can I get someone to review this in the meantime?

astropuffin:
@DirectXMan12 this should be sink/elasticsearch, not influxdb.

andyxning:
Will Review this later.

=======
Andy Xie


2017-02-16 4:24 GMT+08:00 Joseph Schneider <notifications@github.com>:

DirectXMan12:
> this should be sink/elasticsearch, not influxdb.

so it should, I think I was triaging too fast :-P

common/elasticsearch/elasticsearch_test.go
```diff
@@ -96,3 +96,56 @@ func TestCreateElasticSearchServiceForDefaultClusterName(t *testing.T) {
...
+
+	actualClientRefl := reflect.ValueOf(esSvc.EsClient).Elem()
+	expectedClientRefl := reflect.ValueOf(expectedClient).Elem()
+
+	if actualClientRefl.FieldByName("basicAuthUsername").String() != expectedClientRefl.FieldByName("basicAuthUsername").String() {
+		t.Fatal("basicAuthUsername is not equal")
+	}
+	if actualClientRefl.FieldByName("basicAuthUsername").String() != expectedClientRefl.FieldByName("basicAuthUsername").String() {
+		t.Fatal("basicAuthUsername is not equal")
+	}
```
andyxning: The above check for `basicAuthUsername` duplicates.

common/elasticsearch/elasticsearch.go
```diff
@@ -127,8 +127,8 @@ func CreateElasticSearchService(uri *url.URL) (*ElasticSearchService, error) {
 	var startupFns []elastic.ClientOptionFunc
 	if len(opts["nodes"]) > 0 {
 		startupFns = append(startupFns, elastic.SetURL(opts["nodes"]...))
-	} else if uri.Opaque != "" {
-		startupFns = append(startupFns, elastic.SetURL(uri.Opaque))
+	} else if uri.Scheme != "" && uri.Host != "" {
```
andyxning: SGTM.

BTW, i think we need to document this behavior in the elasticsearch sink document.

andyxning:
@astropuffin Seems you have not successfully signing linux foundation cla yet. You need to do that first. :)

andyxning:
@astropuffin 

We need add more info for the root cause. This is useful for us to firstly get what this PR will do and why the original code can not satisfy our goal. :) 

> Root cause:
Opaque would work if it was set, but uri.Parse does not actually set Opaque.

According to [golang url doc](https://golang.org/pkg/net/url/#URL), Opaque will only be setted when uri is in the format like below:
> scheme:opaque[?query][#fragment]

AlmogBaku:
Sgtm. OP please address @andyxning notes

astropuffin:
hey. still working with my VP on the linux foundation signing. Jumping back in on this.

astropuffin:
@andreykurilin re: comment
empirically, when I set the config like so: `--sink=elasticsearch:http://elasticsearch.example.com:9200?esUserName=$ES_USER&esUserSecret=$ES_PASS&sniff=false&maxRetries=1` then the opaque is not set. I was able to confirm this is the behavior of `net/url`'s parse method.

not fully following what the question/action item is here. I hope that addressed it?

andyxning:
@astropuffin Yes, you're right. This PR will solve your empirical usage. :) 

The last step you need is to resolve your linuxfoundation CLA. :)

AlmogBaku:
@astropuffin can your please sign the cla?

DirectXMan12:
@piosz this should merge before 1.3.

astropuffin:
Hey, trying to get the CLA signed. Honestly, I have no idea what the problem is with the signing process. It seems to be stuck in limbo even though we've signed it.

astropuffin:
ok there. finally figured it out. :/

andyxning:
@astropuffin Overall LGTM. Just two nits
* could you please squeeze you commits with `git rebase`. :)
* IMO, we need to document this usage when just using one es backend.

astropuffin:
Ah, I assumed that it would be squashed on merge. I'll add a few lines of documentation and rebase. :)

piosz:
@astropuffin ETA? I want to cut release ASAP and this is the only PR I'm waiting for.

astropuffin:
ETA 30min

piosz:
@k8s-bot ok to test

astropuffin:
rebased and with docs

piosz:
Thanks!

Issue #1516, opened 2017-02-10
Graphite sink - cluster.cpu.usage_rate is 2x the sum what of each node reports.


jbehrends:
heapster-amd64:v1.3.0-beta.1

Graphite sink:

**kubernetes.cluster.cpu.usage_rate** is reporting a value double of **sumSeries(kubernetes.nodes.*.cpu.usage_rate)**

When comparing **.limit** and **.request** values the same way, they match.

theairkit:
@jbehrends Hi, I'll see it!

jsoriano:
In case it gives any clue... where I'm using the riemann sink to forward to graphite I also see this behaviour, `kubernetes.X.cluster.cpu.usage_rate` doubles any of the sums of nodes, pods or namespaces: 
* `sum(kubernetes.X.nodes.*.cpu.usage_rate)`
* `sum(kubernetes.X.nodes.*.pods.*.*.cpu.usage_rate)`
* `sum(kubernetes.X.namespaces.*.cpu.usage_rate)`

@DirectXMan12 could you find someone to try it with other sinks to see if it also happens?

The sink is not doing any operation with the data provided... I'm wondering if the problem could be in the cluster aggregator.

theairkit:
@jbehrends @jsoriano 
Yes, looks like that is not any sink (graphite, riemann etc.) bug,
sink does not do any opertion with data, its only 'proxy' them...

You may create issue on heapster more generally, not 'sink-related'.

(Anyway, for now I'll investigating this behavior)

theairkit:
@jbehrends @jsoriano 
Seems like problem in cluster aggregator, watch PR: https://github.com/kubernetes/heapster/pull/1529

I think, this issue may be closed

jbehrends:
UPDATE

I just pulled down the latest build of this container and tested sine the referenced PR was merged:
gcr.io/google-containers/heapster:v1.3.0

The cluster cpu_usage rate is no longer 2x the sum of the nodes cpu_usage rate, but it's now lower!  Not by a large margin but enough to not look correct.  See screenshot below:

![image](https://cloud.githubusercontent.com/assets/1915783/24160562/71105608-0e1f-11e7-928a-d4938368d481.png)

The drop in cluster usage_rate is when I deployed the updated heapster version.

theairkit:
@jbehrends hmm, I'll investigate it!

PR #1529, opened 2017-02-15
Fix cpu usage_rate calculation

ghost:
Hi
This patchset fixes cpu usage_rate, calculation collected by cluster aggregator.

For now there are difference between:
```
kubernetes.cluster.cpu.usage_rate 
```
and
```
sumSeries(kubernetes.nodes.*.cpu.usage_rate)
```
(I use graphite as datasink)

If I understand correctly, it is need to change condition in cluster_aggregator.go from
```
metricSetType == core.MetricSetTypeNamespace || metricSetType == core.MetricSetTypeSystemContainer
```
to
```
metricSetType == core.MetricSetTypeNamespace
```
See also https://github.com/kubernetes/heapster/issues/1516

Thanks in advance!


DirectXMan12:
I think this makes sense -- IIRC, "system containers" can actual contain other containers due to cAdvisor's definition of "system containers".  cc @piosz PTAL as well.

piosz:
LGTM, thanks!

PR #1536, opened 2017-02-20
Heapster Librato sink

johanneswuerbach:
Sink for https://www.librato.com/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ container:
 
 	docker build --pull -t $(PREFIX)/heapster-$(ARCH):$(VERSION) $(TEMP_DIR)
 ifneq ($(OVERRIDE_IMAGE_NAME),)
-	docker tag -f $(PREFIX)/heapster-$(ARCH):$(VERSION) $(OVERRIDE_IMAGE_NAME)
+	docker tag $(PREFIX)/heapster-$(ARCH):$(VERSION) $(OVERRIDE_IMAGE_NAME)
 endif
 
 ifndef DOCKER_IN_DOCKER
@@ -127,6 +127,9 @@ push-influxdb:
 
 push-grafana:
 	PREFIX=$(PREFIX) make -C grafana push
+
+push-override:
+	docker push $(OVERRIDE_IMAGE_NAME)
 
 gcr-login:
 ifeq ($(findstring gcr.io,$(PREFIX)),gcr.io)
--- a/common/librato/librato.go
+++ b/common/librato/librato.go
@@ -1,0 +1,134 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package librato
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+type Measurement struct {
+	Name  string            `json:"name,omitempty"`
+	Value float64           `json:"value,omitempty"`
+	Tags  map[string]string `json:"tags,omitempty"`
+	Time  int64             `json:"time,omitempty"`
+}
+
+type request struct {
+	Tags         map[string]string `json:"tags,omitempty"`
+	Measurements []Measurement     `json:"measurements,omitempty"`
+}
+
+type Client interface {
+	Write([]Measurement) error
+}
+
+type LibratoClient struct {
+	httpClient *http.Client
+	config     LibratoConfig
+}
+
+func (c *LibratoClient) Write(measurements []Measurement) error {
+	b, err := json.Marshal(&request{
+		Measurements: measurements,
+		Tags:         c.config.Tags,
+	})
+	if nil != err {
+		return err
+	}
+	req, err := http.NewRequest(
+		"POST",
+		c.config.API+"/v1/measurements",
+		bytes.NewBuffer(b),
+	)
+	if nil != err {
+		return err
+	}
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "heapster")
+	req.SetBasicAuth(c.config.Username, c.config.Token)
+	_, err = c.httpClient.Do(req)
+	return err
+}
+
+type LibratoConfig struct {
+	Username string
+	Token    string
+	API      string
+	Prefix   string
+	Tags     map[string]string
+}
+
+func NewClient(c LibratoConfig) *LibratoClient {
+	var netTransport = &http.Transport{
+		Dial: (&net.Dialer{
+			Timeout: 5 * time.Second,
+		}).Dial,
+		TLSHandshakeTimeout: 5 * time.Second,
+	}
+	var httpClient = &http.Client{
+		Timeout:   time.Second * 10,
+		Transport: netTransport,
+	}
+
+	client := &LibratoClient{httpClient: httpClient, config: c}
+	return client
+}
+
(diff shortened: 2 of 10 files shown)


johanneswuerbach:
I signed it!

dieterdemeyer:
@johanneswuerbach we could also use this sink if it gets merged.
Do you happen to have a timeline on this ? If I can help in any way, please let me know.

johanneswuerbach:
@dieterdemeyer I haven't gotten any feedback from any k8s maintainer yet, so hard to say.

We are currently building a custom heapster from this branch using:
`OVERRIDE_IMAGE_NAME=org/custom-heapster:test-7 make container push-override`

dieterdemeyer:
@johanneswuerbach how exactly should I use that ? I'm just starting with Kubernetes, so bear with me..

johanneswuerbach:
@dieterdemeyer you can use a config similar to the once you can find here https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/influxdb/heapster-controller.yaml, but instead of the official image you can use your custom own and instead of influxdb, you can use the librato sink

dieterdemeyer:
@johanneswuerbach Thanks, I'll give it a try..

Makefile
```diff
@@ -89,7 +89,7 @@ container:
 
 	docker build --pull -t $(PREFIX)/heapster-$(ARCH):$(VERSION) $(TEMP_DIR)
 ifneq ($(OVERRIDE_IMAGE_NAME),)
-	docker tag -f $(PREFIX)/heapster-$(ARCH):$(VERSION) $(OVERRIDE_IMAGE_NAME)
+	docker tag $(PREFIX)/heapster-$(ARCH):$(VERSION) $(OVERRIDE_IMAGE_NAME)
```
DirectXMan12: did you mean to change the Makefile?

common/librato/librato.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
```
DirectXMan12: this doesn't need to be in common unless you also plan on writing an eventer sink.

DirectXMan12:
Following up, the code looks decent.  It should be good to merge once we've confirmed that it passes tests and you fix the nits.

DirectXMan12:
@k8s-bot ok to test

Makefile
johanneswuerbach: This line never worked for me as https://docs.docker.com/engine/reference/commandline/tag/ has no `-f` flag. I can send a separate PR if you prefer that.

common/librato/librato.go
johanneswuerbach: I was planning to, but https://www.librato.com/docs/api/#annotations isn't ready for tags yet. I'll contact librato to find out a timeframe.

johanneswuerbach:
Thanks @DirectXMan12 and no worries, I updated the sink owners file and left some comments. Let me know if anything else is required :-)

common/librato/librato.go
DirectXMan12: ack, alright :-)

Makefile
DirectXMan12: huh, you appear to be correct, this should be fine.

dieterdemeyer:
@johanneswuerbach I used the master branch of kubernetes/heapster and built a new image with the following command:

`
OVERRIDE_IMAGE_NAME=myorg/heapster:1.3.0-SNAPSHOT make container push-override
`

I'm trying to use the librato sink but I'm getting the following error message when the pod starts:
`
Failed to create sink: Sink not recognized: "librato
`

I'm using the following configuration in my container command:
`
--sink="librato:?username=<username>&token=<token>&prefix=k8s&tags=cluster"
`

Could you tell me what I'm doing wrong ?

Thanks in advance.

dieterdemeyer:
@johanneswuerbach Never mind, just found the cause.
The documentation found [here](https://github.com/kubernetes/heapster/blob/master/docs/sink-configuration.md) describes quotes around the sink configuration but it actually shouldn't contain quotes.

johanneswuerbach:
Thanks @dieterdemeyer would you mind sending a PR removing the quotes? That would be great 👍

dieterdemeyer:
@johanneswuerbach Sure, will do..
Did you test the Librato sink with real live data ? I have some weird values in the metric k8s.cpu.usage...
Are you available in other channels like the Kubernetes slack instance for a small talk about this ?

johanneswuerbach:
Yes we are currently using it with 3 clusters, but I didn't check all the values yet. Not in the slack yet, but will join once again on a computer

Makefile
piosz: FYI, this broke e2e tests
```
docker tag gcr.io/google_containers/heapster-amd64:v1.3.0-beta.1 heapster:e2e_test
Error response from daemon: Conflict: Tag heapster:e2e_test is already set to image 1b32c9c98284, if you want to replace it, please use -f option
Makefile:68: recipe for target 'container' failed
make[1]: *** [container] Error 1
make[1]: Leaving directory '/go/src/k8s.io/heapster
```
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/heapster/1572/pull-heapster-e2e/238/

I'm going to revert this change (only Makefile part).

piosz: ref #1572

mashayev:
Hii @johanneswuerbach @dieterdemeyer I got the same issue with `Failed to create sink: Sink not recognized: librato`

I removed the quotes from sink and it's the same issue. Maybe `heapster-amd64:v1.3.0-beta.1` image need to be updated?

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: heapster
  namespace: kube-system
spec:
  replicas: 1
  template:
    metadata:
      labels:
        task: monitoring
        k8s-app: heapster
    spec:
      containers:
      - name: heapster
        image: gcr.io/google_containers/heapster-amd64:v1.3.0-beta.1
        imagePullPolicy: Always
        command:
        - /heapster
        - --source=kubernetes:https://kubernetes.default
        - --sink=librato:?username=xyz&token=secret&prefix=k8s&tags=cluster&tag_cluster=staging
        volumeMounts: 
          - name: ssl-certs
            mountPath: "/etc/ssl/certs/ca-certificates.crt"
            readOnly: true
      volumes: 
        - name: ssl-certs
          hostPath: 
            path: "/etc/pki/tls/certs/ca-bundle.crt"
```

johanneswuerbach:
@mashayev beta.1 doesn't contain the librato sink, but the recently released https://github.com/kubernetes/heapster/releases/tag/v1.3.0 does. Could you double check?

mashayev:
@johanneswuerbach Thanks for the help, it works :)

mashayev:
@johanneswuerbach One more question about "The librato sink currently only works with accounts, which support tagged metrics."
If we are using source and not tags(We signed up on or before January 24th, 2017 12:00 PST), it means the sink wont work at all?

Any plan to support both??

johanneswuerbach:
I'm only using tagged metrics and as far as I know source based metrics are also more or less deprecated. As far as I know you can just sent an email to librato support and your account will be whitelisted to use both.

mashayev:
@johanneswuerbach Librato team have enabled the Tags for my account, the Sink is working now for me.
I have another question: is there any option to change the metrics name in the chart?
for example I get this name in the chart:
`i-1q2w3edfrtt.ip-1-1-1-1.ec2.internal.node.ip-1-1-1-1.beta.kubernetes.io_arch:amd64_beta.kubernetes.io_instance-type:m3.large_beta.kubernetes.io_os:linux_failure-domain.beta.kubernetes.io_region:us-east-1_failure-domain.beta.kubernetes.io_zone:us-east-1c_kubernetes.io_hostname:ip-1-1-1-1`

I prefer to get the name of one of the tags or combination of them for example. something like:
`<CLUSTER_REGION>.<ENVIRONMENT>.<POD_NAME>.<POD_NAMESPACE>`
Below is an example of how it works when I send the metrics with Collectd(I can control the way the name will be displayed), is it possible with Librato Sink?

image

Thanks for the help.

johanneswuerbach:
Not sure, as far as I know the metric name is auto-generated from all received tags, but I guess you could ask librato support for clarification.

The sink is only forwarding each metric with all submitted tags to https://www.librato.com/docs/api/#create-a-measurement

mashayev:
Hi @johanneswuerbach,
I wonder if it is possible to filter which metrics will be sent to Librato and which not. From what I see there is no such an option. Can you confirm?

johanneswuerbach:
Yes, that is correct.

PR #1537, opened 2017-02-20
Use types from k8s.io/metrics

DirectXMan12:
This PR changes Heapster over to the types from k8s.io/metrics when serving metrics from the API server.


DirectXMan12:
cc @piosz PTAL

DirectXMan12:
client-go finally synced, so this is unblocked

piosz:
LGTM, thanks a lot for doing it!

PR #1541, opened 2017-02-22
add rss and cache

andyxning:
This PR adds RSS(`memory/rss`) and cache(`memory/cache`) memory metrics.

Currently, `memory/usage` contains both RSS and cache, we can not know what exactly each uses. There are situations that `memory/usage` is the same as `memory/limit` with containers running normally.

This will make it difficult to distinguish whether we should enlarge memory limit in order to make containers live normally instead of OOM killed. We should use `memory/RSS` and `memory/limit` as the judgement metrics, i.e., when `memory/RSS` is approximate to `memory/limit`, we need to enlarge memory limit. This is also compatible with usage behavior in bare machine.

/cc @piosz @DirectXMan12 @AlmogBaku @kubernetes/heapster-reviewers

### DIFF
--- a/docs/storage-schema.md
+++ b/docs/storage-schema.md
@@ -26,6 +26,8 @@ Heapster exports the following metrics to its backends.
 | memory/page_faults_rate | Number of page faults per second. |
 | memory/request | Memory request (the guaranteed amount of resources) in bytes. |
 | memory/usage | Total memory usage. |
+| memory/cache | Cache memory usage. |
+| memory/rss | RSS memory usage. |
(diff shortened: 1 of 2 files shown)


DirectXMan12 (approved):
LGTM :+1:

DirectXMan12:
looks good in general to me.

Any thoughts, @piosz

andyxning:
Friendly ping. :) @piosz

DirectXMan12:
merging, seems good to me.

piosz:
LGTM

PR #1555, opened 2017-03-03
Add additional logging around points of failure

DirectXMan12:
This commit adds additional logging in the rate calculator and summary
source around potential points of failure.  Namely, additional logging
was introduced when the rate calculator skips calculation, and the
summary source was enhanced to log additional information about fetching
and processing data from the Kubelet at levels 9 and 10 (the latter of
which simply logs the entire response from the Kubelet).


DirectXMan12:
@piosz PTAL

I've found these additions to be *very* useful in debugging weird issues with Heapster that we've had over the past few months, so I figured they should probably go in to Heapster.

piosz:
LGTM

PR #1563, opened 2017-03-07
Use -race on amd64 only when running tests

vivkong:
Use -race on amd64 only, -race is only supported by go test on amd64

This will allow tests to be run on s390x.  This is a follow-up to https://github.com/kubernetes/heapster/pull/1357

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,11 @@ sanitize:
 	hooks/run_vet.sh
 
 test-unit: clean sanitize build
+ifeq ($(ARCH),amd64)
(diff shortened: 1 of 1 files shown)


vivkong:
I signed it!

DirectXMan12:
@piosz @luxas opinions on this?

luxas:
I'm fine with this :+1:

luxas (approved):

DirectXMan12:
seems fine to me as well, merging.

piosz:
LGTM

PR #1570, opened 2017-03-16
correct the order of expected and actual

tianshapjq:
correct the order of params to fit the expected and actual places.


piosz:
LGTM, thanks!

PR #1571, opened 2017-03-16
Validate metric resolution

bruceauyeung:
this PR fixes a bug when metric_resolution is less than DefaultScrapeOffset. When metric_resolution is less than DefaultScrapeOffset, some MetricSets will not be scraped from sources

in `Housekeep()` function ( `manager.go` ) , let's assume:
1. we ignore execution time of one iteration of `Housekeep()` function
2. metric_resolution: `4s`
3. now time in first iteration:  `13:03` ( format  minutes : seconds)

then in first iteration:
start : `13:00`
end: `13:04`

and then the second iteration:
now: `13:09`
start: `13:08`
end: `13:12`

we can see MetricSets between 13:04 and 13:08 will not be scraped.

if metric_resolution is equal to or greater than scrapeOffset, this case will not happen.

### DIFF
--- a/metrics/manager/manager_test.go
+++ b/metrics/manager/manager_test.go
@@ -23,15 +23,15 @@ import (
 )
 
 func TestFlow(t *testing.T) {
-	source := util.NewDummyMetricsSource("src", time.Millisecond)
+	source := util.NewDummyMetricsSource("src", 500*time.Millisecond)
 	sink := util.NewDummySink("sink", time.Millisecond)
-	processor := util.NewDummyDataProcessor(time.Millisecond)
+	processor := util.NewDummyDataProcessor(500 * time.Millisecond)
 
-	manager, _ := NewManager(source, []core.DataProcessor{processor}, sink, time.Second, time.Millisecond, 1)
+	manager, _ := NewManager(source, []core.DataProcessor{processor}, sink, MinMetricResolution, time.Millisecond, 1)
 	manager.Start()
 
-	// 4-5 cycles
-	time.Sleep(time.Millisecond * 4500)
+	// 4.5 * MinMetricResolution, about 4-5 cycles
+	time.Sleep(4500 * 30 * time.Millisecond)
 	manager.Stop()
 
 	if sink.GetExportCount() < 4 || sink.GetExportCount() > 5 {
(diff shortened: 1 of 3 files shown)


piosz:
@k8s-bot ok to test

DirectXMan12:
Just FYI, metric resolution really shouldn't be lower than 30s anyway, because of issues with cAdvisor.

bruceauyeung:
@DirectXMan12 would you please give a link about the bug of cAdvisor ? i'd like to know why. thanks!

bruceauyeung:
@DirectXMan12 fixed as you sugguested

bruceauyeung:
@DirectXMan12 fixed test case error

metrics/manager/manager_test.go
```diff
@@ -27,11 +27,11 @@ func TestFlow(t *testing.T) {
 	sink := util.NewDummySink("sink", time.Millisecond)
 	processor := util.NewDummyDataProcessor(time.Millisecond)
 
-	manager, _ := NewManager(source, []core.DataProcessor{processor}, sink, time.Second, time.Millisecond, 1)
+	manager, _ := NewManager(source, []core.DataProcessor{processor}, sink, MinMetricResolution, time.Millisecond, 1)
 	manager.Start()
 
 	// 4-5 cycles
-	time.Sleep(time.Millisecond * 4500)
+	time.Sleep(time.Millisecond * 4500 * 30)
```
DirectXMan12: I'm not entirely certain what the intention here is of multiplying by 30.  Please, let's not have magic numbers (at least not without better comments).

DirectXMan12 (changes requested):
small change, but otherwise looks good

metrics/manager/manager_test.go
bruceauyeung: @DirectXMan12 already done, and the test case error gone.

bruceauyeung:
@DirectXMan12 i already have fixed as you requested, would you mind take an another look ?

metrics/manager/manager_test.go
```diff
@@ -23,15 +23,15 @@ import (
...
-	processor := util.NewDummyDataProcessor(time.Millisecond)
+	processor := util.NewDummyDataProcessor(500 * time.Millisecond)
 
-	manager, _ := NewManager(source, []core.DataProcessor{processor}, sink, time.Second, time.Millisecond, 1)
+	manager, _ := NewManager(source, []core.DataProcessor{processor}, sink, MinMetricResolution, time.Millisecond, 1)
 	manager.Start()
 
-	// 4-5 cycles
-	time.Sleep(time.Millisecond * 4500)
+	// 4.5 * MinMetricResolution, about 4-5 cycles
```
DirectXMan12: now that I think about it, I don't see much reason to actually respect the minimum metric resolution with the dummy sink -- this is just going to make the tests take *way* longer to run.

DirectXMan12: (so, I don't think there's much reason to change the tests here)

DirectXMan12 (changes requested):

andyxning:
@DirectXMan12 
> metric resolution really shouldn't be lower than 30s anyway, because of issues with cAdvisor.

Could you please point out the what issues prevent metric resolution from being set lower than 30s.

DirectXMan12:
@andyxning it's been a few months since I went through that code, but IIRC, most of the important bits are here: https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/google/cadvisor/manager/container.go#L359.

Kubelet sets the default housekeeping "interval" as 10 seconds, sets the max dynamic housekeeping interval as 15 seconds (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/cadvisor_linux.go#L54), and enables dynamic housekeeping.

The duration between housekeeping intervals can be, even without dynamic housekeeping, up to 2 times the default housekeeping interval (that's what `jitter` does), meaning even if we turned off dynamic housekeeping, we get an interval that's actually anywhere between 10s and 20s.

With dynamic housekeeping, we can increase the housekeeping interval from 10s to 15s.  This is applied *before* the jitter, so theoretically, we can get an actual time between housekeepings of up to 30s.

piosz:
Closing, as it's inactive for 6+ months. Please reopen if you still want to work on this.

PR #1577, opened 2017-03-17
instructions to enable oldtimer API

cainelli:
Hi there, I've added basic instructions on how to enable the old-timer API based on #1172. It took me some time to search on the repo. I think others may found useful to see this present on the proposal doc.

Thanks.

### DIFF
--- a/docs/proposals/old-timer.md
+++ b/docs/proposals/old-timer.md
@@ -48,6 +48,8 @@ When used simply to retrieve historical data points, Oldtimer will return the
 same types as the model API.  When the used to retrieve aggregations, Oldtimer
 will return special data types detailed under the "Return Types" section.
 
+Oltimer API can be enabled by passing --historical_source with one of the URIs used in the --sink argument.
+
 ### Paths
 
 `/api/v1/historical/{prefix}/metrics/`: Returns a list of all available


cainelli:
I signed it!

DirectXMan12:
Oldtime was never widely implemented, and is very likely going to be deprecated and removed over the next couple of releases.  I'm not entirely certain it's worth providing these instructions (especially without a deprecation notice).

@piosz WDYT

cainelli:
So will the main endpoints extract metrics older than 15 min. or if we want to do that we'll have to query the backend sink?

DirectXMan12:
> So will the main endpoints extract metrics older than 15 min. or if we want to do that we'll have to query the backend sink?

Eventually we'll have something along the same lines as the new custom metrics API -- a set of type definitions and API paths that will be implemented separately for each sink, except in a clearly historical context (as opposed to the custom metrics API, which is for recent metrics).

docs/proposals/old-timer.md
```diff
@@ -48,6 +48,8 @@ When used simply to retrieve historical data points, Oldtimer will return the
 same types as the model API.  When the used to retrieve aggregations, Oldtimer
 will return special data types detailed under the "Return Types" section.
 
+Oltimer API can be enabled by passing --historical_source with one of the URIs used in the --sink argument.
```
AlmogBaku: `Oldtimer` was misspelled

piosz:
https://github.com/kubernetes/heapster/pull/1577#issuecomment-288213958
+1

PR #1581, opened 2017-03-20
Fix retention policy configuration

yannrouillard:
This pull request fixes several issues with the current code used to configure retention:
 - createRetentionPolicy only being called in case of errors at database creation time,
 - retention policy name conflict with the default policy,
 - retention policy configuration not implemented for events which use the same database.

### DIFF
--- a/metrics/sinks/influxdb/influxdb.go
+++ b/metrics/sinks/influxdb/influxdb.go
@@ -198,36 +198,21 @@ func (sink *influxdbSink) createDatabase() error {
 	if sink.dbExists {
 		return nil
 	}
+
 	q := influxdb.Query{
-		Command: fmt.Sprintf(`CREATE DATABASE %s WITH NAME "default"`, sink.c.DbName),
+		Command: fmt.Sprintf(`CREATE DATABASE %s WITH DURATION %s REPLICATION 1 NAME "default"`, sink.c.DbName, sink.c.RetentionPolicy),
 	}
 
-	if resp, err := sink.client.Query(q); err != nil {
-		if !(resp != nil && resp.Err != nil && strings.Contains(resp.Err.Error(), "already exists")) {
-			err := sink.createRetentionPolicy()
-			if err != nil {
-				return err
-			}
+	if resp, err := sink.client.Query(q); err != nil || resp.Error() != nil {
+		if err != nil {
+			return err
+		} else {
+			return resp.Error()
 		}
 	}
 
 	sink.dbExists = true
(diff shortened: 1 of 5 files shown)


yannrouillard:
I signed it!

metrics/sinks/influxdb/influxdb.go
```diff
@@ -198,16 +198,16 @@ func (sink *influxdbSink) createDatabase() error {
...
 	}
 
-	if resp, err := sink.client.Query(q); err != nil {
-		if !(resp != nil && resp.Err != nil && strings.Contains(resp.Err.Error(), "already exists")) {
-			err := sink.createRetentionPolicy()
-			if err != nil {
-				return err
-			}
+	if resp, err := sink.client.Query(q); err != nil || resp.Error() != nil {
+		if err != nil {
```
DirectXMan12: don't we still only want to return the error if it's not "already exists"?

DirectXMan12 (changes requested):
I have at least one comment here, and I'd like to have someone more familiar with InfluxDB take a look.

DirectXMan12:
@mpvaniersel @piosz PTAL

metrics/sinks/influxdb/influxdb.go
yannrouillard: sure but influxdb doesn't return an error if the db already exists so we are already fine.
Will find the doc reference to that fact (and I tested it, otherwise it would fail the second time it is launched)

yannrouillard: From https://docs.influxdata.com/influxdb/v1.2/query_language/database_management/#create-database 
"A successful CREATE DATABASE query returns an empty result. If you attempt to create a database that already exists, InfluxDB does nothing and does not return an error."

However this message is not present in doc of influxdb v0.9 and I've tested it with v1.2.
I can do additional tests with older version of influxdb.
What version of influxdb do we support?

mpvaniersel:
Sorry, I don't have my dev environment for heapster set up anymore. I can't be much help I'm afraid.

metrics/sinks/influxdb/influxdb.go
piosz: We support [v1.1.1](https://github.com/kubernetes/heapster/blob/master/influxdb/Makefile#L22)

dieterdemeyer:
Is it expected behaviour that when you change the retention policy in heapster, the retention policy won't be updated ?
Currently, it only sets the retention policy to the correct duration value when no k8s database exists.

DirectXMan12:
yes, that's expected, I believe

adrianchifor:
Any update on this?

PR #1590, opened 2017-04-02
VPA Recommender Sink and Recommender JSON Client added.

KarolKraskiewicz:
This is a DataSink implementation, which will be sending container utilization data to VPA Recommender, described in [Pull Request 338](https://github.com/kubernetes/community/pull/338)

@mwielgus @kgrygiel and @piosz please review.

### DIFF
--- a/metrics/core/types.go
+++ b/metrics/core/types.go
@@ -117,14 +117,20 @@ func (this *LabeledMetric) GetValue() interface{} {
 	}
 }
 
+//all the metrics from single resource (like pod or container)
 type MetricSet struct {
-	CreateTime     time.Time
-	ScrapeTime     time.Time
-	MetricValues   map[string]MetricValue
-	Labels         map[string]string
+	//time when measured resource was created
+	CreateTime time.Time
+	ScrapeTime time.Time
+	//standard metrics of resource, mapped by metric name from metrics.go
+	MetricValues map[string]MetricValue
+	//labels for entire object (not single value)
+	Labels map[string]string
+	//metrics for sub-resources (like disks)
 	LabeledMetrics []LabeledMetric
 }
 
+//Representation of all the metrics from single "scrape"
 type DataBatch struct {
 	Timestamp time.Time
 	// Should use key functions from ms_keys.go
--- a/metrics/sinks/vpa/container_utilization.go
+++ b/metrics/sinks/vpa/container_utilization.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vpa
+
+import (
+	"github.com/pkg/errors"
+	"k8s.io/heapster/metrics/core"
+	"time"
+)
+
+type utilizationSnapshotSet struct {
+	Containers []*containerUtilizationSnapshot `json:"containers"`
+}
+
+func (set *utilizationSnapshotSet) addContainer(utilizationSnapshot *containerUtilizationSnapshot) {
+	set.Containers = append(set.Containers, utilizationSnapshot)
+}
+
+type containerUtilizationSnapshot struct {
+	CreateTime     time.Time `json:"createTime"`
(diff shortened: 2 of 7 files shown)


KarolKraskiewicz:
I signed it! (The cla)

KarolKraskiewicz:
I decided to push 1 updated to Recommender, consolidating info about all containers, instead of sending single request per container. It should save some time due to network latency and simplify processing on Recommender side. 

It means however, we will be able to send information about only up to 20 000 containers:
- info about single container = 512B
- 512B x 20 000 = 10MB which is a limit of body size when handling http post request in go.

The 20k containers limit will be raised when sink/recommender communication protocol is changed from JSON to Protocol Buffer. It should get us to up to 30-40k and from here we can go further by switching to multipart post.

DirectXMan12:
Is this intended to be a temporary solution, or a long-term one?  In the long term, I don't think we should tie the VPA to Heapster, since Heapster will eventually no longer be a necessary cluster component.

Is there any reason the VPA can't pull this information from the resource metrics API instead?  That way, everything will continue to work if people run metrics-server or something else.

metrics/core/types.go
```diff
@@ -117,14 +117,20 @@ func (this *LabeledMetric) GetValue() interface{} {
 	}
 }
 
+//all the metrics from single resource (like pod or container)
 type MetricSet struct {
-	CreateTime     time.Time
```
mwielgus: Let's move these comments to a separate pr.

metrics/sinks/vpa/container_utilization.go
```diff
@@ -0,0 +1,84 @@
...
+
+type containerUtilizationSnapshot struct {
+	CreateTime     time.Time `json:"createTime"`
+	ScrapTime      time.Time `json:"scrapTime"`
+	ContainerName  string    `json:"containerName"`
+	ContainerImage string    `json:"containerImage"`
+	PodId          string    `json:"podId"`
+	// Amount of requested memory.
+	// Units: Bytes.
+	MemoryRequest int64 `json:"memoryRequest"`
```
mwielgus: I guess we should probably use maps here. CPU and Memory are not the only resources we may want to use.

metrics/sinks/vpa/container_utilization.go
```diff
@@ -0,0 +1,84 @@
...
+func (set *utilizationSnapshotSet) addContainer(utilizationSnapshot *containerUtilizationSnapshot) {
+	set.Containers = append(set.Containers, utilizationSnapshot)
+}
+
+type containerUtilizationSnapshot struct {
+	CreateTime     time.Time `json:"createTime"`
+	ScrapTime      time.Time `json:"scrapTime"`
+	ContainerName  string    `json:"containerName"`
+	ContainerImage string    `json:"containerImage"`
+	PodId          string    `json:"podId"`
```
mwielgus: PodID (as uid) is useless. Replace with podname + podnamespace.

metrics/sinks/vpa/container_utilization.go
```diff
@@ -0,0 +1,84 @@
...
+
+package vpa
+
+import (
+	"github.com/pkg/errors"
+	"k8s.io/heapster/metrics/core"
+	"time"
+)
+
+type utilizationSnapshotSet struct {
```
mwielgus: UpperCase

mwielgus: Add an information that this set may be only a pice of a multipart batch.

metrics/sinks/vpa/recommender.go
```diff
@@ -0,0 +1,76 @@
...
+			glog.Warningf("Unable to create utilization snapshot, due to error: %s", err.Error())
+		}
+	}
+
+	sink.sendAllSnapshots(&snapshotSet)
+}
+
+func (sink *recommenderSink) sendAllSnapshots(set *utilizationSnapshotSet) {
+	if len(set.Containers) > 0 {
+		_, err := sink.recommenderClient.SendJSON(set)
```
mwielgus: Add batching.

metrics/sinks/vpa/recommender.go
```diff
@@ -0,0 +1,76 @@
...
+func (sink *recommenderSink) Name() string {
+	return "VPA Recommender Sink"
+}
+
+func (sink *recommenderSink) Stop() {
+	// nothing needs to be done.
+}
+
+func isContainerMetricSet(metricSet *core.MetricSet) bool {
+	return metricSet.Labels[core.LabelContainerName.Key] != "" && metricSet.Labels[core.LabelContainerBaseImage.Key] != ""
```
mwielgus: Please check if we have a helper function for this.

mwielgus:
@DirectXMan12 
Yeah, unfortunately I have to agree with you :). We had a conversation with @piosz and @KarolKraskiewicz, @GFilipek and decided that reversing the data traffic makes sense.

DirectXMan12:
We should close this, then

PR #1610, opened 2017-04-24
[GKE] Export Cpu and Memory Requests

dashpole:
I am trying to expose metrics useful for node allocatable.  I am happy to add tests, but I didnt see any for other similar stats that we expose.

### DIFF
--- a/metrics/api/v1/api.go
+++ b/metrics/api/v1/api.go
@@ -49,7 +49,9 @@ func NewApi(runningInKubernetes bool, metricSink *metricsink.MetricSink, histori
 		gkeMetrics[val.Name] = val.MetricDescriptor
 	}
 	gkeMetrics[core.MetricCpuLimit.Name] = core.MetricCpuLimit.MetricDescriptor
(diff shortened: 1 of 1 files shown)


dashpole:
@piosz is this the correct way to expose pod requests in gke?

DirectXMan12:
Please note that this is specifically in relation to GKE in the PR title.

dashpole:
@piosz ping

DirectXMan12:
please also put `[GKE]` in the commit message title.

dashpole:
cc @dchen1107

dchen1107:
/lgtm

@piosz

dchen1107:
cc/ @fgrzadkowski

dashpole:
@piosz and I spoke offline.  Heapster will soon be replaced by the new Monitoring Server.  Since container requests and limits can be obtained using the kubernetes API, I plan to move container limits from querying Heapster, to using the API directly.  I also plan to add container requests during that process.  This will achieve my goals of adding container requests, and help facilitate the transition from heapster to the metrics server.  Closing this PR as it is no longer needed.

PR #1613, opened 2017-04-26
Fix exporting CPU metrics

aleksandra-malinowska:
Stackdriver sink took metric data from wrong field (FloatValue), causing it to export incorrect value. It should use IntValue instead.

cc @piosz

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -68,12 +68,6 @@ var (
 		Name:       "container.googleapis.com/container/uptime",
 	}
 
-	utilizationMD = &metricMetadata{
-		MetricKind: "GAUGE",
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Please actually write was was originally wrong and what the correct behavior is for changes like this (also, probably put "Stackdriver" somewhere in your commit message title).

aleksandra-malinowska:
@DirectXMan12 Thanks, I fixed description. Leaving the commit message as it is since it's already been merged (unless this is important enough to revert?)

DirectXMan12:
It's not important enough to revert, but *please* keep an eye on this in the future.

DirectXMan12:
@kubernetes/heapster-maintainers PRs should *not* be merged until they have an appropriate commit description and title.

PR #1626, opened 2017-05-04

watercraft:
Issue 1625
I'm testing with InfluxDB V0.9 and an older version of Heapster so the problem may be solved in a different way.
However, I have a fix that works for us to consider.
The problem is that we are not collecting container stats for a pod that doesn't have labels.
This version of Heapster is pushing down a tag called "labels" for the container that is an empty string.
When I remove this tag with the empty string in the InflluxDB sink code, the data points are taken.


watercraft:
Just signed the CLA; can you re-run the robot?

PR #1627, opened 2017-05-04

### LINKED ISSUE #1625: InfluxDB can't sink pod without labels
I'm testing with InfluxDB V0.9 and an older version of Heapster so the problem may be solved in a different way.
However, I have a fix that works for us to consider.
The problem is that we are not collecting container stats for a pod that doesn't have labels.
This version of Heapster is pushing down a tag called "labels" for the container that is an empty string.
When I remove this tag with the empty string in the InflluxDB sink code, the data points are taken.

watercraft:
Fixes #1625

### DIFF
--- a/metrics/sinks/influxdb/influxdb.go
+++ b/metrics/sinks/influxdb/influxdb.go
@@ -82,11 +82,16 @@ func (sink *influxdbSink) ExportData(dataBatch *core.DataBatch) {
 
 			point := influxdb.Point{
 				Measurement: measurementName,
-				Tags:        metricSet.Labels,
+				Tags:        make(map[string]string, len(metricSet.Labels)),
 				Fields: map[string]interface{}{
 					fieldName: value,
 				},
 				Time: dataBatch.Timestamp.UTC(),
+			}
+			for key, value := range metricSet.Labels {
+				if value != "" {
+					point.Tags[key] = value
+				}
 			}
 			dataPoints = append(dataPoints, point)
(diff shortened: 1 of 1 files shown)


metrics/sinks/influxdb/influxdb.go
```diff
@@ -82,12 +82,17 @@ func (sink *influxdbSink) ExportData(dataBatch *core.DataBatch) {
 
 			point := influxdb.Point{
 				Measurement: measurementName,
-				Tags:        metricSet.Labels,
+				Tags:        make(map[string]string),
```
DirectXMan12: small nit: I'd pre-initialize this with the size of metricSet.Labels

DirectXMan12:
@k8s-bot ok to test

watercraft:
DirectXMan12, does my change 4436e46 address your feedback from this morning?
I don't see any text with the requested changes after that changeset.

metrics/sinks/influxdb/influxdb.go
watercraft: Commit 4436e46 on watercraft/heapster is intended to address this feedback.
Is there more I need to do?

watercraft:
Forgive me if I've done this wrong. Following instructions from the web I created a new branch to squash the commits and created yet another pull request from that branch.  Can you approve that one?

DirectXMan12:
This one looks squashed, now, actually.  For future reference, you don't have to open a new PR -- just `git rebase -i` the current PR to squash, and then `git push --force-with-lease`.

PR #1633, opened 2017-05-08

watercraft:
size tag map allocation in influxdb sync


DirectXMan12:
merged the original, closing this one.

PR #1635, opened 2017-05-09
add cluster_name to influxdb sink

### LINKED ISSUE #1632: InfluxDB Sink : Tags support
We have 3+ Kubernetes clusters. I spinoff a heapster pod on each cluster and sending metrics data to single centralized InfluxDB. All looking good and able to see the data. Since I am sending all metrics data from 3 clusters to single InfluxDB database, i am unable to differentiate the data. Is it possible to add tags like cluster=east cluster=east etc at heapster Sink definition? 

i want to query something like

select * from heapster_table where cluster=east or cluster=west etc 

I looked into documentation and couldn't find anyway to add tags at influxDB sink definition.

andyxning:
Fix #1632.

Add `cluster_name` tag to Influxdb for both heapster and eventer.

/cc @kubernetes/heapster-maintainers @kotarusv

### DIFF
--- a/common/influxdb/influxdb.go
+++ b/common/influxdb/influxdb.go
@@ -40,6 +40,7 @@ type InfluxdbConfig struct {
 	WithFields      bool
 	InsecureSsl     bool
 	RetentionPolicy string
+	ClusterName     string
 }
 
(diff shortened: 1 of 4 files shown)


DirectXMan12:
LGTM :+1:

PR #1642, opened 2017-05-12
GCM: check err when parsing config URL

yu8833:
In CreateGCMSink function, check err when parsing raw query string, make sure the ***opts*** is valid.
Like in [NewKafkaClient](https://github.com/kubernetes/heapster/blob/master/common/kafka/kafka.go#L125), [CreateElasticSearchService](https://github.com/kubernetes/heapster/blob/master/common/elasticsearch/elasticsearch.go#L109).

Signed-off-by: yupengzte <yu.peng36@zte.com.cn>


YuPengZTE:
Thanks @DirectXMan12 .

DirectXMan12:
Please do so for the commit message as well. Otherwise LGTM :+1:

PR #1624, opened 2017-05-04
Fix grafana environment variables used in setup bootstrap

adambkaplan:
Resolves #1623

PR #1647, opened 2017-05-16
bump grafana to 4.2.0

caarlos0:
Can you guys also push a new version of the image when this is merged?

The actual version (4.1.2) was never pushed as well...

Thanks!

### DIFF
--- a/grafana/Makefile
+++ b/grafana/Makefile
@@ -19,12 +19,12 @@
 
 all: build
 
-VERSION?=v4.1.2
+VERSION?=v4.2.0
 PREFIX?=gcr.io/google_containers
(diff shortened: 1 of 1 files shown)


caarlos0:
ping @piosz

andyxning:
@caarlos0 This PR can be closed as we have update the grafana version to v4.2.0 in #1624 .

The only problem is that we just merge the PR without actually compile and push the new image to gcr.io registry. This requires Google people to do. @piosz @DirectXMan12 PTAL.

caarlos0:
@andyxning yah, pushing it go gcr.io would be great! :D

Issue #1201, opened 2016-06-23
Heapster should be able to write custom metrics from a kube cluster running outside GCP


activars:
The current `gcm` sink only works with GCP metadata endpoint.

When running heapster in kube cluster, it would be nice that heapster can:
1. Detect if the environment is outside GCP
2. If a credential is provided for Stackdriver/GCM, it can write metrics to GCM after authenticated with GCP private key authentication

note: created an issue referring to the feature request in #1175

markine:
This would be super nice. I've been able to get the Stackdriver monitoring agent running in Kubernetes 1.7.2 on AWS with a service account json private key mounted in /etc/google/auth/. It would be great if we could follow a similar pattern with heapster.

markine:
FYI looks like https://github.com/kubernetes/heapster/pull/1657 got merged.

markine:
I've confirmed that the above patch works by building master privately.

PR #1651, opened 2017-05-19
[WIP] Ability to use GCM in an AWS environment

georgebuckerfield:
We had a use-case internally to run Kubernetes in AWS but export heapster metrics to GCM/Stackdriver (something similar mentioned [here](https://github.com/kubernetes/heapster/issues/1201)). We've got a container running on a development cluster with this forked version and everything looks like it's working as expected. I'm pretty new to Kubernetes though, so any feedback is very welcome!

There are two environment variables that must be provided to the container for it to work:

Variable | Value
---------|------
GCP_PROJECT_ID | Your project ID from GCP
GOOGLE_APPLICATION_CREDENTIALS | Path to the json file containing GCP credentials

The credentials file can be mounted into the container using a volume from a secret.

WIP because so far I've only been able to test this on AWS.


DirectXMan12:
Please separate Godeps updates into their own commit, otherwise it's difficult to review the PR.

georgebuckerfield:
Thanks @DirectXMan12, I was wondering if I should have done that. I think I'll close this PR and open a new, tidier one.

DirectXMan12:
For future reference, no need to close the PR, you can just `git push --force-with-lease`

PR #1656, opened 2017-05-22
Fix renaming of "runtime" container to "docker-daemon"

### LINKED ISSUE #1639: Containers "runtime" are renamed to "docker-daemon"
The reason is that due to legacy issue in Heapster we needed to translate "runtime" to "docker-daemon" for system containers and it seems that by accident we do it for all containers. See more details https://github.com/kubernetes/heapster/blob/master/metrics/sources/summary/summary.go#L351.

loburm:
This PR fix #1639. Previously all containers named "runtime" have been renamed to "docker-daemon" (that's because reverse procedure was done in kubernetes). Now we are renaming to "docker-daemon" only system containers to avoid cases when container which was originally named as "runtime" renamed to "docker-daemon".

### DIFF
--- a/metrics/sources/summary/summary.go
+++ b/metrics/sources/summary/summary.go
@@ -163,8 +163,8 @@ func (this *summaryMetricsSource) decodeNodeStats(metrics map[string]*MetricSet,
 	metrics[NodeKey(node.NodeName)] = nodeMetrics
 
 	for _, container := range node.SystemContainers {
-		key := NodeContainerKey(node.NodeName, this.getContainerName(&container))
-		containerMetrics := this.decodeContainerStats(labels, &container)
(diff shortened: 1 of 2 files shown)


DirectXMan12:
@piosz didn't you just fix this?

DirectXMan12:
Please provide a more detailed description in both the commit body and the PR description.  Please try also to not have too long of a commit message title (otherwise it wraps).

loburm:
Thanks for the comments, I have updated PR description and commit title.
cc @timstclair

piosz:
/lgtm
/approve
Thanks @loburm for the fix. Could you please cherry-pick this to release-1.3?

piosz:
@k8s-bot ok to test

PR #1657, opened 2017-05-22
Authenticate with GCM without GCE metadata

georgebuckerfield:
This adds functionality to authenticate against GCM without requiring the GCE metadata service. It instead uses the`GOOGLE_APPLICATION_CREDENTIALS` environment variable to lookup a credentials file downloaded from GCP. If this variable is not set, the process falls back to using the GCE metadata. 

This change allows users running clusters in AWS/on-prem to store heapster metrics in GCM/Stackdriver.

### DIFF
--- a/common/gce/gce.go
+++ b/common/gce/gce.go
@@ -15,7 +15,10 @@
 package gce
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
@@ -25,6 +28,8 @@ import (
 const (
 	waitForGCEInterval = 5 * time.Second
 	waitForGCETimeout  = 3 * time.Minute
+	gcpCredentialsEnv  = "GOOGLE_APPLICATION_CREDENTIALS"
+	gcpProjectIdEnv    = "GOOGLE_PROJECT_ID"
 )
 
 func EnsureOnGCE() error {
@@ -35,4 +40,68 @@ func EnsureOnGCE() error {
 		}
 	}
 	return fmt.Errorf("not running on GCE")
+}
+
+func GetProjectId() (string, error) {
+	// Try the environment variable first.
+	if projectId, err := getProjectIdFromEnv(); err != nil {
+		glog.V(4).Infof("Unable to get GCP project ID from environment variable: %v", err)
+	} else {
+		return projectId, nil
+	}
+
+	// Try the default credentials file.
+	if projectId, err := getProjectIdFromFile(); err != nil {
+		glog.V(4).Infof("Unable to get GCP project ID from default credentials file: %v", err)
+	} else {
+		return projectId, nil
+	}
+
+	// Finally, fallback on the metadata service.
+	projectId, err := getProjectIdFromMeta()
+	if err != nil {
+		return "", fmt.Errorf("unable to get GCP project ID: %v", err)
+	}
+	return projectId, nil
+}
+
+func getProjectIdFromEnv() (string, error) {
+	projectId, set := os.LookupEnv(gcpProjectIdEnv)
+	if set != true {
+		return "", fmt.Errorf("environment variable %s not found", gcpProjectIdEnv)
+	}
+	return projectId, nil
+}
+
+func getProjectIdFromFile() (string, error) {
+	file, set := os.LookupEnv(gcpCredentialsEnv)
+	if set != true {
+		return "", fmt.Errorf("environment variable %s not found", gcpCredentialsEnv)
+	}
+	conf, err := ioutil.ReadFile(file)
+	if err != nil {
+		return "", err
+	}
+	var gcpConfig struct {
+		ProjectId *string `json:"project_id"`
+	}
+	err = json.Unmarshal(conf, &gcpConfig)
+	if err != nil {
+		return "", err
+	}
+	if gcpConfig.ProjectId == nil {
+		return "", fmt.Errorf("field project_id not found")
+	}
+	return *gcpConfig.ProjectId, nil
+}
+
+func getProjectIdFromMeta() (string, error) {
+	if err := EnsureOnGCE(); err != nil {
+		return "", err
+	}
+	projectId, err := metadata.ProjectID()
+	if err != nil {
+		return "", err
+	}
+	return projectId, nil
 }
--- a/metrics/sinks/gcm/gcm.go
+++ b/metrics/sinks/gcm/gcm.go
@@ -23,7 +23,6 @@ import (
 	gce_util "k8s.io/heapster/common/gce"
 	"k8s.io/heapster/metrics/core"
 
-	gce "cloud.google.com/go/compute/metadata"
 	"github.com/golang/glog"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
@@ -321,21 +320,21 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
 		return nil, fmt.Errorf("invalid metrics parameter: %s", metrics)
 	}
 
-	if err := gce_util.EnsureOnGCE(); err != nil {
-		return nil, err
-	}
-
-	// Detect project ID
-	projectId, err := gce.ProjectID()
+	client, err := google.DefaultClient(oauth2.NoContext, gcm.MonitoringScope)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating oauth2 client: %v", err)
 	}
 
 	// Create Google Cloud Monitoring service.
-	client := oauth2.NewClient(oauth2.NoContext, google.ComputeTokenSource(""))
 	gcmService, err := gcm.New(client)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating GCM service: %v", err)
+	}
+
+	// Get the GCP Project ID.
+	projectId, err := gce_util.GetProjectId()
+	if err != nil {
+		return nil, fmt.Errorf("error getting GCP project ID: %v", err)
 	}
 
 	sink := &gcmSink{


georgebuckerfield:
I signed it!

piosz:
@igorpeshansky from Stackdriver team.

georgebuckerfield:
Is there anything else I need to do to move this along? Or is it unlikely to be merged? We're using a build of my forked version without any issues for a little while now but it would be great to switch back to the upstream build. Thanks!

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +324,39 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
 
-	// Create Google Cloud Monitoring service.
-	client := oauth2.NewClient(oauth2.NoContext, google.ComputeTokenSource(""))
-	gcmService, err := gcm.New(client)
-	if err != nil {
-		return nil, err
+	if os.Getenv(gcpCredentialEnv) != "" {
+		// GCP credentials supplied, so assume we're not in GCE
+		if os.Getenv(gcpProjectEnv) == "" {
+			return nil, fmt.Errorf("no GCP project ID supplied")
```
igorpeshansky: There is another option -- the credentials file actually has the project id embedded in it. If the user didn't supply one, why not try the one from the credentials file, and then the one from the metadata server, in that order?

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +324,39 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
+		if err := gce_util.EnsureOnGCE(); err != nil {
+			return nil, err
+		}
+		// Detect project ID
+		projectId, err = gce.ProjectID()
+		if err != nil {
+			return nil, err
+		}
+		// Create Google Cloud Monitoring service.
+		client := oauth2.NewClient(oauth2.NoContext, google.ComputeTokenSource(""))
```
igorpeshansky: This is actually already done by `google.DefaultClient` when the credentials file isn't present. Looks like a bunch of that code can be reused.

igorpeshansky (changes requested):

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +324,37 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
-	// Create Google Cloud Monitoring service.
-	client := oauth2.NewClient(oauth2.NoContext, google.ComputeTokenSource(""))
 	gcmService, err := gcm.New(client)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating GCM service: %s", err)
+	}
+
+	// Try and get the project ID, GCE metadata first
+	projectId, err := gce.ProjectID()
```
igorpeshansky: Two comments here:
(a) You want the project id from the credentials to override the one from the environment,
(b) You probably still want a way to supply the project id explicitly, in case the credentials aren't owned by the target project but simply have access to it.

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +324,37 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
+	if err != nil {
+		file := os.Getenv(gcpCredentialEnv)
+		if file == "" {
+			return nil, fmt.Errorf("no %s environment variable detected", gcpCredentialEnv)
+		}
+		conf, err := ioutil.ReadFile(file)
+		if err != nil {
+			return nil, fmt.Errorf("error reading default credentials file: %s", err)
+		}
+		err = json.Unmarshal(conf, &gcpConfig)
```
igorpeshansky: Note: older credentials files may not have an explicit project id field, but have it embedded in the email address. Not sure it it's worth handling those cases here.

igorpeshansky (changes requested):

metrics/sinks/gcm/gcm.go
georgebuckerfield: @igorpeshansky thanks for the feedback! I'll re-order where it looks for the project ID and add another option to provide it explicitly via an environment variable. So we'd check in this order:
- environment variable 
- credentials file
- GCE metadata

Does that sound right?

igorpeshansky: Yes, this is the order that makes the most sense from the Stackdriver perspective.

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +325,43 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
-		return nil, err
+		return nil, fmt.Errorf("error creating GCM service: %s", err)
+	}
+
+	// Try and get the project ID from an environment variable first
+	var projectId string
+	projectId = os.Getenv(gcpProjectId)
+	if projectId == "" {
+		// If the variable is not set, move on to the default credentials file
+		file := os.Getenv(gcpCredentialEnv)
```
igorpeshansky: [Optional] The Google OAuth2 libraries also look for this file in the default location `/etc/google/auth/application_default_credentials.json`. @crassirostris and @piosz should decide if it makes sense to match this behavior.

igorpeshansky (changes requested):

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +325,43 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
+			conf, err := ioutil.ReadFile(file)
+			if err != nil {
+				return nil, fmt.Errorf("error reading default credentials file: %s", err)
+			}
+			var gcpConfig struct {
+				ProjectId string `json:"project_id"`
+			}
+			err = json.Unmarshal(conf, &gcpConfig)
+			if err != nil {
+				return nil, fmt.Errorf("error loading default credentials file: %s", err)
```
georgebuckerfield: I think that if the credentials are supplied in PK12 format, then `google.DefaultClient(oauth2.NoContext, gcm.MonitoringScope)` will fail earlier in the process? At least from testing it locally, I get:
```
error getting credentials using GOOGLE_APPLICATION_CREDENTIALS environment variable: invalid character '\u0082' after top-level value
```
if I point `GOOGLE_APPLICATION_CREDENTIALS` at a downloaded p12 file. 
But we could still have it fall back on the metadata server if there was a problem unmarshalling the json file for any other reason. 

I'm also thinking it would make sense to move some of this out into some functions in `common/gce/gce.go` to keep everything a bit tidier (`GetProjectIdFromEnv()`, `GetProjectIdFromFile()`, etc). 

Thanks again for taking the time to review this.

metrics/sinks/gcm/gcm.go
```diff
@@ -321,21 +325,44 @@ func CreateGCMSink(uri *url.URL) (core.DataSink, error) {
...
 
 	// Create Google Cloud Monitoring service.
-	client := oauth2.NewClient(oauth2.NoContext, google.ComputeTokenSource(""))
 	gcmService, err := gcm.New(client)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating GCM service: %s", err)
+	}
+
+	// Try and get the project ID from the environment variable first
```
igorpeshansky: [Optional] Adding a trailing "." would be consistent with the above comments, but not the ones below... I'll let you pick the style you want to follow.

igorpeshansky (changes requested):

metrics/sinks/gcm/gcm.go
igorpeshansky: That's fair -- the P12 format is listed as deprecated on the key download page.

I'll let the Kubernetes team comment on the refactoring, but keep in mind that extracting the project id from credentials is actually not GCE-specific (though it might be GCP-specific).

igorpeshansky (approved):
You'll need @crassirostris and @piosz to actually approve.

crassirostris:
@georgebuckerfield Sorry for the delay. I actually think that your comment about moving this code to `k8s.io/heapster/common/gce` looks very reasonable, then in the target code you would have clean step

```
projectId, err := gce_utils.GetProjectId()
if err != nil {
  return nil, fmt.Errorf(...)
}
```

georgebuckerfield:
No problem, @crassirostris - I'll work on moving that code this weekend, I think I have most of it in a separate branch already. Let me know if there's anything else obvious that needs looking at.

crassirostris:
@georgebuckerfield Make code less nested, but I think that'll be obvious with a separate function :)

metrics/sinks/gcm/gcm.go
```diff
@@ -36,6 +35,8 @@ const (
 	maxNumLabels    = 10
 	// The largest number of timeseries we can write to per request.
 	maxTimeseriesPerRequest = 200
+	gcpCredentialsEnv       = "GOOGLE_APPLICATION_CREDENTIALS"
```
crassirostris: Why do you need those now?

common/gce/gce.go
```diff
@@ -36,3 +41,57 @@ func EnsureOnGCE() error {
 	}
 	return fmt.Errorf("not running on GCE")
 }
+
+func GetProjectId() (string, error) {
+	var projectId string
```
crassirostris: I think you can just remove those two lines

common/gce/gce.go
```diff
@@ -36,3 +41,57 @@ func EnsureOnGCE() error {
...
+}
+
+func getProjectIdFromEnv() string {
+	return os.Getenv(gcpProjectIdEnv)
+}
+
+func getProjectIdFromFile(file string) (string, error) {
+	conf, err := ioutil.ReadFile(file)
+	if err != nil {
+		return "", fmt.Errorf("Error reading default credentials file: %s", err)
```
crassirostris: Errors should not be capitalized. Check the rest of the code for this

crassirostris (changes requested):

metrics/sinks/gcm/gcm.go
georgebuckerfield: My bad, forgot to remove them.

common/gce/gce.go
```diff
@@ -36,3 +41,55 @@ func EnsureOnGCE() error {
 	}
 	return fmt.Errorf("not running on GCE")
 }
+
+func GetProjectId() (string, error) {
+	// Try the environment variable first.
+	if projectId := getProjectIdFromEnv(); projectId != "" {
+		glog.Infof("Using GCP project ID from environment variable: %s", projectId)
```
crassirostris: I think such comments should not be V(0), but rather V(4)

common/gce/gce.go
```diff
@@ -36,3 +41,55 @@ func EnsureOnGCE() error {
...
+	}
+
+	// Try the default credentials file, if the environment variable is set.
+	if file := os.Getenv(gcpCredentialsEnv); file != "" {
+		projectId, err := getProjectIdFromFile(file)
+		if err == nil {
+			glog.Infof("Using GCP project ID from default credentials file: %s", projectId)
+			return projectId, nil
+		} else {
+			glog.Infof("Unable to get project ID from default credentials file: %s", err)
```
crassirostris: Let's be consistent with the rest of the code and use %v for formatting errors

crassirostris (changes requested):
I think the overall readability will benefit from splitting out the main function into 3 high-level ones:

```
if projectId, err := tryGetFromEnvVar(); err != nil {
  return projectId, nil
} else {
  glog.V(4).Infof("... %v", err)
}

if projectId, err := tryGetFromFile(); err != nil {
  return projectId, nil
} else {
  glog.V(4).Infof("... %v", err)
}

projectId, err := tryGetFromMeta()
if err != nil {
  return projectId, nil
}
return "", fmt.Errorf("... %v", err)
```

crassirostris (approved):
I generally prefer %s for strings and %v for errors, but leave it to you :)

LGTM, please squash commits

georgebuckerfield:
No, you're right, consistency with the rest of the code is better. I'll squash the commits when I get home tonight. Thanks for all your help with this!

crassirostris:
@georgebuckerfield Thank you for your contribution!

Whoops, thought it's already being tested
/ok-to-test

DirectXMan12:
Tests pass, has LGTM from sink maintainer.  Merging

PR #1659, opened 2017-05-23
delete pod_namespace && decrease influxdb labels

### LINKED ISSUE #1654: Decrease Influxdb tags
Currently, we have a lot of influxdb tags. IMO, some of them are duplicated. This may cause a problem when the cluster size growing and the number of data stored in Influxdb is growing. 

According to [Influxdb Doc in v1.2](https://docs.influxdata.com/influxdb/v1.2/concepts/schema_and_data_layout/#don-t-have-too-many-series), the more cardinality the series are, the more memory will be used, thus Influxdb will do more and more GC during service. Even Golang is enhancing its GC, we should also enhance our heapster usage to decrease Influxdb memory.

Assumption of Duplication:
* `namespace_id` and `namespace_name`
  * Prefer `namepsace_name`
* `pod_id` and `pod_name`
  * Prefer `pod_name`
* `pod_namespace` has been marked as deprecated. When will this tag be deleted?
* `nodename` and `hostname`
  * Prefer `nodename`

@kubernetes/heapster-maintainers @piosz @DirectXMan12

andyxning:
Fix #1654 

What does this PR do: 
* delete `pod_namespace` label
* remove `pod_id`, `namespace_id` and `hostname`.

/cc @kubernetes/heapster-maintainers @piosz @DirectXMan12

### DIFF
--- a/metrics/core/labels.go
+++ b/metrics/core/labels.go
@@ -36,11 +36,6 @@ var (
 		Key:         "pod_name",
 		Description: "The name of the pod",
 	}
-	// Deprecated label
-	LabelPodNamespace = LabelDescriptor{
-		Key:         "pod_namespace",
-		Description: "The namespace of the pod",
-	}
 	LabelNamespaceName = LabelDescriptor{
 		Key:         "namespace_name",
 		Description: "The name of the namespace",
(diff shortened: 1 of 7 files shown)


andyxning:
@piosz Many thanks. 👏

PR #1662, opened 2017-05-25
remove host_id from influxdb

andyxning:
According to [doc about `host_id`](https://github.com/kubernetes/heapster/blob/master/docs/storage-schema.md#labels), it is 
> Cloud-provider specified or user specified Identifier of a node

Since host_id is retrievable using the node name, host_id is redundant.

/cc @kubernetes/heapster-maintainers @piosz @DirectXMan12

### DIFF
--- a/metrics/sinks/influxdb/influxdb.go
+++ b/metrics/sinks/influxdb/influxdb.go
@@ -39,6 +39,7 @@ var influxdbBlacklistLabels = map[string]struct{}{
 	core.LabelPodNamespaceUID.Key: {},
 	core.LabelPodId.Key:           {},
 	core.LabelHostname.Key:        {},
+	core.LabelHostID.Key:          {},
 }
 
 const (


DirectXMan12:
do we even need `host_id` at all?  Is `nodename` not sufficient?

andyxning:
I am not sure whether both `host_id` and `nodename` are all the same always not matter deployed on bare metal or GCE or AWS. 

`host_id` is the value of `node.Spec.ExternalID`. `nodename` is the value of `node.Nodename`

andyxning:
Friendly ping @piosz @DirectXMan12 :)

andyxning:
@piosz @DirectXMan12 PTAL. :)

DirectXMan12:
sure, but as has been previously suggested, if you know one, you can probably find the other -- they're effectively two parallel axes, so we shouldn't store both and cause extra indices to be created in the TSDB.

andyxning:
@DirectXMan12 Comments addressed. PTAL.

After reading code about heapster and kubernetes. The value for `host_id` is the value of `node.Spec.ExternalID`.  The value for `ExternalID` is specific to vendors. Its value maybe the same as `node.Name`(i.e., `nodename` in heapster) and may not.

So, if we can have an endpoint to query 'ExternalID` from `node.Name`(i.e., `nodename` in heapster), we can remove `host_id` from InfluxDB.

DirectXMan12:
I've updated the PR description a bit to match the latest discussion.  Merging.

andyxning:
@DirectXMan12 Many thanks. 👏

PR #1666, opened 2017-05-30
Hawkular-sink updates to definition cache handling

burmanm:
* disablePreCache parameter set to true will prevent initial fetch of definitions from the metrics server (instead building it from the scratch when storing datapoints)
* Each time datapoint is stored, the tags are compared to see if there's an update (such as labels)
* Increase concurrency of registerLabeledIfNecessary with changes the mutexes (they only prevent concurrent cache modification now)

cc @mwringe

### DIFF
--- a/docs/sink-configuration.md
+++ b/docs/sink-configuration.md
@@ -88,6 +88,7 @@ The following options are available:
 * `batchSize`- How many metrics are sent in each request to Hawkular-Metrics (default is 1000)
 * `concurrencyLimit`- How many concurrent requests are used to send data to the Hawkular-Metrics (default is 5)
 * `labelTagPrefix` - A prefix to be placed in front of each label when stored as a tag for the metric (default is `labels.`)
+* `disablePreCache` - Disable cache initialization by fetching metric definitions from Hawkular-Metrics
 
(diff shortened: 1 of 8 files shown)


DirectXMan12:
@mwringe can you do a review?

docs/sink-configuration.md
```diff
@@ -88,6 +88,7 @@ The following options are available:
 * `batchSize`- How many metrics are sent in each request to Hawkular-Metrics (default is 1000)
 * `concurrencyLimit`- How many concurrent requests are used to send data to the Hawkular-Metrics (default is 5)
 * `labelTagPrefix` - A prefix to be placed in front of each label when stored as a tag for the metric (default is `labels.`)
+* `disablePreCache` - Disable cache initialization by fetching metric definitions from Hawkular-Metrics
```
mwringe: should this be 'disablePreCache' or just 'disableCache'. Do we have multiple types of caches?

burmanm: This does not disable the caching itself, only the preloading from Hawkular-Metrics. For that reason, the `disableCache` would be quite misleading.

burmanm:
And what happened to this..?

DirectXMan12:
Merging, since this is from the sink owner and seems to have a generally positive review otherwise.

PR #1677, opened 2017-06-08
Fix the link of image heapster-grafana-amd64

hoangphuocbk:
In gcr.io/google-containers repository, the image heapster-grafana-amd64 with tag v4.2.0 doesn't exist, there are only v4.0.2.


hoangphuocbk:
I'm sorry because there are one person before me pointing it out.

PR #1671, opened 2017-06-02
fix heapster-grafana-amd64 version

mmaquevice:
from non existing 4.2.0 to 4.0.2

PR #1683, opened 2017-06-12
Fix grafana image version

prat0318:
`v4.2.0` is not available, i think it is a typo for `v4.0.2`.


andyxning:
@prat0318 Please take a look at #1671 .

prat0318:
A dupe of #1671

PR #1684, opened 2017-06-13
kafka sink log improvment

Kokan:
The https://github.com/optiopay/kafka driver used for kafka sink, that has a configuration to provide logs via a Logger interface. This contains useful information, but by default those logs are thrown away.

This modification contains an implementation of that Logger calling glog directly.


Kokan:
I've signed it.

DirectXMan12:
This seems reasonable to me. Assigning to the Kafka sink maintainer for a second review.

DirectXMan12:
LGTM since the maintainer isn't responding.  Merging.

PR #1692, opened 2017-06-21
Bumped Kubernetes version to v1.7.0-beta.2 in integration tests

piosz:

PR #1691, opened 2017-06-21
Bumped go version to 1.8

piosz:

PR #1693, opened 2017-06-21
Bumped version to v1.4.0-beta.0

piosz:


piosz:
To be merged after #1692 and #1691.

PR #1700, opened 2017-06-26
Update influxdb and grafana to newer versions

### LINKED ISSUE #1698: Unable to build grafana for other platforms
After #1624 is merged I'm not able to build grafana for other platforms than `amd64`.

See https://github.com/kubernetes/heapster/pull/1624/files#r123724923

luxas:
Updates grafana to v4.3.2 and influxdb to v1.3.0

fixes https://github.com/kubernetes/heapster/issues/1698 as a consequence

/cc @piosz @DirectXMan12 @mwielgus

### DIFF
--- a/grafana/Makefile
+++ b/grafana/Makefile
@@ -19,16 +19,15 @@
 
 all: build
 
-VERSION?=v4.2.0
+VERSION?=v4.3.2
+DEB_VERSION?=4.3.2
 PREFIX?=gcr.io/google_containers
 ARCH?=amd64
 TEMP_DIR:=$(shell mktemp -d)
 LDFLAGS=-w -X main.version=$(VERSION) -X main.commit=unknown-dev -X main.timestamp=0 -extldflags '-static'
-DEB_BUILD=4.2.0
 KUBE_CROSS_IMAGE=gcr.io/google_containers/kube-cross:v1.8.3-1
 
-# s390x
-ALL_ARCHITECTURES=amd64 arm arm64 ppc64le
+ALL_ARCHITECTURES=amd64 arm arm64 ppc64le s390x
 ML_PLATFORMS=linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x
 
 # Set default base image dynamically for each arch
@@ -36,9 +35,9 @@ ifeq ($(ARCH),amd64)
 	BASEIMAGE?=busybox
 	CC=gcc
 endif
-ifeq ($(ARCH),armhfrm)
+ifeq ($(ARCH),arm)
 	BASEIMAGE?=armhf/busybox
-	CC=arm-linux-gnueabi-gcc
+	CC=arm-linux-gnueabihf-gcc
 endif
 ifeq ($(ARCH),arm64)
 	BASEIMAGE?=aarch64/busybox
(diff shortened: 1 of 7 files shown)


andyxning:
@luxas I have no idea why we need this for fix #1698. If you need to update grafana and influxdb, can you please split the update commits in another PR?  

We should make this PR suitable to just fix #1698 :)

luxas:
@andyxning Why not update the others too?
Sure, it can be done separately, but they should be updated anyway so why not?

andyxning:
@luxas No special tastes for me. Just to KISS. :)

If we update both, how about we update the description to doc the update. :)

luxas:
Done

influxdb/Makefile
```diff
@@ -69,7 +69,7 @@ push: gcr-login $(addprefix sub-push-,$(ALL_ARCHITECTURES))
 
 sub-push-%:
 	$(MAKE) ARCH=$* PREFIX=$(PREFIX) VERSION=$(VERSION) build
-	docker push $(PREFIX)/heapster-influxdb-$(ARCH):$(VERSION)
+	docker push $(PREFIX)/heapster-influxdb-$*:$(VERSION)
```
andyxning: @luxas Should `(ARCH)` be replaced with `*`?

andyxning:
@piosz When this PR is merged, we should build and push `heapster-grafana-amd64:v4.3.2` and `heapster-influxdb-amd64:v1.3.0` to gcr.io.

influxdb/Makefile
luxas: yes, this was an oversight from earlier.
`$*` resolves to the `%` parameter, which is the arch being built. `$(ARCH)` in this case resolves statically to amd64 every time

andyxning: Good catch! And glad to see that i have learned another info about `make`. 

The link below is just for any user interesting in this to consult: https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html

grafana/Makefile
```diff
@@ -19,26 +19,25 @@
...
 # Set default base image dynamically for each arch
 ifeq ($(ARCH),amd64)
 	BASEIMAGE?=busybox
 	CC=gcc
 endif
-ifeq ($(ARCH),armhfrm)
+ifeq ($(ARCH),arm)
 	BASEIMAGE?=armhf/busybox
-	CC=arm-linux-gnueabi-gcc
+	CC=arm-linux-gnueabihf-gcc
```
andyxning: I am not familiar with `armhf`, so i have no idea about this change. Others seem good to me.

luxas: Basically the newer kube-cross image has only the armv7 variant (armhf) of gcc.
Earlier versions of kube-cross came bundled with the armv5 (armel) variant of gcc

andyxning:
@luxas @piosz We should also add new relesse note about this new version in `RELEASES.md`.

grafana/Makefile
andyxning: sgtm

luxas:
ping @piosz @DirectXMan12 @mwielgus Please merge and push new images

luxas:
ping @piosz @DirectXMan12 @mwielgus

andyxning:
@luxas Please resolve the conflicts. :)

miry:
@luxas in the merged PR we have grafana v4.4.1, I think you can use the latest versions:

Grafana https://github.com/grafana/grafana/releases/tag/v4.4.3
influxbd v1.3.3 https://github.com/influxdata/influxdb/releases/tag/v1.3.3

DirectXMan12:
can we close this because of the new PR?

luxas:
Yes, closing

Issue #1697, opened 2017-06-22
Heapster should handle duplicated stats from kubelet gracefully


fgrzadkowski:
Ref https://github.com/kubernetes/kubernetes/issues/47853

piosz:
@loburm could you please take a look? Thanks!

DirectXMan12:
we need to make sure we log this if it happens, among other things.  In the past, we've had too many spots in the collection code which silently ignores errors or missing data.

DirectXMan12:
@piosz: it looks @loburm isn't yet on the list of people that I can assign to, it looks like

piosz:
@DirectXMan12 I asked Filip to add him to heapster-reviewers team.

fgrzadkowski:
@DirectXMan12 @piosz Fixed.

piosz:
@fgrzadkowski thx

loburm:
With #1708 merged, I'm closing this issue.

piosz:
Unfortunately this is not fixed yet. We've seen duplicated containers with non-zero resource usage. Will send a fix soon.

piosz:
#1790 fixed the problem in 1.4. Decreasing the priority - we need to fix this in HEAD as well.

PR #1708, opened 2017-06-30
Ignore stats from terminated container if new was created.

loburm:
It may happen that cadvisor continues to report metrics for container
that was terminated even if a new one had been created. In such cases
heapster randomly selects metric from the terminated or running
container. This commit adds simple heuristic to ignore container that
is older and don't have any cpu usage.

### DIFF
--- a/metrics/sources/summary/summary.go
+++ b/metrics/sources/summary/summary.go
@@ -194,8 +194,28 @@ func (this *summaryMetricsSource) decodePodStats(metrics map[string]*MetricSet, 
 
 	for _, container := range pod.Containers {
 		key := PodContainerKey(ref.Namespace, ref.Name, container.Name)
+		// This check ensures that we are not replacing metrics of running container with metrics of terminated one if
+		// there are two exactly same containers reported by kubelet.
+		if _, exist := metrics[key]; exist {
+			glog.V(8).Infof("Metrics reported from two containers with the same key: %v. Create time of "+
+				"containers are %v and %v. Metrics from the older container are going to be dropped.", key,
+				container.StartTime.Time, metrics[key].CreateTime)
+			if containerIsTerminated(&container, metrics[key].CreateTime) {
+				continue
+			}
+		}
 		metrics[key] = this.decodeContainerStats(podMetrics.Labels, &container, false)
 	}
+}
+
+func containerIsTerminated(container *stats.ContainerStats, otherStartTime time.Time) bool {
+	if container.StartTime.Time.Before(otherStartTime) {
+		if *container.CPU.UsageNanoCores == 0 && *container.Memory.UsageBytes == 0 {
+			return true
+		}
+		glog.Warningf("Two identical containers are reported and the older one is not terminated: %v", container)
+	}
+	return false
 }
 
 func (this *summaryMetricsSource) decodeContainerStats(podLabels map[string]string, container *stats.ContainerStats, isSystemContainer bool) *MetricSet {
--- a/metrics/sources/summary/summary_test.go
+++ b/metrics/sources/summary/summary_test.go
@@ -134,6 +134,7 @@ func TestDecodeSummaryMetrics(t *testing.T) {
 			Containers: []stats.ContainerStats{
 				genTestSummaryContainer(cName00, seedPod0Container0),
 				genTestSummaryContainer(cName01, seedPod0Container1),
+				genTestSummaryTerminatedContainer(cName00, seedPod0Container0),
 			},
 		}, {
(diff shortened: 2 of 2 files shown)


piosz:
cc @yujuhong @dashpole

metrics/sources/summary/summary.go
```diff
@@ -194,10 +194,22 @@ func (this *summaryMetricsSource) decodePodStats(metrics map[string]*MetricSet,
 
 	for _, container := range pod.Containers {
 		key := PodContainerKey(ref.Namespace, ref.Name, container.Name)
+		if _, exist := metrics[key]; exist && containerIsTerminated(&container, metrics[key].CreateTime) {
```
piosz: IIUC the logic is as follow:
- if we first encounter an old container, then the new one will just replace it
- if we first encounter the new, this check will ensure that the old will not replace it

It's not that obvious. Please add a comment explaining this.

metrics/sources/summary/summary.go
```diff
@@ -194,10 +194,22 @@ func (this *summaryMetricsSource) decodePodStats(metrics map[string]*MetricSet,
...
+			continue
+		}
 		metrics[key] = this.decodeContainerStats(podMetrics.Labels, &container, false)
 	}
 }
 
+func containerIsTerminated(container *stats.ContainerStats, otherStartTime time.Time) bool {
+	if container.StartTime.Time.Before(otherStartTime) && *container.CPU.UsageNanoCores == 0 {
```
piosz: How about avoiding double checking of the same condition `container.StartTime.Time.Before(otherStartTime)` by implementing this in the following way:
```
if container.StartTime.Time.Before(otherStartTime) {
    if *container.CPU.UsageNanoCores == 0 {
        return true
    }
    glog.Warningf(...)
}
```

piosz (changes requested):

metrics/sources/summary/summary.go
loburm: Comment was added.

metrics/sources/summary/summary.go
loburm: Thanks, like that it's much better.

loburm:
All comments have been resolved.

metrics/sources/summary/summary.go
```diff
@@ -194,10 +194,25 @@ func (this *summaryMetricsSource) decodePodStats(metrics map[string]*MetricSet,
 
 	for _, container := range pod.Containers {
 		key := PodContainerKey(ref.Namespace, ref.Name, container.Name)
+		// This check ensures that we are not replacing metrics of running container with metrics of terminated one if
+		// there are two exactly same containers reported by kubelet.
+		if _, exist := metrics[key]; exist && containerIsTerminated(&container, metrics[key].CreateTime) {
+			continue
```
DirectXMan12: please please please always log conditions like this (at high log level, like 8).  Otherwise, it's impossible to debug user deployments without sending a patched Heapster container to the user.

DirectXMan12 (changes requested):
one place were we should have logging, but otherwise LGTM :+1:

metrics/sources/summary/summary.go
loburm: Thanks for the comment. I have tried to include all information in the log that could help to debug issue.

Could you check it once again?

DirectXMan12 (approved):

DirectXMan12:
looks good to me, will merge soon.

loburm:
Thanks a lot!

metrics/sources/summary/summary.go
```diff
@@ -194,10 +194,30 @@ func (this *summaryMetricsSource) decodePodStats(metrics map[string]*MetricSet,
...
+				continue
+			}
+		}
 		metrics[key] = this.decodeContainerStats(podMetrics.Labels, &container, false)
 	}
 }
 
+func containerIsTerminated(container *stats.ContainerStats, otherStartTime time.Time) bool {
+	if container.StartTime.Time.Before(otherStartTime) {
+		if *container.CPU.UsageNanoCores == 0 {
```
yujuhong: CPU usage may not be accurate. It's possible that the process didn't run for the past time period. Could you use memory instead?

loburm: Thanks for the comment, it really makes sense. I think checking of CPU and Memory usage should be good enough heuristic.

I have updated a code and test.

loburm:
@piosz do you have any other comments?

piosz (approved):

loburm:
Fix #1697.

PR #1713, opened 2017-07-07
Wavefront sink: filter out tags with blank values

basilisk487:
Blank tag values are not valid in Wavefront data format and fail the data point entirely, so we need to filter them out before sending


DirectXMan12:
cc @ezeev

DirectXMan12:
ping @ezeev

ezeev:
LGTM @basilisk487 @DirectXMan12

DirectXMan12:
merging as per owner response

PR #1714, opened 2017-07-07
Print the name of the sink if failed

Kokan:
* Kafka did not implemented core.DataSink Name function.
* Factory now prints the name of the sink failed.


DirectXMan12:
LGTM as long as it passes tests.

DirectXMan12:
Merging since tests pass.

Issue #1719, opened 2017-07-11
Monitor PVC disk io/space usage


dnavre:
I wasn't able to find anything for this. FS stats that heapster provides are, as I understand it, node related and in no consistent way track the disk parameters of a PVC.

This, in my opinion, is a crucial component of any production kubernetes setup.

I investigated monitoring tools other than heapster but wasn't able to find anything with this functionality. The only normal solution that I see as of this moment is creating manually configured agent-containers in each of the pods which will have labeling and configuration for all the PVCs. This seems to be an overkill to me. Agents that monitor the disk usage and are installed as a DaemonSet I guess won't be able to provide this functionality cause they will confuse the PVCs, as the same PVC may be mounted as /dev/xvda and the next time as /dev/xvdb.

So, is there a ready solution that I'm missing or is this indeed something that still needs to be implemented?

P.S.
I'm not sure if heapster is the right place to open this issue, maybe it should have been in the main kubernetes project?

DirectXMan12:
Yeah, open in the main Kubernetes repo.  Heapster merely scrapes the metrics provided by Kubelet

dnavre:
I just checked the Kublet API and it is returning that data. Here's an example:
```
{
    "podRef": {
     "name": "mongo-shard-one-0",
     "namespace": "default",
     "uid": "248d8673-6553-11e7-845b-0203d18125f4"
    },
    "startTime": "2017-07-10T10:10:51Z",
    "containers": [
     {
      "name": "mongo",
      "startTime": "2017-07-10T10:12:22Z",
      "cpu": {
       "time": "2017-07-14T19:39:12Z",
       "usageNanoCores": 24500208,
       "usageCoreNanoSeconds": 17524856640381
      },
      "memory": {
       "time": "2017-07-14T19:39:12Z",
       "availableBytes": 669958144,
       "usageBytes": 2916573184,
       "workingSetBytes": 2475769856,
       "rssBytes": 2034851840,
       "pageFaults": 6895170,
       "majorPageFaults": 1683
      },
      "rootfs": {
       "time": "2017-07-14T19:39:12Z",
       "availableBytes": 13413863424,
       "capacityBytes": 19988041728,
       "usedBytes": 330313728,
       "inodesFree": 4950160,
       "inodes": 5242880,
       "inodesUsed": 45
      },
      "logs": {
       "time": "2017-07-14T19:39:12Z",
       "availableBytes": 13413863424,
       "capacityBytes": 19988041728,
[...]

PR #1723, opened 2017-07-14
Modification of the summary endpoint scraping test to check for corre…

dnavre:
Did this to find a bug with Heapster not reporting pod volume usage data even though it's available in the Kublet API. Issue #1719.

The test case actually proved that the suspect code actually works correctly and the issue is probably somewhere else.

### DIFF
--- a/metrics/sources/summary/summary_test.go
+++ b/metrics/sources/summary/summary_test.go
@@ -62,6 +62,7 @@ const (
 	seedPod2           = 5000
 	seedPod2Container0 = 6000
 	seedPod2Container1 = 7000
+	seedPod3Container0 = 9000
 )
 
 const (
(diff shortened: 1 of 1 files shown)


DirectXMan12:
I'm fine with testing the volume scraping logic, but can you adapt this to just work in the main decoding test?

dnavre:
@DirectXMan12 , as requested I moved the test logic to the decoding test case. :)

metrics/sources/summary/summary_test.go
```diff
@@ -109,6 +118,7 @@ func testingSummaryMetricsSource() *summaryMetricsSource {
 }
 
 func TestDecodeSummaryMetrics(t *testing.T) {
+
```
piosz: remove empty line

piosz:
/lgtm
Just one minor comment. Please squash commits before merge.

dnavre:
Squashed commits. Looks like there's nothing else to do, right? )

DirectXMan12:
LGTM.  merging.

PR #1730, opened 2017-07-17
Refresh namespace values at Pods dashboard

cieni:
The current template for the Pods dashboard won't refresh the list of available namespace, leaving only `default` and `kube-system` as options. It took me some time to figure out why, because both _pod_ and _node_ selector (at the Cluster dashboard) were automatically reloading its values.


cieni:
I signed it!

PR #1732, opened 2017-07-18
add service account heapster

davidstack:
kubernetes does not set the heapster serviceaccount as a default serviceaccount,so i think we should add this serviceaccount


davidstack:
I signed it!

DirectXMan12:
Hmm... I though this was already created.

cc @luxas

luxas:
https://github.com/master/heapster/blob/master/deploy/kube-config/influxdb/heapster.yaml

davidstack:
oh,i did not use this file.thanks.i closed this pull request

PR #1747, opened 2017-08-03
Don't log api errors with log level set to zero

jsoriano:
API errors can be quite spammy if some service is requesting pod metrics and the cluster has lots of short-lived pods, as can be when using ScheduledJobs. Even if heapster is started with `--v=0`.

This change makes these messages to require at least a log level of 2 to be logged.


DirectXMan12:
This seems reasonable.  We commonly use these errors to diagnose issues with Heapster, but `V(2)` is sufficiently low that it's probably fine to just run with that when we care about it.

PR #1748, opened 2017-08-04
Storing and/or ignoring user-provided labels

### LINKED ISSUE #807: Push user labels for pods and containers as separate tags in InfluxDB
As of now we serialize all the user defined labels into a single tag in InfluxDB. 
This was done to meet the number of lables limitation in GCM.
We should make this schema storage dependent.

bartebor:
This PR resolves #807.

There are two new options:
- `--store-label` stores label as separate label/tag in addition to concatenated one
- `--ignore-label` allows to prevent label from being used in concatenation

Custom tags allow for more efficient data retrieval from InfluxDB. Ignoring labels reduces redundancy.
Using both options in order allows for seamless transition from old (regexp based) to new (exact match) retrievals.

### DIFF
--- a/metrics/util/util.go
+++ b/metrics/util/util.go
@@ -21,27 +21,57 @@ import (
 	v1listers "k8s.io/client-go/listers/core/v1"
 	kube_api "k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/heapster/metrics/core"
 	"sort"
 	"strings"
 	"time"
 )
 
 var labelSeperator string
+var storedLabels map[string]string
+var ignoredLabels map[string]string
 
-// Concatenates a map of labels into a Seperator-seperated key:value pairs.
-func LabelsToString(labels map[string]string) string {
-	output := make([]string, 0, len(labels))
-	for key, value := range labels {
-		output = append(output, fmt.Sprintf("%s:%s", key, value))
+// Copies labels from input map to output:
+// - all labels, unless found in ignoredLabels, are concatenated into a Seperator-seperated key:value pairs and stored under core.LabelLabels.Key
+// - labels found in storedLabels are additionally stored under key provided
+func CopyLabels(in map[string]string, out map[string]string) {
+	labels := make([]string, 0, len(in))
+
+	for key, value := range in {
+		if mappedKey, exists := storedLabels[key]; exists {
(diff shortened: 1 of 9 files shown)


bartebor:
I signed cncf-cla.

metrics/util/util.go
```diff
@@ -21,29 +21,59 @@ import (
...
 	kube_api "k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/heapster/metrics/core"
 	"sort"
 	"strings"
 	"time"
 )
 
 var labelSeperator string
+var storedLabels map[string]string
```
DirectXMan12: add some godoc here for clarity.

metrics/util/util.go
```diff
@@ -21,29 +21,59 @@ import (
...
+var ignoredLabels map[string]string
 
-// Concatenates a map of labels into a Seperator-seperated key:value pairs.
-func LabelsToString(labels map[string]string) string {
-	output := make([]string, 0, len(labels))
-	for key, value := range labels {
-		output = append(output, fmt.Sprintf("%s:%s", key, value))
+// Copies labels from input map to output:
```
DirectXMan12: godoc should be of the form

`<method name> does a thing`, so here it should look like:

```
// CopyLabels copies the given set of pod labels into a set of metric labels, using the following logic:
// - all labels....
// ...
```

metrics/util/util.go
```diff
@@ -21,29 +21,59 @@ import (
...
-	return strings.Join(output, labelSeperator)
+	sort.Strings(labels)
+	out[core.LabelLabels.Key] = strings.Join(labels, labelSeperator)
 }
 
 func SetLabelSeperator(seperator string) {
 	labelSeperator = seperator
 }
 
+func SetStoredLabels(labels []string) {
```
DirectXMan12: if possible, I'd love to avoid globals here.  Don't worry about it too much, but if you can find a nice way to avoid globals, that would be great.

DirectXMan12:
LGTM, will merge once tests pass

/ok-to-test

bartebor:
Looks like I missed a release, that's what happens when you take a vacation :wink:

bartebor:
@DirectXMan12 @piosz Will this PR make it into next release?

PR #1752, opened 2017-08-08
Added request duration metric to SD sink

piosz:


crassirostris:
@piosz Duration -> latency mb? WDYT?

piosz:
sounds better

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -308,6 +318,7 @@ func (sink *StackdriverSink) sendRequest(req *sd_api.CreateTimeSeriesRequest) {
 	timeseriesSent.
 		WithLabelValues(strconv.Itoa(responseCode)).
 		Add(float64(len(req.TimeSeries)))
+	requestsLatency.Observe(float64(time.Since(startTime)) / float64(time.Millisecond))
```
crassirostris: I strongly feed about `float64(time.Since(startTime)) -> time.Since(startTime).Seconds()`

But not so sure about `float64(time.Millisecond) -> time.Millisecond.Seconds()`, but it's still better IMO

piosz: done

piosz:
I've changed also histogram to summary which is more suitable for this case.

PTAL

crassirostris:
@piosz 

I'm not sure about the name change (request -> requests). See https://github.com/kubernetes/kubernetes/blob/master/pkg/client/metrics/prometheus/prometheus.go#L33 or https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd/metrics/metrics.go#L55

But otherwise LGTM

Also FYI, https://github.com/kubernetes/kubernetes/blob/master/pkg/client/metrics/prometheus/prometheus.go#L33 uses histogram

piosz:
done

crassirostris:
Thanks!

PR #1755, opened 2017-08-09
Introduced min export interval to SD sink

piosz:
This is needed to control how often the metrics are pushed to SD.

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -42,6 +42,8 @@ type StackdriverSink struct {
 	zone              string
 	stackdriverClient *sd_api.Service
 	requestQueue      chan *sd_api.CreateTimeSeriesRequest
+	minInterval       time.Duration
+	lastExportTime    time.Time
 }
 
 type metricMetadata struct {
@@ -165,6 +167,13 @@ func (sink *StackdriverSink) processMetrics(metricValues map[string]core.MetricV
 }
 
 func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
+	// Make sure we don't export metrics too often.
+	if dataBatch.Timestamp.Before(sink.lastExportTime.Add(sink.minInterval)) {
+		glog.V(2).Infof("Skipping batch from %s because there hasn't passed %s from last export time %s", dataBatch.Timestamp, sink.minInterval, sink.lastExportTime)
+		return
+	}
+	sink.lastExportTime = dataBatch.Timestamp
+
 	req := getReq()
 	for _, metricSet := range dataBatch.MetricSets {
 		switch metricSet.Labels["type"] {
@@ -237,6 +246,15 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 		cluster_name = opts["cluster_name"][0]
 	}
 
+	minInterval := time.Nanosecond
+	if len(opts["min_interval_sec"]) >= 1 {
+		if interval, err := strconv.Atoi(opts["min_interval_sec"][0]); err != nil {
+			return nil, fmt.Errorf("Min interval should be an integer, found: %v", opts["min_interval_sec"][0])
+		} else {
+			minInterval = time.Duration(interval) * time.Second
+		}
+	}
+
 	if err := gce_util.EnsureOnGCE(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -165,6 +167,13 @@ func (sink *StackdriverSink) processMetrics(metricValues map[string]core.MetricV
 }
 
 func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
+	// Make sure we don't export metrics too often.
+	if dataBatch.Timestamp.Before(sink.lastExportTime.Add(sink.minInterval)) {
+		glog.Infof("Skipping batch from %s because there hasn't passed %s from last export time %s", dataBatch.Timestamp, sink.minInterval, sink.lastExportTime)
```
crassirostris: %s -> %v?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -237,6 +246,16 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 		cluster_name = opts["cluster_name"][0]
 	}
 
+	minInterval := time.Nanosecond
+	if len(opts["min_interval_sec"]) >= 1 {
+		if interval, err := strconv.Atoi(opts["min_interval_sec"][0]); err != nil {
+			return nil, fmt.Errorf("Min interval should be an integer, found: %v", opts["min_interval_sec"][0])
+		} else {
```
crassirostris: else is unnecessary

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -237,6 +246,16 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 		cluster_name = opts["cluster_name"][0]
 	}
 
+	minInterval := time.Nanosecond
```
crassirostris: `var minInterval time.Duration`?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -237,6 +246,16 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
...
+	if len(opts["min_interval_sec"]) >= 1 {
+		if interval, err := strconv.Atoi(opts["min_interval_sec"][0]); err != nil {
+			return nil, fmt.Errorf("Min interval should be an integer, found: %v", opts["min_interval_sec"][0])
+		} else {
+			minInterval = time.Duration(interval) * time.Second
+		}
+
```
crassirostris: Extra empty line

metrics/sinks/stackdriver/stackdriver.go
crassirostris: Should it be V(0)? V(2) maybe or even V(4)?

crassirostris (changes requested):

metrics/sinks/stackdriver/stackdriver.go
crassirostris: That doesn't effectively change the behavior, so leave it to you and to the consistency

piosz: Time, Duration has string method defined. It's well formated:
```
I0809 09:21:05.043400       1 stackdriver.go:175] Skipping batch from 2017-08-09 09:21:00 +0000 UTC because there hasn't passed 1m40s from last export time 2017-08-09 09:20:00 +0000 UTC
```

piosz: V(2) seems to be good.

crassirostris: I know, it's the same with errors, I just prefer using %s when it's string and %v otherwise. Don't know how it works in heapster tho, that's why I leave it up to you

metrics/sinks/stackdriver/stackdriver.go
piosz: I prefer to have explicit default.

metrics/sinks/stackdriver/stackdriver.go
piosz: It is - interval is visible only within the scope of `if`

metrics/sinks/stackdriver/stackdriver.go
piosz: removed

metrics/sinks/stackdriver/stackdriver.go
crassirostris: Ack

piosz:
PTAL

metrics/sinks/stackdriver/stackdriver.go
crassirostris: This is an explicit default in go, e.g. linter will ask to change `var x int = 0` to `var x int`. I don't like 1 nanosecond as an initial value, but leave it up to you

crassirostris (approved):
Thanks

PR #1757, opened 2017-08-09
Revisit concurrency model in SD sink

piosz:
The new concurrency model is as follow:
1. After preparing all requests in `ExportData` method, the function `sendRequests` is called asynchronously.
2. SD requests should be handled e2e in <3 seconds. We want to send all requests to SD within the specified period `batchExportTimeoutSec`. This means that one worked can handle at least `n=batchExportTimeoutSec/3` requests. To have some safety margin we create 5 extra workers.
3. We have synchronous channel `requestQueue` to which all requests are written.
4. When all requests are processed or the timeout happen the channel is closed and all workers are stopped.

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -16,6 +16,7 @@ package stackdriver
 
 import (
 	"fmt"
+	"math/rand"
 	"net/url"
 	"reflect"
 	"strconv"
@@ -26,7 +27,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
-	googleapi "google.golang.org/api/googleapi"
+	"google.golang.org/api/googleapi"
 	sd_api "google.golang.org/api/monitoring/v3"
 	gce_util "k8s.io/heapster/common/gce"
 	"k8s.io/heapster/metrics/core"
@@ -34,16 +35,21 @@ import (
 
 const (
 	maxTimeseriesPerRequest = 200
+	// 2 seconds on SD side, 1 extra for networking overhead
+	sdRequestLatencySec           = 3
+	httpResponseCodeUnknown       = -100
+	httpResponseCodeClientTimeout = -1
 )
 
 type StackdriverSink struct {
-	project           string
-	cluster           string
-	zone              string
-	stackdriverClient *sd_api.Service
-	requestQueue      chan *sd_api.CreateTimeSeriesRequest
-	minInterval       time.Duration
-	lastExportTime    time.Time
+	project               string
+	cluster               string
+	zone                  string
+	stackdriverClient     *sd_api.Service
+	minInterval           time.Duration
+	lastExportTime        time.Time
+	batchExportTimeoutSec int
+	initialDelaySec       int
 }
 
 type metricMetadata struct {
@@ -151,7 +157,6 @@ func (sink *StackdriverSink) Name() string {
 }
 
 func (sink *StackdriverSink) Stop() {
-	close(sink.requestQueue)
 }
 
 func (sink *StackdriverSink) processMetrics(metricValues map[string]core.MetricValue,
@@ -174,6 +179,7 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
 	}
 	sink.lastExportTime = dataBatch.Timestamp
 
+	requests := []*sd_api.CreateTimeSeriesRequest{}
 	req := getReq()
 	for key, metricSet := range dataBatch.MetricSets {
 		switch metricSet.Labels["type"] {
@@ -201,7 +207,7 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
 		for _, ts := range timeseries {
 			req.TimeSeries = append(req.TimeSeries, ts)
 			if len(req.TimeSeries) >= maxTimeseriesPerRequest {
-				sink.requestQueue <- req
+				requests = append(requests, req)
 				req = getReq()
 			}
 		}
@@ -213,15 +219,112 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
 				req.TimeSeries = append(req.TimeSeries, point)
 			}
 			if len(req.TimeSeries) >= maxTimeseriesPerRequest {
-				sink.requestQueue <- req
+				requests = append(requests, req)
 				req = getReq()
 			}
 		}
 	}
 
 	if len(req.TimeSeries) > 0 {
-		sink.requestQueue <- req
+		requests = append(requests, req)
 	}
+
+	go sink.sendRequests(requests)
+}
+
+func (sink *StackdriverSink) sendRequests(requests []*sd_api.CreateTimeSeriesRequest) {
+	// Each worker can handle at least batchExportTimeout/sdRequestLatencySec requests within the specified period.
+	// 5 extra workers just in case.
+	workers := 5 + len(requests)/(sink.batchExportTimeoutSec/sdRequestLatencySec)
+	requestQueue := make(chan *sd_api.CreateTimeSeriesRequest)
+	completedQueue := make(chan bool)
+
+	// Launch Go routines responsible for sending requests
+	for i := 0; i < workers; i++ {
+		go sink.requestSender(requestQueue, completedQueue)
+	}
+
+	timeout := time.Duration(sink.batchExportTimeoutSec) * time.Second
+	timeoutSending := time.After(timeout)
+	timeoutCompleted := time.After(timeout)
+
+forloop:
+	for i, r := range requests {
+		select {
+		case requestQueue <- r:
+			// yet another request added to queue
+		case <-timeoutSending:
+			glog.Warningf("Timeout while exporting metrics to Stackdriver. Dropping %d out of %d requests.", len(requests)-i, len(requests))
+			// TODO(piosz): consider cancelling requests in flight
+			// Report dropped requests in metrics.
+			for _, req := range requests[i:] {
+				requestsSent.WithLabelValues(strconv.Itoa(httpResponseCodeClientTimeout)).Inc()
+				timeseriesSent.
+					WithLabelValues(strconv.Itoa(httpResponseCodeClientTimeout)).
+					Add(float64(len(req.TimeSeries)))
+			}
+			break forloop
+		}
+	}
+
+	// Close the channel in order to cancel exporting routines.
+	close(requestQueue)
+
+	workersCompleted := 0
+	for {
+		select {
+		case <-completedQueue:
+			workersCompleted++
+			if workersCompleted == workers {
+				glog.V(4).Infof("All %d workers successfully finished sending requests to SD.", workersCompleted)
+				return
+			}
+		case <-timeoutCompleted:
+			glog.Warningf("Only %d out of %d workers successfully finished sending requests to SD. Some metrics might be lost.", workersCompleted, workers)
+			return
+		}
+	}
+}
+
+func (sink *StackdriverSink) requestSender(reqQueue chan *sd_api.CreateTimeSeriesRequest, completedQueue chan bool) {
+	defer func() {
+		completedQueue <- true
+	}()
+	time.Sleep(time.Duration(rand.Intn(1000*sink.initialDelaySec)) * time.Millisecond)
+	for {
+		select {
+		case req, active := <-reqQueue:
+			if !active {
+				return
+			}
+			sink.sendOneRequest(req)
+		}
+	}
+}
+
+func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest) {
+	startTime := time.Now()
+	empty, err := sink.stackdriverClient.Projects.TimeSeries.Create(fullProjectName(sink.project), req).Do()
+
+	var responseCode int
+	if err != nil {
+		glog.Warningf("Error while sending request to Stackdriver %v", err)
+		switch reflect.Indirect(reflect.ValueOf(err)).Type() {
+		case reflect.Indirect(reflect.ValueOf(&googleapi.Error{})).Type():
+			responseCode = err.(*googleapi.Error).Code
+		default:
+			responseCode = httpResponseCodeUnknown
+		}
+	} else {
+		responseCode = empty.ServerResponse.HTTPStatusCode
+	}
+
+	requestsSent.WithLabelValues(strconv.Itoa(responseCode)).Inc()
+	timeseriesSent.
+		WithLabelValues(strconv.Itoa(responseCode)).
+		Add(float64(len(req.TimeSeries)))
+	requestLatency.Observe(time.Since(startTime).Seconds() / time.Millisecond.Seconds())
+
 }
 
 func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
@@ -233,18 +336,6 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 	}
 
 	opts := uri.Query()
-	var (
-		workers int
-		err     error
-	)
-	if len(opts["workers"]) >= 1 {
-		workers, err = strconv.Atoi(opts["workers"][0])
-		if err != nil {
-			return nil, fmt.Errorf("Number of workers should be an integer, found: %v", opts["workers"][0])
-		}
-	} else {
-		workers = 1
-	}
 
 	cluster_name := ""
 	if len(opts["cluster_name"]) >= 1 {
@@ -257,6 +348,21 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 			return nil, fmt.Errorf("Min interval should be an integer, found: %v", opts["min_interval_sec"][0])
 		} else {
 			minInterval = time.Duration(interval) * time.Second
+		}
+	}
+
+	batchExportTimeoutSec := 60
+	var err error
+	if len(opts["batch_export_timeout_sec"]) >= 1 {
+		if batchExportTimeoutSec, err = strconv.Atoi(opts["batch_export_timeout_sec"][0]); err != nil {
+			return nil, fmt.Errorf("Batch export timeout should be an integer, found: %v", opts["batch_export_timeout_sec"][0])
+		}
+	}
+
+	initialDelaySec := sdRequestLatencySec
+	if len(opts["initial_delay_sec"]) >= 1 {
+		if initialDelaySec, err = strconv.Atoi(opts["initial_delay_sec"][0]); err != nil {
+			return nil, fmt.Errorf("Initial delay should be an integer, found: %v", opts["initial_delay_sec"][0])
 		}
 	}
 
@@ -283,15 +389,14 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 		return nil, err
 	}
 
-	requestQueue := make(chan *sd_api.CreateTimeSeriesRequest)
-
 	sink := &StackdriverSink{
-		project:           projectId,
-		cluster:           cluster_name,
-		zone:              zone,
-		stackdriverClient: stackdriverClient,
-		requestQueue:      requestQueue,
-		minInterval:       minInterval,
+		project:               projectId,
+		cluster:               cluster_name,
+		zone:                  zone,
+		stackdriverClient:     stackdriverClient,
+		minInterval:           minInterval,
+		batchExportTimeoutSec: batchExportTimeoutSec,
+		initialDelaySec:       initialDelaySec,
 	}
 
 	// Register sink metrics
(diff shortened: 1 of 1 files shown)


piosz:
cc @crassirostris @loburm

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -30,20 +30,23 @@ import (
 	sd_api "google.golang.org/api/monitoring/v3"
 	gce_util "k8s.io/heapster/common/gce"
 	"k8s.io/heapster/metrics/core"
+	"net/http"
```
crassirostris: Move to the top

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -30,20 +30,23 @@ import (
...
-	requestQueue      chan *sd_api.CreateTimeSeriesRequest
-	minInterval       time.Duration
-	lastExportTime    time.Time
+	project               string
+	cluster               string
+	zone                  string
+	stackdriverClient     *sd_api.Service
+	minInterval           time.Duration
+	lastExportTime        time.Time
+	batchExportTimeoutSec int
```
crassirostris: Why not time.Duration?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -208,15 +211,90 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
 
 	if len(req.TimeSeries) > 0 {
-		sink.requestQueue <- req
+		requests = append(requests, req)
 	}
+
+	go sink.sendRequests(requests)
+}
+
+func (sink *StackdriverSink) sendRequests(requests []*sd_api.CreateTimeSeriesRequest) {
```
crassirostris: It will not timeout iff `len(requests) - workers` are processed within `batchExportTimeoutSec`, not `len(requests)`

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -208,15 +211,90 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+				timeseriesSent.
+					WithLabelValues(strconv.Itoa(http.StatusTooManyRequests)).
+					Add(float64(len(requests[j].TimeSeries)))
+			}
+			break forloop
+		}
+	}
+
+	// Close the channel in order to cancel exporting routines.
+	close(requestQueue)
```
crassirostris: Defer this right after channel creation mb?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -208,15 +211,90 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+		requests = append(requests, req)
 	}
+
+	go sink.sendRequests(requests)
+}
+
+func (sink *StackdriverSink) sendRequests(requests []*sd_api.CreateTimeSeriesRequest) {
+	// Each worker can handle at least batchExportTimeout/sdRequestLatencySec requests within the specified period.
+	// 5 extra workers just in case.
+	workers := 5 + len(requests)/(sink.batchExportTimeoutSec/sdRequestLatencySec)
```
crassirostris: Why not send them all asynchronously? I'm not saying we should do that, but srsly, what's the reasoning?

crassirostris: Especially having in mind that this is not a true timeout anyway, we can leave some requests hanging while exiting this routine

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -255,6 +321,14 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
 		}
 	}
 
+	batchExportTimeoutSec := 60
+	var err error
+	if len(opts["batch_export_timeout_sec"]) >= 1 {
```
crassirostris: Why not have a duration in the parameters?

crassirostris (changes requested):

metrics/sinks/stackdriver/stackdriver.go
mwielgus: Because the whole metric pass assumes that it is synchronous. We don't want to start another metric scraping if there are too many pending sink sync operations. Heapster, especially in large  clusters, has memory limits set with assumption that there are at most 2 (or 3, don't remember exactly) full scrape results in the memory. Having more may cross the limit and if unlucky with free resources, crash the whole service.

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -208,15 +211,90 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+
+func (sink *StackdriverSink) sendRequests(requests []*sd_api.CreateTimeSeriesRequest) {
+	// Each worker can handle at least batchExportTimeout/sdRequestLatencySec requests within the specified period.
+	// 5 extra workers just in case.
+	workers := 5 + len(requests)/(sink.batchExportTimeoutSec/sdRequestLatencySec)
+	requestQueue := make(chan *sd_api.CreateTimeSeriesRequest)
+
+	// Launch Go routines responsible for sending requests
+	for i := 0; i < workers; i++ {
+		go sink.requestSender(requestQueue)
```
mwielgus: This will create DDOS on StackDriver infrastructure. All Heapsters in GKE clusters scrape  nodes at roughly same time and at the same time they will start pushing the metrics. Please add a random sleep for 5-10 second somewhere.

metrics/sinks/stackdriver/stackdriver.go
crassirostris: I had in mind that all outcoming requests should be canceled when timeout expires

metrics/sinks/stackdriver/stackdriver.go
crassirostris: Maybe instead select a random offest from the range [0, 1m) to start from instead? That'll guarantee the delay between two scrapes while offloading the API

mwielgus: It is a bad idea. We scrape metrics at the slightly different times but we round things up to the closest minute for clarity and for being consistent with the subsequent Heapster runs and different clusters, should the user want do some serious performance debugging.
So it is important that we start scraping nodes at the right, consistent time.

crassirostris: OK, "scraping start offset" -> "sending requests start offset"

My point is, constant offset is better than constantly shrinking and expanding the collection interval, for the purposes of setting up the timeouts

mwielgus: I would set timeouts irrespective of offset and definitely not random. If we have to introduce up to 5 sec offset we should reduce the total timeout by 5 sec.

crassirostris: > I would set timeouts irrespective of offset and definitely not random

Sure!

> If we have to introduce up to 5 sec offset we should reduce the total timeout by 5 sec

OTOH, if the offset is always the same, we can avoid reducing the timeout, which saves us 5 sec to send metrics with higher reliability :)

piosz (commented):
PTAL

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	}
+
+	timeoutTime := time.Now().Add(time.Duration(sink.batchExportTimeoutSec) * time.Second)
+
+forloop:
+	for i := range requests {
```
crassirostris: `for i, r := range requests` and then use `r` instead of `requests[i]` below

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+		now := time.Now()
+		select {
+		case requestQueue <- requests[i]:
+			// yet another request added to queue
+		case <-time.After(timeoutTime.Sub(now)):
+			glog.Warningf("Timeout while exporting metrics to Stackdriver. Dropping %d out of %d requests.", len(requests)-i, len(requests))
```
crassirostris: As usual, should it be a warning, if it's exported in metrics?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+			// TODO(piosz): consider cancelling requests in flight
+			// Report dropped requests in metrics.
+			for j := i; j < len(requests); j++ {
```
crassirostris: `for req := range requests[i:]`

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+				requestsSent.WithLabelValues(strconv.Itoa(http.StatusTooManyRequests)).Inc()
```
crassirostris: I'm little bit hesitant to use the existing http code to describe some client's actions. What if server starts to respond with 429? Maybe switch to using something off the RFC? -1 or -2 for example

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+}
+
+func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest) {
+	startTime := time.Now()
+	empty, err := sink.stackdriverClient.Projects.TimeSeries.Create(fullProjectName(sink.project), req).Do()
+
+	var responseCode int
+	if err != nil {
+		glog.Errorf("Error while sending request to Stackdriver %v", err)
+		switch reflect.Indirect(reflect.ValueOf(err)).Type() {
```
crassirostris: Why not

```
if apiErr, ok := err.(*googleapi.Error); ok {
  responseCode = apiErr.Code
}
```

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+		case reflect.Indirect(reflect.ValueOf(&googleapi.Error{})).Type():
+			responseCode = err.(*googleapi.Error).Code
+		default:
+			responseCode = -1
```
crassirostris: Move -1 to constant?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	// Launch Go routines responsible for sending requests
+	for i := 0; i < workers; i++ {
+		go sink.requestSender(requestQueue)
+	}
+
+	timeoutTime := time.Now().Add(time.Duration(sink.batchExportTimeoutSec) * time.Second)
+
+forloop:
+	for i := range requests {
+		now := time.Now()
```
crassirostris: Why a separate variable?

crassirostris (commented):
Unleashed the nit power!

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	}
+}
+
+func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest) {
+	startTime := time.Now()
+	empty, err := sink.stackdriverClient.Projects.TimeSeries.Create(fullProjectName(sink.project), req).Do()
+
+	var responseCode int
+	if err != nil {
+		glog.Errorf("Error while sending request to Stackdriver %v", err)
```
crassirostris: Definitely NOT an error, not even an warning, since exposed as a metric. I'd say V(2)

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +217,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+
+	timeoutTime := time.Now().Add(time.Duration(sink.batchExportTimeoutSec) * time.Second)
+
+forloop:
+	for i := range requests {
+		now := time.Now()
+		select {
+		case requestQueue <- requests[i]:
+			// yet another request added to queue
+		case <-time.After(timeoutTime.Sub(now)):
```
mwielgus: This is misleading. You can add all requests to the queue but it doesn't mean that the operation completed.

metrics/sinks/stackdriver/stackdriver.go
piosz: done

metrics/sinks/stackdriver/stackdriver.go
piosz: When I debug a problem the first thing I take a look are logs.

metrics/sinks/stackdriver/stackdriver.go
piosz: done

metrics/sinks/stackdriver/stackdriver.go
piosz: done

metrics/sinks/stackdriver/stackdriver.go
piosz: This is just a copy/paste. Let's leave as it is right now

metrics/sinks/stackdriver/stackdriver.go
piosz: done

metrics/sinks/stackdriver/stackdriver.go
piosz: Error msg is not exposed as metric. We need to make it visible - we can't hide everything in logs. Changed to warning.

metrics/sinks/stackdriver/stackdriver.go
piosz: will explain f2f

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +218,92 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	requestQueue := make(chan *sd_api.CreateTimeSeriesRequest)
+	// Close the channel in order to cancel exporting routines.
+	defer close(requestQueue)
+
+	// Launch Go routines responsible for sending requests
+	for i := 0; i < workers; i++ {
+		go sink.requestSender(requestQueue)
+	}
+
+	timeoutTime := time.Now().Add(time.Duration(sink.batchExportTimeoutSec) * time.Second)
```
crassirostris:
```
timeout := time.Duration(sink.batchExportTimeoutSec) * time.Second
t := time.NewTimer(timeout)
defer t.Stop()
```

and then 

```
select {
case rq <- r:
  ...
case <- t.C:
  ...
}
```

crassirostris: Or just `t := time.After(timeout)`

piosz: done

metrics/sinks/stackdriver/stackdriver.go
piosz: reimplemented to timeout on all requests

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,17 +218,102 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+					WithLabelValues(strconv.Itoa(httpResponseCodeClientTimeout)).
+					Add(float64(len(req.TimeSeries)))
+			}
+			break forloop
+		}
+	}
+}
+
+func (sink *StackdriverSink) requestSender(queue chan *sd_api.CreateTimeSeriesRequest, timeout <-chan time.Time) {
+	time.Sleep(time.Duration(rand.Intn(1000*sdRequestLatencySec)) * time.Millisecond)
```
mwielgus: 3 second random latency might be too small to prevent ddos in the future. Please increase the value to 5-10 and make it a flag.

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,17 +218,102 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+			resp := make(chan int)
+			sink.sendOneRequest(req, resp)
+			responseCode := httpResponseCodeClientTimeout
+
+			select {
+			case responseCode = <-resp:
+				// Request sent successfully
+				requestLatency.Observe(time.Since(startTime).Seconds() / time.Millisecond.Seconds())
+			case <-timeout:
+				glog.V(2).Infof("Request to SD not completed on time.")
```
mwielgus: Info?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,17 +218,102 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+		select {
+		case req, active := <-queue:
+			if !active {
+				return
+			}
+
+			startTime := time.Now()
+			resp := make(chan int)
+			sink.sendOneRequest(req, resp)
+			responseCode := httpResponseCodeClientTimeout
```
mwielgus: make it uninitialized and set to timeout in timeout case

crassirostris:
@mwielgus AFAIU, the latest change was addressing you comment

> This is misleading. You can add all requests to the queue but it doesn't mean that the operation completed.
  
But I think now it's even more misleading. Now the request in-flight will be marked as timed out and the response code will be just dropped. It makes the statistics less accurate without introducing any benefits: the response will still be parsed, so it doesn't reduce the memory/cpu footprint.

By the way, to your original comment -- you cannot dump all requests to the queue, b/c it's non-blocking and the fact that an item went this way means implies a guarantee that it will be sent and the response will be recorded. IMO, without a hard limit, what was before is a better thing to do: sent requests will be completed and their response codes will be recorded, not sent requests will be marked as timed out. Maybe instead of the last change, to make things less confusing, just rename some stuff keeping the logic intact?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,17 +218,102 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	for i := 0; i < workers; i++ {
+		go sink.requestSender(requestQueue, time.After(time.Duration(sink.batchExportTimeoutSec)*time.Second))
+	}
+
+	timeout := time.After(time.Duration(sink.batchExportTimeoutSec) * time.Second)
+
+forloop:
+	for i, r := range requests {
+		select {
+		case requestQueue <- r:
```
mwielgus: This piece hasn't improve :/

mwielgus:
@crassirostris 
From the concurrency perspective the new version is slightly better but i agree that it doesn't solve the problem completely. My main concern is that we still mix adding things for execution with timeouts. And because of that we still have convoluted logic. The main improvement that was made was logging the timeouted requests in case of small clusters.

I don't understand the second part of your message

metrics/sinks/stackdriver/stackdriver.go
piosz: obsolete

metrics/sinks/stackdriver/stackdriver.go
piosz: obsolete

metrics/sinks/stackdriver/stackdriver.go
piosz: done

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +219,110 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	for {
+		select {
+		case <-completedQueue:
+			workersCompleted++
+			if workersCompleted == workers {
+				glog.V(4).Infof("All %d workers successfully finished sending requests to SD.", workersCompleted)
+				return
+			}
+		case <-timeoutCompleted:
+			glog.Infof("Only %d out of %d workers successfully finished sending requests to SD. Some metrics might be lost.", workersCompleted, workers)
```
crassirostris: V(0) -> V(2) or Info -> Warning

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -213,15 +219,110 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	// Close the channel in order to cancel exporting routines.
+	defer close(requestQueue)
+
+	// Launch Go routines responsible for sending requests
+	for i := 0; i < workers; i++ {
+		go sink.requestSender(requestQueue, completedQueue)
+	}
+
+	timeoutSending := time.After(time.Duration(sink.batchExportTimeoutSec) * time.Second)
+	timeoutCompleted := time.After(time.Duration(sink.batchExportTimeoutSec) * time.Second)
```
crassirostris: Small nit: extract variable

crassirostris (commented):
SG

crassirostris (approved):

metrics/sinks/stackdriver/stackdriver.go
piosz: done

metrics/sinks/stackdriver/stackdriver.go
piosz: done

piosz:
Comments applied and squashed commits - will merge on green.

PR #1767, opened 2017-08-11
[Stackdriver] Fixed problem with zero pod creation time

piosz:
Stackdriver sink doesn't accept zero as start time.

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -175,10 +175,15 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
 	sink.lastExportTime = dataBatch.Timestamp
 
 	req := getReq()
-	for _, metricSet := range dataBatch.MetricSets {
+	for key, metricSet := range dataBatch.MetricSets {
 		switch metricSet.Labels["type"] {
 		case core.MetricSetTypeNode, core.MetricSetTypePod, core.MetricSetTypePodContainer, core.MetricSetTypeSystemContainer:
 		default:
+			continue
+		}
+
+		if metricSet.CreateTime.IsZero() {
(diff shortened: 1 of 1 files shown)


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -175,13 +175,18 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
-	for _, metricSet := range dataBatch.MetricSets {
+	for key, metricSet := range dataBatch.MetricSets {
 		switch metricSet.Labels["type"] {
 		case core.MetricSetTypeNode, core.MetricSetTypePod, core.MetricSetTypePodContainer, core.MetricSetTypeSystemContainer:
 		default:
 			continue
 		}
 
+		if metricSet.CreateTime.IsZero() {
+			glog.Infof("Skipping incorrect metric set %s because create time is zero", key)
```
crassirostris: Should it be V(0)?

crassirostris:
One comment, otherwise lgtm

metrics/sinks/stackdriver/stackdriver.go
piosz: I don't have a strong opinion here, but I think we should rather make visible that some metrics are dropped. This should be also logged rarely - it occurs with probability 0.1-1% once per container (after its start), so it's not spammy thing.

I can change if you feel it's not useful.

crassirostris: If someone sees this in logs, what this person does? If the answer is "nothing", I think this definitely this like should not be V(0)

Ideally, I'd prefer dropping this to V(2) at least and record a metric if it's not possible to figure it out from existing metrics already. This way the losses are visible, but the log is not flooded

piosz: SGTM, changed to V(2)

PR #1770, opened 2017-08-14
[SD sink] Various improvements to SD sink

piosz:
cherry-pick of #1752, #1755, #1767, #1757

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -175,10 +175,15 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
 	sink.lastExportTime = dataBatch.Timestamp
 
 	req := getReq()
-	for _, metricSet := range dataBatch.MetricSets {
+	for key, metricSet := range dataBatch.MetricSets {
 		switch metricSet.Labels["type"] {
 		case core.MetricSetTypeNode, core.MetricSetTypePod, core.MetricSetTypePodContainer, core.MetricSetTypeSystemContainer:
 		default:
+			continue
+		}
+
+		if metricSet.CreateTime.IsZero() {
+			glog.V(2).Infof("Skipping incorrect metric set %s because create time is zero", key)
 			continue
 		}
 
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -134,6 +134,14 @@ var (
 		},
 		[]string{"code"},
 	)
+	requestLatency = prometheus.NewSummary(
+		prometheus.SummaryOpts{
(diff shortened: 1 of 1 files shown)


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -191,17 +219,114 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+
+	// Close the channel in order to cancel exporting routines.
+	close(requestQueue)
+
+	workersCompleted := 0
+	for {
+		select {
+		case <-completedQueue:
+			workersCompleted++
+			if workersCompleted == workers {
```
mwielgus: you can move this if condition to for, will be cleaner

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -191,17 +219,114 @@ func (sink *StackdriverSink) ExportData(dataBatch *core.DataBatch) {
...
+	timeout := time.Duration(sink.batchExportTimeoutSec) * time.Second
+	timeoutSending := time.After(timeout)
+	timeoutCompleted := time.After(timeout)
+
+forloop:
+	for i, r := range requests {
+		select {
+		case requestQueue <- r:
+			// yet another request added to queue
+		case <-timeoutSending:
```
mwielgus: I would remove the timeout here and use buffered channels instead. There is no point in having timeout control in 2 places.

metrics/sinks/stackdriver/stackdriver.go
piosz: will send a follow up PR - this is a pure cherry-pick

metrics/sinks/stackdriver/stackdriver.go
piosz: Without this you won't be able to distinguish when a request was not picked at all vs. the case when it's in progress.

mwielgus: How useful this information is?

metrics/sinks/stackdriver/stackdriver.go
mwielgus: ok.

metrics/sinks/stackdriver/stackdriver.go
piosz: It's for accounting purposes: we want to track number of requests/timeseries successfully delivered to SD. Those which are not picked by any thread are clearly dropped.

mwielgus: Just FYI, switching to buffered channels will require you to change the stop method and introduce an extra check around line 295, so it is not a single line :/. 
Anyway as this is a cherry-pick I will not push for any changes but please consider refactoring it in the future.

PR #1791, opened 2017-08-30
Add gzip compression for kafka sinks

juanluisvaladas:
Currently kafka sends events and metrics uncompressed, which causes huge sizes.


juanluisvaladas:
I signed it!

PR #1796, opened 2017-08-31
Remove the dashboard element from the grafana dashboards

thedebugger:


thedebugger:
FYI, When importing in Grafana v4.3, dashboard used to be blank

andyxning:
@thedebugger IIUC, you mean that grafana dashboard is not compatible with Grafana v4.3?

Can you please post the error scene?

thedebugger:
@andyxning it is compatible but since the committed json has a dashboard element, it used to look like this when you import it -- https://gist.github.com/thedebugger/45054f1a009b14d0ea47b1c980950924

So to grafana it seems the dashboard is empty - no panels/rows comes up. If you remove the dashboard element, then import works. 

Let me know if you need more information

andyxning:
@thedebugger IIUC, you mean that there is an additional `dashboard` element?

andyxning:
@thedebugger Just tested this in our env and it is true. But this is a breaking change, imo. Have you test that after delete `dashboard` element it will still work with older grafana version?

andyxning:
@thedebugger How do you find the solution to fix this problem. i.e., delete dashboard element. This is a breaking change and i think grafana should document it somewhere.

thedebugger:
@andrewsykim my hunch is that something went wrong when we exported the dashboard and committed here? And yes agreed grafana should handle if this is a breaking change. I tried finding the grafana export schema but documentation is very limited. Can you try if you can dig something up?

PR #1801, opened 2017-09-08
fast return when scrape error

andyxning:
When error in scraping a knode metrics, heapster should fail fast after error checking.

Currently, even when scraping error, an successful scrape log will also emitted which is confusing.

/cc @DirectXMan12 @piosz

### DIFF
--- a/metrics/manager/manager.go
+++ b/metrics/manager/manager.go
@@ -125,7 +125,12 @@ func (rm *realManager) housekeep(start, end time.Time) {
 	go func(rm *realManager) {
 		// should always give back the semaphore
 		defer func() { rm.housekeepSemaphoreChan <- struct{}{} }()
-		data := rm.source.ScrapeMetrics(start, end)
+		data, err := rm.source.ScrapeMetrics(start, end)
+
+		if err != nil {
+			glog.Errorf("Error in scraping metrics for %s: %v", rm.source.Name(), err)
+			return
+		}
 
 		for _, p := range rm.processors {
 			newData, err := process(p, data)
@@ -139,7 +144,6 @@ func (rm *realManager) housekeep(start, end time.Time) {
 
 		// Export data to sinks
 		rm.sink.ExportData(data)
-
 	}(rm)
 }
 
--- a/metrics/sources/kubelet/kubelet.go
+++ b/metrics/sources/kubelet/kubelet.go
@@ -233,17 +233,20 @@ metricloop:
 	return metricSetKey, cMetrics
 }
 
-func (this *kubeletMetricsSource) ScrapeMetrics(start, end time.Time) *DataBatch {
+func (this *kubeletMetricsSource) ScrapeMetrics(start, end time.Time) (*DataBatch, error) {
 	containers, err := this.scrapeKubelet(this.kubeletClient, this.host, start, end)
+
 	if err != nil {
-		glog.Errorf("error while getting containers from Kubelet %s: %v", this.host, err)
+		return nil, err
 	}
+
 	glog.V(2).Infof("successfully obtained stats from %s for %v containers", this.host, len(containers))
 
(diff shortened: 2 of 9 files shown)


metrics/sources/kubelet/kubelet.go
```diff
@@ -235,15 +235,19 @@ metricloop:
...
-	if err != nil {
-		glog.Errorf("error while getting containers from Kubelet %s: %v", this.host, err)
-	}
-	glog.V(2).Infof("successfully obtained stats from %s for %v containers", this.host, len(containers))
 
 	result := &DataBatch{
 		Timestamp:  end,
 		MetricSets: map[string]*MetricSet{},
 	}
+
```
DirectXMan12: we shouldn't be returning a result at all if there's an error.  Change the signature of the method to also return an error (you might also have to change the signature of the interface for metric sources).

andyxning:
@DirectXMan12 comments addressed. PTAL.

metrics/manager/manager.go
```diff
@@ -125,21 +125,22 @@ func (rm *realManager) housekeep(start, end time.Time) {
...
-		for _, p := range rm.processors {
-			newData, err := process(p, data)
-			if err == nil {
-				data = newData
-			} else {
-				glog.Errorf("Error in processor: %v", err)
-				return
+		data, err := rm.source.ScrapeMetrics(start, end)
+
+		if err == nil {
```
DirectXMan12: it's generally better to handle the short case in an `if`:

```go
if err != nil {
    glog.Errorf("unable to scrape metrics: %v", err)
    return
}

for _, p := range ...
```

metrics/sources/kubelet/kubelet.go
```diff
@@ -233,25 +233,30 @@ metricloop:
 	return metricSetKey, cMetrics
 }
 
-func (this *kubeletMetricsSource) ScrapeMetrics(start, end time.Time) *DataBatch {
+func (this *kubeletMetricsSource) ScrapeMetrics(start, end time.Time) (*DataBatch, error) {
 	containers, err := this.scrapeKubelet(this.kubeletClient, this.host, start, end)
+
 	if err != nil {
 		glog.Errorf("error while getting containers from Kubelet %s: %v", this.host, err)
```
DirectXMan12: remove this line, we'll log it later on.  Just return an error like `return fmt.Errorf("unable to scrape Kubelet for container metrics: %v", err)`

metrics/sources/manager.go
```diff
@@ -96,21 +96,23 @@ func (this *sourceManager) ScrapeMetrics(start, end time.Time) *DataBatch {
...
-
-			select {
-			case channel <- metrics:
-				// passed the response correctly.
-				return
-			case <-time.After(timeForResponse):
-				glog.Warningf("Failed to send the response back %s", source)
-				return
+			metrics, err := scrape(source, start, end)
+			if err == nil {
```
DirectXMan12: again, `if err != nil`:

```go
if err != nil {
    glog.Errorf("error scraping source: %v")
    continue
```

make sure to log the error, and not just discard it

metrics/sources/summary/summary.go
```diff
@@ -95,12 +95,12 @@ func (this *summaryMetricsSource) ScrapeMetrics(start, end time.Time) *DataBatch
 
 	if err != nil {
 		glog.Errorf("error while getting metrics summary from Kubelet %s(%s:%d): %v", this.node.NodeName, this.node.IP, this.node.Port, err)
```
DirectXMan12: remove this line, and just return a more descriptive error.

andyxning:
@DirectXMan12 Comments addressed. PTAL.

andyxning:
ping @DirectXMan12

Issue #1804, opened 2017-09-10
InfluxDB pod in CrashLoopBackoff - need advice on how to debug


heckj:
I have an otherwise blank and new 1.7.4 Kubernetes cluster that I created with kubeadm on 3 baremetal machines. I was attempting to add Heapster onto the cluster (akin to what can be found in minikube), and followed the instructions at https://github.com/kubernetes/heapster/blob/master/docs/influxdb.md

The pod for influxDB is reporting "Error":

```console
NAME                                    READY     STATUS    RESTARTS   AGE
etcd-silver                             1/1       Running   0          20h
heapster-603813915-jhxbt                1/1       Running   0          31m
kube-apiserver-silver                   1/1       Running   0          20h
kube-controller-manager-silver          1/1       Running   0          20h
kube-dns-2425271678-wqxrm               3/3       Running   0          20h
kube-flannel-ds-02q5k                   2/2       Running   1          20h
kube-flannel-ds-bdqm5                   2/2       Running   0          20h
kube-flannel-ds-n2nbg                   2/2       Running   0          20h
kube-proxy-2w7pg                        1/1       Running   0          20h
kube-proxy-37czn                        1/1       Running   0          20h
kube-proxy-chcd5                        1/1       Running   0          20h
kube-scheduler-silver                   1/1       Running   0          20h
kubernetes-dashboard-3313488171-mpn5b   1/1       Running   0          20h
monitoring-grafana-2175968514-wkqv3     1/1       Running   0          31m
monitoring-influxdb-1957622127-stdd4    0/1       Error     11         31m
```

I used `kubectl --namespace=kube-system describe pod monitoring-influxdb-1957622127-stdd4` to get details, and the container looks like it was pulled successfully, but returned an error code when starting.

Events:
```console
Events:
  FirstSeen	LastSeen	Count	From			SubObjectPath			Type		Reason			Message
  ---------	--------	-----	----			-------------			--------	------			-------
  32m		32m		1	default-scheduler					Normal		Scheduled		Successfully assigned monitoring-influxdb-1957622127-stdd4 to shuttle
  32m		32m		1	kubelet, shuttle					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "influxdb-storage"
  32m		32m		1	kubelet, shuttle					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "default-token-q6vbd"
[...]

PR #1805, opened 2017-09-11
reverting influxdb and grafana images to earlier versions in deployment

heckj:
possible resolution of https://github.com/kubernetes/heapster/issues/1804

Although I don't know if this is the correct or "best" resolution, but the images pushed to the GCR.io at the recent versions have some issues.

### DIFF
--- a/deploy/kube-config/influxdb/grafana.yaml
+++ b/deploy/kube-config/influxdb/grafana.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: grafana
-        image: gcr.io/google_containers/heapster-grafana-amd64:v4.4.3
+        image: gcr.io/google_containers/heapster-grafana-amd64:v4.4.1
         ports:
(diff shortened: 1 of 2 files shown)


DirectXMan12:
I'd prefer we actually figure out what the issue is in the new images, if we can.

andyxning:
@heckj I don't think we have changed the configuration for influxdb image, so maybe this is related with the influxdb itself.

heckj:
I could definitely use some verification, but it appears to be an issue with the binaries that were loaded as "influxd" on the images hosted by google. I definitely agree I'd prefer to resolve the issues with the images and not roll back if possible, but frankly I don't know how to do anything about the images themselves. I'm not even sure who to contact, other than the bug I filed (#1804) and put up this PR.

DirectXMan12:
we'll need one of the Google folks to make any image changes.  I'll poke one of them and see what we can do.

heckj:
closing

PR #1807, opened 2017-09-13
Use RSS instead of Usage when filtering out terminated containers

### LINKED ISSUE #1697: Heapster should handle duplicated stats from kubelet gracefully

piosz:
fix #1697


piosz:
cc @loburm @yujuhong

yujuhong:
lgtm

PR #1753, opened 2017-08-08
Bump cloud image to 1.4.0

007:

PR #1808, opened 2017-09-13
Bumped version to v1.5.0-beta.0

piosz:


007:
@piosz @DirectXMan12 @crassirostris Per https://github.com/kubernetes/heapster/pull/1753#issuecomment-321331616 can you update release process / internal scripts / docs / whatnot so that versions of dependent projects and configs get updated consistently for the 1.5 non-beta release?

PR #1809, opened 2017-09-14
Hawkular, reduce the sink's memory usage in definition caching

burmanm:
Instead of storing the struct of the metric definition in memory, store only a hash of it. In cases with large amount of metrics, this should reduce the memory usage (and GC load). This also removes a need for reflect.DeepEqual, which should speed up as a side-effect the recent check.

Also, pushes the descriptorTag filtering to the HWKMETRICS side, instead of filtering at the client side.

@mwringe

### DIFF
--- a/metrics/sinks/hawkular/client.go
+++ b/metrics/sinks/hawkular/client.go
@@ -15,10 +15,12 @@
 package hawkular
 
 import (
+	"bytes"
 	"fmt"
+	"hash/fnv"
 	"math"
-	"reflect"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


DirectXMan12:
Would like another review on it.
@mwringe, can you take a look?  If not, I'll try and do a deeper review.

DirectXMan12:
is this still relevant @burmanm ?  If so, we can make sure it's still good to go, and merge it.

burmanm:
@DirectXMan12 I have more improvements done for this in my branch (waiting for some QE work) - so don't merge yet.

burmanm:
I'll open a new one from another branch.

PR #1810, opened 2017-09-15
make influxdb bind-address using ip

andyxning:
This PR move Influxdb `bind-address` from `localhost:8088` to `127.0.0.1:8088`. This can help us avoid setting up heapster in a Kubernetes cluster when some dns settings is not correct, like something happened in the [#1804](https://github.com/kubernetes/heapster/issues/1804).

/cc @DirectXMan12

### DIFF
--- a/influxdb/config.toml
+++ b/influxdb/config.toml
@@ -1,5 +1,5 @@
 reporting-disabled = true
-bind-address = "localhost:8088"
+bind-address = "127.0.0.1:8088"
 
 [meta]
   dir = "/data/meta"


DirectXMan12:
I really think it's fine to leave this as localhost, and call any machine that doesn't have localhost as the loopback misconfigured.

[RFC 6761 section 6.3](https://tools.ietf.org/html/rfc6761#section-6.3) states

> Users may assume that IPv4 and IPv6 address
> queries for localhost names will always resolve to the respective
> IP loopback address.

Furthermore, there's a draft in place to further solidify this: https://tools.ietf.org/html/draft-west-let-localhost-be-localhost-06.

Closing.

andyxning:
@DirectXMan12 Thanks for reviewing this. According to the RFC we should think it is ok to use `localhost` in places where `127.0.0.1` is expected. 

However, this needs another DNS query from translating `localhost` to `127.0.0.1`. If a machine can not do this translation correctly just as the situation in #1804 , It is always ok use `127.0.0.1` to work make influxdb work properly.

Actually, In circumstances where IP is ok and will not push more operation burden, I will prefer Ip instead of a DNS name.

PR #1822, opened 2017-09-29
Fix an issue when memory/bytes_used is not reported.

loburm:
Export both evictable and non-evictable metrics.

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -498,7 +498,18 @@ func (sink *StackdriverSink) TranslateMetric(timestamp time.Time, labels map[str
 		return ts
 	case "memory/bytes_used":
 		point := sink.intPoint(timestamp, timestamp, value.IntValue)
-		return createTimeSeries(resourceLabels, memoryBytesUsedMD, point)
+		ts := createTimeSeries(resourceLabels, memoryBytesUsedMD, point)
+		ts.Metric.Labels = map[string]string{
+			"memory_type": "evictable",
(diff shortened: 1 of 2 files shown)


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -498,7 +498,18 @@ func (sink *StackdriverSink) TranslateMetric(timestamp time.Time, labels map[str
 		return ts
 	case "memory/bytes_used":
 		point := sink.intPoint(timestamp, timestamp, value.IntValue)
-		return createTimeSeries(resourceLabels, memoryBytesUsedMD, point)
+		ts := createTimeSeries(resourceLabels, memoryBytesUsedMD, point)
+		ts.Metric.Labels = map[string]string{
+			"memory_type": "evictable",
+		}
+		return ts
+	case core.MetricMemoryWorkingSet.MetricDescriptor.Name:
```
fgrzadkowski: For my own education - where does this name come from?

aleksandra-malinowska:
LGTM provided Stackdriver confirmed this metric will work with label

aleksandra-malinowska (approved):

metrics/sinks/stackdriver/stackdriver.go
loburm: core.MetricMemoryWorkingSet is a metric that we fetch from kubelet. At the same time bytes_used (and minor_page_faults) are created manually in preprocessMemoryMetrics function.

piosz:
@loburm nice fix. Thanks!

PR #1824, opened 2017-09-29
Log Stackdriver request when it fails.

x13n:
More data for debugging.

### DIFF
--- a/metrics/sinks/stackdriver/stackdriver.go
+++ b/metrics/sinks/stackdriver/stackdriver.go
@@ -15,6 +15,7 @@
 package stackdriver
 
 import (
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"net/url"
@@ -302,6 +303,15 @@ func (sink *StackdriverSink) requestSender(reqQueue chan *sd_api.CreateTimeSerie
 	}
 }
 
+func marshalRequestAndLog(printer func([]byte), req *sd_api.CreateTimeSeriesRequest) {
+	reqJson, errJson := json.Marshal(req)
+	if errJson != nil {
+		printer(reqJson)
+	} else {
+		glog.Errorf("Couldn't marshal Stackdriver request %v", errJson)
+	}
+}
+
 func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest) {
 	startTime := time.Now()
 	empty, err := sink.stackdriverClient.Projects.TimeSeries.Create(fullProjectName(sink.project), req).Do()
@@ -309,6 +319,12 @@ func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest)
 	var responseCode int
 	if err != nil {
 		glog.Warningf("Error while sending request to Stackdriver %v", err)
+		// Convert request to json and log it, but only if logging level is equal to 2 or more.
+		if glog.V(2) {
+			marshalRequestAndLog(func(reqJson []byte) {
+				glog.V(2).Infof("The request was: %s", reqJson)
+			}, req)
+		}
 		switch reflect.Indirect(reflect.ValueOf(err)).Type() {
 		case reflect.Indirect(reflect.ValueOf(&googleapi.Error{})).Type():
 			responseCode = err.(*googleapi.Error).Code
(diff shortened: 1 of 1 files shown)


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -309,6 +310,12 @@ func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest)
 	var responseCode int
 	if err != nil {
 		glog.Warningf("Error while sending request to Stackdriver %v", err)
+		reqJson, errJson := json.Marshal(req)
+		if errJson != nil {
+			glog.Warningf("The request was %s", reqJson)
```
piosz: V(2)?

piosz: Also please print all requests as V(10)

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -302,20 +303,39 @@ func (sink *StackdriverSink) requestSender(reqQueue chan *sd_api.CreateTimeSerie
...
 func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest) {
 	startTime := time.Now()
 	empty, err := sink.stackdriverClient.Projects.TimeSeries.Create(fullProjectName(sink.project), req).Do()
 
 	var responseCode int
 	if err != nil {
 		glog.Warningf("Error while sending request to Stackdriver %v", err)
+		if glog.V(2) {
+			marshalRequestAndLog(func(reqJson []byte) {
+				glog.Warningf("The request was: %s", reqJson)
```
piosz: glog.Infof.V(2)

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -302,20 +303,39 @@ func (sink *StackdriverSink) requestSender(reqQueue chan *sd_api.CreateTimeSerie
...
+}
+
 func (sink *StackdriverSink) sendOneRequest(req *sd_api.CreateTimeSeriesRequest) {
 	startTime := time.Now()
 	empty, err := sink.stackdriverClient.Projects.TimeSeries.Create(fullProjectName(sink.project), req).Do()
 
 	var responseCode int
 	if err != nil {
 		glog.Warningf("Error while sending request to Stackdriver %v", err)
+		if glog.V(2) {
```
piosz: How about adding a comment explaining this if?

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -302,20 +303,39 @@ func (sink *StackdriverSink) requestSender(reqQueue chan *sd_api.CreateTimeSerie
...
+			}, req)
+		}
 		switch reflect.Indirect(reflect.ValueOf(err)).Type() {
 		case reflect.Indirect(reflect.ValueOf(&googleapi.Error{})).Type():
 			responseCode = err.(*googleapi.Error).Code
 		default:
 			responseCode = httpResponseCodeUnknown
 		}
 	} else {
+		if glog.V(10) {
```
piosz: ditto

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -302,20 +303,39 @@ func (sink *StackdriverSink) requestSender(reqQueue chan *sd_api.CreateTimeSerie
...
+			marshalRequestAndLog(func(reqJson []byte) {
+				glog.Infof("Stackdriver request sent: %s", reqJson)
```
piosz: glog.Infof.V(10)

piosz (commented):
Looks good, just a few nits

piosz:
cc @loburm

metrics/sinks/stackdriver/stackdriver.go
x13n: There is no such syntax. There is glog.V(2).Infof, however. Quoting from glog doc, those two calls are (almost) equivalent:

	if glog.V(2) { glog.Info("log this") }

	glog.V(2).Info("log this")

The only difference is, when V(2) is false, the expression inside won't get evaluated. I'm using the former, since instead of simply logging a string I'm also lazy-evaluating conversion to json.

metrics/sinks/stackdriver/stackdriver.go
x13n: Added.

metrics/sinks/stackdriver/stackdriver.go
piosz: Sorry, I was not clear enough. When you are reading the code IMO it's easier to understand:
```
if glog.V(2) {
  glog.V(2).Info("log this") }
}
```
than
```
if glog.V(2) {
  glog.Warninf("log this") }
}
```
because in the second case your first impression is that the log severity is `Warning`, but then when you look closer, it turns out the this is in fact `V(2)`.

x13n: Ack, changed.

piosz:
/lgtm
Thanks!

piosz:
@x13n could you please cherry-pick to release-1.4?

PR #1850, opened 2017-10-17
Code refactoring and simplification

atzannes:
Two functions, `GetAllRawContainers` and `GetSummary`, both started by building an URL. For both functions, the step of building that URL was semantically equivalent although their code was slightly different (which I found confusing). 

I factored-out that common functionality into  two functions, `getScheme` and `getUrl`, and now `GetAllRawContainers` and `GetSummary` just call `getUrl`

### DIFF
--- a/metrics/sources/kubelet/kubelet_client.go
+++ b/metrics/sources/kubelet/kubelet_client.go
@@ -130,33 +130,35 @@ type statsRequest struct {
 	Subcontainers bool `json:"subcontainers,omitempty"`
 }
 
+func (self *KubeletClient) getScheme() string {
+	if self.config != nil && self.config.EnableHttps {
+		return "https"
+	} else {
+		return "http"
+	}
+}
+
+func (self *KubeletClient) getUrl(host Host, path string) string {
+	url := url.URL{
+		Scheme: self.getScheme(),
+		Host:   host.String(),
+		Path:   path,
+	}
+
+	return url.String()
+}
+
 // Get stats for all non-Kubernetes containers.
 func (self *KubeletClient) GetAllRawContainers(host Host, start, end time.Time) ([]cadvisor.ContainerInfo, error) {
-	scheme := "http"
-	if self.config != nil && self.config.EnableHttps {
-		scheme = "https"
-	}
+	url := self.getUrl(host, "/stats/container/")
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
/ok-to-test

please squash the fixup down into the original commit.  Otherwise looks good.

DirectXMan12 (changes requested):
please rebase this to take into account the IPv6 changes (namely, we now use `host.String()` to take care of IPv6 compat)

metrics/sources/kubelet/kubelet_client.go
```diff
@@ -130,33 +130,35 @@ type statsRequest struct {
...
-		Scheme: scheme,
-		Host:   host.String(),
-		Path:   "/stats/container/",
+		Scheme: self.getScheme(),
+		Host:   fmt.Sprintf("%s:%d", host.IP, host.Port),
+		Path:   path,
 	}
 
-	return self.getAllContainers(url.String(), start, end)
+	return url.String()
```
atzannes: I'm wondering if it is preferable to return the URL object instead of the stirng here. All uses of this function at this point just use the String, so that's why I opted for returning the string instead of the object, but I'm on the fence.

metrics/sources/kubelet/kubelet_client.go
```diff
@@ -130,33 +130,35 @@ type statsRequest struct {
...
+	} else {
+		return "http"
 	}
+}
 
+func (self *KubeletClient) getUrl(host Host, path string) string {
 	url := url.URL{
-		Scheme: scheme,
+		Scheme: self.getScheme(),
 		Host:   host.String(),
```
atzannes: Switched to using host.String() after rebasing.

atzannes:
@DirectXMan12 rebased and switched to using host.String() as you said. Good catch, thanks!

metrics/sources/kubelet/kubelet_client.go
DirectXMan12: I'd return the URL itself.

atzannes:
Anything else we need to wrap this up?

piosz:
/lgtm
Thanks @atzannes for your contribution.

atzannes:
it's my pleasure

PR #1853, opened 2017-10-18
fix typo metric_resolution argument

TinySong:
fix document of heapster argument metric_resolution


andyxning:
@TinySong Could you please also change `--metric_resolution` to `--metric-resolution`. As heapster command line argument will replace all `_` with `-`.

Issue #1690, opened 2017-06-20
Unable to launch heapster with apiserver


BenjaminITS:
Hey,

I want to launch heapster with apiserver on kubernetes 1.6.1 but I always get the same error :

> I0620 14:15:14.915656       1 heapster.go:72] /heapster --source=kubernetes.summary_api:'' --api-server=true --client-ca-file=/srv/kubernetes/ca.crt --tls-cert-file=/srv/kubernetes/server.cert --tls-private-key-file=/srv/kubernetes/server.key
I0620 14:15:14.915802       1 heapster.go:73] Heapster version v1.3.0
I0620 14:15:14.916299       1 configs.go:61] Using Kubernetes client with master "https://10.3.0.1:443" and version v1
I0620 14:15:14.916376       1 configs.go:62] Using kubelet port 10255
I0620 14:15:15.004367       1 heapster.go:196] Starting with Metric Sink
E0620 14:15:15.011457       1 heapster.go:122] Could not create the API server: missing clientCA file
I0620 14:15:15.107040       1 heapster.go:106] Starting heapster on port 8082

Here it is how I generate my certs :

> openssl genrsa -out ca.key 2048
openssl req -x509 -new -nodes -key ca.key -days 10000 -out ca.crt -subj "/CN=localhost"
openssl genrsa -out server.key 2048
openssl req -new -key server.key -out apiserver.csr -subj "/CN=localhost" -config openssl.conf
openssl x509 -req -in apiserver.csr -CA ca.crt -CAkey cca.key -CAcreateserial -out server.cert -days 3650 -extensions v3_req -extfile openssl.conf

With this openssl.conf :

> [req]
req_extensions = v3_req
distinguished_name = req_distinguished_name
[req_distinguished_name]
[ v3_req ]
basicConstraints = CA:TRUE
subjectAltName = @alt_names
[alt_names]
DNS.1 = heapster.kube-system
DNS.2 = localhost

And after I launch heapster with these yaml files :

> apiVersion: v1
kind: Secret
metadata:
  name: heapster-apiserver-secrets
  namespace: kube-system
  labels:
    k8s-app: heapster
type: Opaque
data:
  ca.crt: LS0t[...]LQo=
  server.cert: LS0t[...]Cg==
  server.key: LS0t[...]tLQo=

> kind: Service
apiVersion: v1
metadata:
  name: heapster
  namespace: kube-system
  labels:
    kubernetes.io/cluster-service: "true"
[...]

PR #1854, opened 2017-10-19
Update the error message for missing the client CA

ldavis-ca:
https://github.com/kubernetes/heapster/issues/1690


DirectXMan12:
code in vendor is included from other projects.  You'll need to change the code there.

Issue #1852, opened 2017-10-17
Heapster need IPv4 to work


valentin2105:
Hi,

I'm trying to use heapster in k8s IPv6 only environment, 
It seem that Heapster try to get the IPv4 IP of the node, but my workers doesn't get any. 

Kubelet is running with `  --node-ip=<IPv6>`

```
E1017 21:26:05.000410       1 kubelet.go:280] Node florida.k8s.domain.com has no valid hostname and/or IP address: florida.k8s.domain.com
E1017 21:26:05.000420       1 kubelet.go:280] Node texas.k8s.domain.com has no valid hostname and/or IP address: texas.k8s.domain.com
E1017 21:26:05.000427       1 kubelet.go:280] Node alabama.k8s.domain.com has no valid hostname and/or IP address: alabama.k8s.domain.com
```

I can't give more information if needed.

DirectXMan12:
hmm... first, seems we always error out if we don't get an IP address (which seems wrong), but you should probably still report a node IP on the node, even with an IP v6 address.  What does `kubectl get node -o yaml` show under the addresses section?

valentin2105:
`kubectl get node -o yaml `

```
- apiVersion: v1
  kind: Node
  metadata:
    annotations:
      node.alpha.kubernetes.io/ttl: "0"
      volumes.kubernetes.io/controller-managed-attach-detach: "true"
    creationTimestamp: 2017-10-16T04:58:50Z
    labels:
      beta.kubernetes.io/arch: amd64
      beta.kubernetes.io/os: linux
      kubernetes.io/hostname: texas.k8s.srv.example.com
    name: texas.k8s.srv.example.com
    namespace: ""
  spec:
    externalID: texas.k8s.srv.example.com
  status:
    addresses:
    - address: 1204:f200:d:f:c8bc:23ff:ffee:f71f
      type: InternalIP
    - address: texas.k8s.srv.example.com
      type: Hostname
    allocatable:
      cpu: "8"
      memory: 8076356Ki
      pods: "110"
    capacity:
      cpu: "8"
      memory: 8178756Ki
      pods: "110"

```

DirectXMan12:
whoops, I missed a line when I was reading over the code last time: https://github.com/kubernetes/heapster/blob/e90c142b41c709bf47403e5756536ac961294f3d/metrics/sources/kubelet/kubelet.go#L323.

[...]

PR #1861, opened 2017-10-29
Handle IPs as net.IP objects in the way to support IPv6

jsoriano:
This PR tries to solve problems seen in #1852, handling addresses indiferently if the are IPv4 or IPv6. For that it uses `net.IP` type and `net.JoinHostPort` function.
It also makes public the function `kubelet.GetHostnameAndIP` used to obtain hostname and ip of a node so this code can be reused.

### DIFF
--- a/metrics/heapster.go
+++ b/metrics/heapster.go
@@ -20,10 +20,12 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


DirectXMan12:
@piosz we'll need to pick this over to metrics-server, if you haven't already

PR #1866, opened 2017-11-07
Set max-row-limit to default value

JHK:
The [Influx Documentation](https://docs.influxdata.com/influxdb/v1.3/administration/config/#max-row-limit-0) states a default value of `0` for the `max-row-limit` parameter. This circumvents an [issue with Grafana](https://github.com/grafana/grafana/issues/7380), where InfluxDB just responds with partial data for large data sets.


JHK:
I signed it!

JHK:
I signed it!

PR #1450, opened 2017-01-09
add disk io metric

andyxning:
Address #690.

This PR will add disk io metrics to heapster. It will add these metrics:
* `disk/io_read_bytes`: Number of bytes read from a disk partition
* `disk/io_write_bytes`: Number of bytes written to a disk partition
* `disk/io_read_bytes_rate`: Number of bytes read from a disk partition per second
*  `disk/io_write_bytes_rate`: Number of bytes written to a disk partition per second

all these metrics will add `resource_id` label with values in `major:minor` format.

FYI：[kernel blkio cgroup](https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt)

PR #1879, opened 2017-11-18
bump cadvisor dep to support disk io device name

andyxning:
This PR will update cAdvisor dep to support device name for disk io metrics. This is related to #1450.

/cc @DirectXMan12 @piosz @loburm @dashpole


andyxning:
@loburm Thanks for reviewing this. Could you help me to merge this since @DirectXMan12 and @piosz are not available for now. :)

PR #1881, opened 2017-11-18
remove original counter metrics for influxdb

### LINKED ISSUE #1880: Metrics simplification for influxDB
For now, counter type metrics will have both counter metrics and the rate metrics. However, the most and only useful metrics are the rate metrics actually. This is fine for small clusters. However, when the clusters and running pods becomes more and more larger, the original counter metrics will only add stress on the metrics sinks.

For example, `cpu/usage` is useless IMO as for `cpu/usage_rate` is more useful.
IMHO, metrics in below tables can be deleted before exported to sinks

| Metric Name | Description |
|------------|-------------|
| cpu/usage | CPU hard limit in millicores. |
| memory/page_faults | Number of page faults. |
| memory/major_page_faults | Number of major page faults. |
| network/rx | Cumulative number of bytes received over the network. |
| network/rx_errors |  Cumulative number of errors while receiving over the network. |
| network/tx | Cumulative number of bytes sent over the network. |
|network/tx_errors | Cumulative number of errors while sending over the network. |

With deleting original counter metrics, we can reduce the stress for metric sinks and without loss any useful metrics. :)

/cc @DirectXMan12 @loburm @piosz

andyxning:
Fix #1880 .

This PR addresses #1880 by adding a new influxdb uri param `disable_counter_metrics` to control whether removing original counter metrics or not described in #1880 for influxdb. This is quite useful when cluster becomes larger. Less stress will be pushed to InfluxDB.

/cc @DirectXMan12 @loburm @piosz

### DIFF
--- a/common/influxdb/influxdb.go
+++ b/common/influxdb/influxdb.go
@@ -32,15 +32,16 @@ type InfluxdbClient interface {
 }
 
 type InfluxdbConfig struct {
-	User            string
-	Password        string
-	Secure          bool
-	Host            string
-	DbName          string
-	WithFields      bool
-	InsecureSsl     bool
-	RetentionPolicy string
-	ClusterName     string
+	User                  string
+	Password              string
+	Secure                bool
+	Host                  string
+	DbName                string
+	WithFields            bool
+	InsecureSsl           bool
+	RetentionPolicy       string
+	ClusterName           string
+	DisableCounterMetrics bool
 }
 
 func NewClient(c InfluxdbConfig) (InfluxdbClient, error) {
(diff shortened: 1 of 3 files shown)


andyxning:
Friendly ping @DirectXMan12 @loburm @piosz PTAL.

andyxning:
Ping @kubernetes/heapster-maintainers

loburm:
Is there any way to warn users about removal of this metrics? I don't want to break people that depend on this metrics without giving them some time to migrate to other metrics.

andyxning:
@loburm I have checked that `cpu/usage` related metrics are not used in grafana dashboards. I agree that this is not a backward compatible change. We should document this in release note and change log. But aside from this, i have not thought any good way to inform users about this change.

loburm:
Just not sure if it's not going to be better UX if we first explicitly deprecate it (in next release notes). And only after some short amount of time (1-2 month) remove it?

DirectXMan12:
the best would probably be to deprecate it after a Kube release (e.g. when 1.10 releases)

andyxning:
@loburm @DirectXMan12 Maybe we can add a warning deprecation release note in the next heapster release and state that after the next 1~2 heaspter release, those counter original metrics will be deleted for InfluxDB.

If this is ok, i'd like to postpone this PR until then.

loburm:
@DirectXMan12 do you propose to remove it in 1.10, or just mark as deprecated and remove in 1.11?

DirectXMan12:
Deprecate in 1.9, remove in 1.10, probably, although we could leave it deprecated for another release if you want and not remove until 1.11.  I will say though, I'm not entirely convinced that this change is a good idea, having read over the original issue (see my comment there).

andyxning:
@loburm @DirectXMan12 I have added a comment to the original issue. I will add a knob to control the sinking of counter metrics.

andyxning:
@DirectXMan12 @loburm I have added `disable_counter_metrics` for influxdb uri param and updated the description for this PR. PTAL.

andyxning:
Ping @DirectXMan12 @piosz

andyxning:
@piosz @DirectXMan12 Could you please take a look at this.

Issue #1872, opened 2017-11-12
Tracking issue for adding GPU metrics to heapster


andyxning:
This is a tracking issue for adding GPU metrics to heapster.
* bump cadvisor to [latest](https://github.com/google/cadvisor/commit/c99f2418fa7bbc6fc54159fa50f0043f882e9e7c)
* add support for GPU metrics for kubelet summary api endpoint
* add support for GPU metrics for normal cAdvisor endpoint

/cc @DirectXMan12 @piosz @loburm @mindprince @vishh

piosz:
It's not fixed yet

PR #1883, opened 2017-11-20
add gpu metrics

### LINKED ISSUE #1449: GPU metrics on kubernetes
The GPU task will be supported on kubernetes soon.
I wonder how to add GPU metrics on heapster, so the users can view the usage of GPU information via Grafana+InfluxDB.

andyxning:
This PR partially fix PR #1872  .

This PR adds support for scrape GPU related metrics when using normal cAdvisor endpoint instead of kubelet summary api. 

* `accelerator/memory_total`
* `accelerator/memory_used`
* `accelerator/duty_cycle`

are added as new accelerator metrics.

/cc @DirectXMan12 @loburm @piosz @mindprince #

### DIFF
--- a/docs/storage-schema.md
+++ b/docs/storage-schema.md
@@ -31,6 +31,9 @@ Heapster exports the following metrics to its backends.
 | memory/cache | Cache memory usage. |
 | memory/rss | RSS memory usage. |
 | memory/working_set | Total working set usage. Working set is the memory being used and not easily dropped by the kernel. |
+| accelerator/memory_total | Memory capacity of an accelerator. |
+| accelerator/memory_used | Memory used of an accelerator. |
+| accelerator/duty_cycle | Duty cycle of an accelerator. |
 | network/rx | Cumulative number of bytes received over the network. |
 | network/rx_errors | Cumulative number of errors while receiving over the network. |
 | network/rx_errors_rate | Number of errors while receiving over the network per second. |
@@ -60,6 +63,9 @@ Heapster tags each metric with the following labels.
 | namespace_id   | UID of the namespace of a Pod                                                 |
 | namespace_name | User-provided name of a Namespace                                             |
 | resource_id    | A unique identifier used to differentiate multiple metrics of the same type. e.x. Fs partitions under filesystem/usage |
+| make  | Make of the accelerator (nvidia, amd, google etc.) |
+| model | Model of the accelerator (tesla-p100, tesla-k80 etc.) |
+| accelerator_id    | ID of the accelerator |
 
 **Note**
   * Label separator can be configured with Heapster `--label-separator`. Comma-seperated label pairs is fine until we use [Bosun](http://bosun.org) as alert system and use `group by labels` to search for labels.
--- a/metrics/core/metrics.go
+++ b/metrics/core/metrics.go
@@ -71,6 +71,9 @@ var LabeledMetrics = []Metric{
 	MetricFilesystemAvailable,
 	MetricFilesystemInodes,
 	MetricFilesystemInodesFree,
+	MetricAcceleratorMemoryTotal,
+	MetricAcceleratorMemoryUsed,
+	MetricAcceleratorDutyCycle,
 }
 
 var NodeAutoscalingMetrics = []Metric{
@@ -601,7 +604,7 @@ var MetricFilesystemUsage = Metric{
 		Units:       UnitsBytes,
 		Labels:      metricLabels,
 	},
-	HasLabeledMetric: func(spec *cadvisor.ContainerSpec) bool {
+	HasLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) bool {
 		return spec.HasFilesystem
 	},
 	GetLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []LabeledMetric {
@@ -632,7 +635,7 @@ var MetricFilesystemLimit = Metric{
 		Units:       UnitsBytes,
 		Labels:      metricLabels,
 	},
-	HasLabeledMetric: func(spec *cadvisor.ContainerSpec) bool {
+	HasLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) bool {
 		return spec.HasFilesystem
 	},
 	GetLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []LabeledMetric {
@@ -663,7 +666,7 @@ var MetricFilesystemAvailable = Metric{
 		Units:       UnitsBytes,
 		Labels:      metricLabels,
 	},
-	HasLabeledMetric: func(spec *cadvisor.ContainerSpec) bool {
+	HasLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) bool {
 		return spec.HasFilesystem
 	},
 	GetLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []LabeledMetric {
@@ -694,7 +697,7 @@ var MetricFilesystemInodes = Metric{
 		Units:       UnitsBytes,
 		Labels:      metricLabels,
 	},
-	HasLabeledMetric: func(spec *cadvisor.ContainerSpec) bool {
+	HasLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) bool {
 		return spec.HasFilesystem
 	},
 	GetLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []LabeledMetric {
@@ -727,7 +730,7 @@ var MetricFilesystemInodesFree = Metric{
 		Units:       UnitsBytes,
 		Labels:      metricLabels,
 	},
-	HasLabeledMetric: func(spec *cadvisor.ContainerSpec) bool {
+	HasLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) bool {
 		return spec.HasFilesystem
 	},
 	GetLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []LabeledMetric {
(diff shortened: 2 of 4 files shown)


mindprince:
Why use the cAdvisor endpoint instead of the summary API?

Also, when we expose these metrics, we would probably expose them as
`accelerator/...`

andyxning:
@mindprince 

> Why use the cAdvisor endpoint instead of the summary API?

For now at least i use the original cAdvisor api to collect metrics. I also think that there are some people use normal cAdvisor endpoint instead of summary endpoint. :)

>when we expose these metrics, we would probably expose them as
accelerator/...

Done. PTAL.

piosz:
LGTM from me. I'll merge it on @mindprince's LGTM.

BTW is anyone going to add support for this in Summary API?

docs/storage-schema.md
```diff
@@ -31,6 +31,10 @@ Heapster exports the following metrics to its backends.
 | memory/cache | Cache memory usage. |
 | memory/rss | RSS memory usage. |
 | memory/working_set | Total working set usage. Working set is the memory being used and not easily dropped by the kernel. |
+| accelerator/memory_capacity | Memory capacity of an accelerator . |
+| accelerator/memory_usage | Memory used of an accelerator . |
+| accelerator/utility | Utility of an accelerator . |
```
mindprince: Use the names that are used in the upstream API.
```
accelerator/memory_total
accelerator/memory_used
accelerator/duty_cycle
```

docs/storage-schema.md
```diff
@@ -31,6 +31,10 @@ Heapster exports the following metrics to its backends.
...
+| accelerator/node_nvidia_capacity | Nvidia capacity of a node . |
```
mindprince: Let's not use `nvidia` in the metric name and instead use labels if possible.

metrics/core/metrics.go
```diff
@@ -71,6 +72,9 @@ var LabeledMetrics = []Metric{
 	MetricFilesystemAvailable,
 	MetricFilesystemInodes,
 	MetricFilesystemInodesFree,
+	MetricAcceleratorMemoryCapacity,
+	MetricAcceleratorMemoryUsed,
+	MetricAcceleratorUtility,
```
mindprince: Instead of Utility use DutyCycle. They are different things.

metrics/core/metrics.go
```diff
@@ -71,6 +72,9 @@ var LabeledMetrics = []Metric{
 	MetricFilesystemAvailable,
 	MetricFilesystemInodes,
 	MetricFilesystemInodesFree,
+	MetricAcceleratorMemoryCapacity,
```
mindprince: MemoryTotal

metrics/processors/node_autoscaling_enricher.go
```diff
@@ -49,6 +49,7 @@ func (this *NodeAutoscalingEnricher) Process(batch *core.DataBatch) (*core.DataB
 			capacityMem, _ := node.Status.Capacity[kube_api.ResourceMemory]
 			allocatableCpu, _ := node.Status.Allocatable[kube_api.ResourceCPU]
 			allocatableMem, _ := node.Status.Allocatable[kube_api.ResourceMemory]
+			capacityNvidia, _ := node.Status.Capacity[kube_api.ResourceNvidiaGPU]
```
mindprince: ResourceNvidiaGPU is a deprecated resource and is going away. The new resources are exposed through device plugins. I would not add this metric at all for now.

metrics/core/metrics.go
```diff
@@ -741,6 +755,111 @@ var MetricFilesystemInodesFree = Metric{
...
+
+		return true
+	},
+	GetLabeledMetric: func(spec *cadvisor.ContainerSpec, stat *cadvisor.ContainerStats) []LabeledMetric {
+		result := make([]LabeledMetric, 0, len(stat.Accelerators))
+		for _, ac := range stat.Accelerators {
+			result = append(result, LabeledMetric{
+				Name: "accelerator/utility",
+				Labels: map[string]string{
+					LabelResourceID.Key: fmt.Sprintf("%s:%s:%s", ac.Make, ac.Model, ac.ID),
```
mindprince: Am I reading this right? This is just one label with key=LabelResourceID.Key and value=concatenation of Make, Model and ID? If so, I don't think we want this.

All these metrics should have three extra labels.
make, model and accelerator_id.

rohitagarwal003 (changes requested):
You will need to add tests.

mindprince:
@piosz I took a quick look and left some comments.

I also have a question if we even want to do this or just want to do through the summary API? Can both co-exist?

metrics/core/metrics.go
andyxning: Will do.

metrics/processors/node_autoscaling_enricher.go
andyxning: > ResourceNvidiaGPU is a deprecated resource and is going away. The new resources are exposed through device plugins.

Does the resources has a predefined or patterned names? @mindprince

andyxning: As for now, It seems that [`alpha.kubernetes.io/nvidia-gpu` is still available](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3840). So, i think it is fine to add this for now. 

WDYT @piosz @mindprince

metrics/core/metrics.go
andyxning: Will do.

metrics/core/metrics.go
andyxning: Will do.

docs/storage-schema.md
andyxning: Will do.

metrics/processors/node_autoscaling_enricher.go
andyxning: I have add a test for `ResourceNvidiaGPU` resource name, so even it is not supported and removed in the future, everything should work fine.

docs/storage-schema.md
andyxning: Will do.

andyxning:
@mindprince @piosz @DirectXMan12 comments addressed. PTAL.

andyxning:
@mindprince 
> I also have a question if we even want to do this or just want to do through the summary API? Can both co-exist?

normal api and summary api should both be supported, IMO. And, they can co-exist.

andyxning:
Pring @mindprince

DirectXMan12:
Hmm... I'd very much just like to say "always use the summary API", and just deprecate the "legacy" stats API usage entirely.  WDYT, @piosz @kubernetes/heapster-maintainers ?

mindprince:
Agreed. And these were added to the summary API in #1891

andyxning:
@DirectXMan12 @mindprince 

Until kubernetes disable cAdvisor integration, we should support `legacy` stats API usage. Just like what we have discussed in #1881 , this deprecation of `legacy` stats API is depends on the kubernetes. And, for this time i agree that we should make an deprecation statement along side kubernetes release on which cAdvisor integration is disabled.

andyxning:
@DirectXMan12 @loburm @mindprince Please take a look at this PR. Thanks.

metrics/core/labels.go
```diff
@@ -64,6 +64,18 @@ var (
 		Key:         "resource_id",
 		Description: "Identifier(s) specific to a metric",
 	}
+	LabelAcceleratorMake = LabelDescriptor{
+		Key:         "accelerator_make",
```
mindprince: Get rid of this diff. It's changing the labels from what I added in the other PR.

metrics/core/labels.go
```diff
@@ -135,6 +135,9 @@ var podLabels = []LabelDescriptor{
 
 var metricLabels = []LabelDescriptor{
 	LabelResourceID,
+	LabelAcceleratorMake,
```
mindprince: Adding these labels here adds these labels to a lot of metrics where they don't make sense at all. You should use `acceleratorLabels` instead.

mindprince: All the diff from this file can go away.

metrics/core/metrics.go
```diff
@@ -755,32 +758,121 @@ var MetricAcceleratorMemoryTotal = Metric{
 	MetricDescriptor: MetricDescriptor{
 		Name:        "accelerator/memory_total",
 		Description: "Total accelerator memory (in bytes)",
-		Labels:      acceleratorLabels,
```
mindprince: Keep this.

metrics/processors/node_autoscaling_enricher.go
```diff
@@ -49,6 +49,19 @@ func (this *NodeAutoscalingEnricher) Process(batch *core.DataBatch) (*core.DataB
 			capacityMem, _ := node.Status.Capacity[kube_api.ResourceMemory]
 			allocatableCpu, _ := node.Status.Allocatable[kube_api.ResourceCPU]
 			allocatableMem, _ := node.Status.Allocatable[kube_api.ResourceMemory]
+			if capacityNvidia, exists := node.Status.Capacity[kube_api.ResourceNvidiaGPU]; exists {
```
mindprince: As I said before, this is deprecated and the resource management WG doesn't want to add more code that depends on this. Please get rid of this metric completely.

rohitagarwal003 (changes requested):
I am not sure what the final decision would be in terms of whether we want to add this at all or not.

But I took another pass and left some comments.

metrics/core/labels.go
andyxning: I think it is a good thing to make all accelerator related labels `accelerator` prefixed. This may be useful for classification.

metrics/core/labels.go
andyxning: Will do.

metrics/core/metrics.go
andyxning: Will do.

metrics/processors/node_autoscaling_enricher.go
andyxning: Will do.

andyxning:
@mindprince All comments are addressed. PTAL.

metrics/core/labels.go
mindprince: These labels only show up for accelerator metrics (because we use `acceleratorLabels` only in accelerator metrics). Because of this, I think it is better to not repeat accelerator in the label name. `accelerator_id` is a special case because `id` is usually used for other things.

Also, this will keep things consistent with Stackdriver and prometheus.

metrics/core/metrics.go
mindprince: Minor nit: if you move this field from last to here, the diff would be shorter.

andyxning: Will do.

andyxning:
@mindprince Comments addressed.

andyxning:
Ping @piosz @DirectXMan12

DirectXMan12:
> Until kubernetes disable cAdvisor integration, we should support legacy stats API usage or otherwise this is also a backward breaking change

Whether or not Kubernetes has cAdvisor integration is independent of whether we support the legacy stats API.  We can certainly choose to deprecate it and/or not add new features.  Ceasing to add new features is *not* a breaking change.

My vote would be to simply not add new features to our legacy API support, and mark it as deprecated.

andyxning:
@DirectXMan12 What is the deprecation schedule for legacy stats api?

Ping @piosz @kubernetes/heapster-maintainers I want to hear more voice on this.

andyxning:
@piosz Could you please take a look at this?

PR #1885, opened 2017-11-21
Fix few bugs in Heapster exporting metrics for new Stackdriver resource model

kawych:
Changes in this PR:
- Corrected names for several metrics
- Corrected label set for volume and node daemon metrics
- Metrics for volume computed from existing filesystem metrics retrieved by summary API


piosz (approved):
LGTM

PR #1888, opened 2017-11-25
add disable_metric_sink command line argument

andyxning:
This PR will add `disable-metric-sink` command line argument to heapster. 

As for now if we deploy heapster in standalone we do not need the history metrics for HPA. So, i think we should add an command option to disable sink metric which is consuming many memory when the cluster is large.

/cc @DirectXMan12 @piosz @loburm

### DIFF
--- a/metrics/heapster.go
+++ b/metrics/heapster.go
@@ -86,7 +86,7 @@ func main() {
 		glog.Fatalf("Failed to get kubernetes address: %v", err)
 	}
 	sourceManager := createSourceManagerOrDie(opt.Sources)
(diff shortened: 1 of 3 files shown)


andyxning:
Ping @kubernetes/heapster-maintainers

loburm:
/lgtm
@piosz could you also take a look?

DirectXMan12:
Hmm... since we're now sometimes deploying Heapster alongside metrics-server, this actually makes sense.  I think I'm in favor.  WDYT @piosz

andyxning:
@loburm @DirectXMan12 Conflicts addressed. PTAL.

Since @loburm and you have agreed on the PR, can we merge this?

andyxning:
Ping @piosz

DirectXMan12 (approved):
LGTM

PR #1820, opened 2017-09-26
Fix formatting

nlebas:

PR #1894, opened 2017-11-29
Implemented PVC stats aggregation. Exposed PVC stats through Heapster API.

verult:
Introducing the PVC resource, along with node, namespace, pod, and container.

Added new API endpoints `/api/v1/model/namespaces/{namespace-name}/pvcs/...`

This allows users to obtain PVC usage stats independent of pods.

### DIFF
--- a/metrics/sources/summary/summary.go
+++ b/metrics/sources/summary/summary.go
@@ -45,8 +45,14 @@ var (
 	)
 )
 
-// Prefix used for the LabelResourceID for volume metrics.
-const VolumeResourcePrefix = "Volume:"
+const (
+	// Prefix used for the LabelResourceID for volume metrics.
+	VolumeResourcePrefix = "Volume:"
+	// Prefix used for the LabelResourceID for pvc metrics.
+	PVCResourcePrefix = "PVC:"
+
+	PodRefSeparator = "|"
+)
 
 func init() {
 	prometheus.MustRegister(summaryRequestLatency)
@@ -128,6 +134,7 @@ func (this *summaryMetricsSource) decodeSummary(summary *stats.Summary) map[stri
 	this.decodeNodeStats(result, labels, &summary.Node)
 	for _, pod := range summary.Pods {
 		this.decodePodStats(result, labels, &pod)
+		this.decodePVCStats(result, labels, &pod)
 	}
 
 	glog.V(9).Infof("End summary decode")
@@ -208,7 +215,7 @@ func (this *summaryMetricsSource) decodePodStats(metrics map[string]*MetricSet, 
 }
 
 func (this *summaryMetricsSource) decodeContainerStats(podLabels map[string]string, container *stats.ContainerStats, isSystemContainer bool) *MetricSet {
-	glog.V(9).Infof("Decoding container stats stats for container %s...", container.Name)
+	glog.V(9).Infof("Decoding container stats for container %s...", container.Name)
 	containerMetrics := &MetricSet{
 		Labels:              this.cloneLabels(podLabels),
 		MetricValues:        map[string]MetricValue{},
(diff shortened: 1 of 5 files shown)


verult:
/test continuous-integration/travis-ci/pr

Looks like the event manager test is flaky. It happened in this PR as well: https://github.com/kubernetes/heapster/pull/1820

verult:
Is it necessary to keep PVC stats as LabeledMetrics? It's convenient, since I can reuse `decodeFsStats`, but AFAIU the labels are used to distinguish between volumes, and that's not necessary here.

deploy/kube-config/influxdb/heapster.yaml
```diff
@@ -24,7 +24,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command:
         - /heapster
-        - --source=kubernetes:https://kubernetes.default
+        - --source=kubernetes.summary_api:''
```
DirectXMan12: please don't do unrelated changes in the same commit.

metrics/api/v1/model_handlers.go
```diff
@@ -60,6 +60,10 @@ type clusterMetricsFetcher interface {
 	availablePodMetrics(request *restful.Request, response *restful.Response)
 	podMetrics(request *restful.Request, response *restful.Response)
 
+	namespacePVCList(request *restful.Request, response *restful.Response)
```
DirectXMan12: I don't think this is necessary.  The model API is more-or-less deprecated (see #1898), and we don't really want to encourage its use by adding new stuff to it.

metrics/sources/summary/summary.go
```diff
@@ -233,6 +240,64 @@ func (this *summaryMetricsSource) decodeContainerStats(podLabels map[string]stri
...
+				CollectionStartTime: pod.StartTime.Time,
+				ScrapeTime:          vol.Time.Time,
+			}
+
+			this.decodeFsStats(pvcMetrics, PVCResourcePrefix+pvc.Name, &vol.FsStats)
+			metrics[pvcMetricName] = pvcMetrics
+
+		} else {
+			labels := metrics[pvcMetricName].Labels
+			labels[LabelPodRefs.Key] = appendPodRef(labels[LabelPodRefs.Key], pod.PodRef.Name)
```
DirectXMan12: nope, don't do this.  Leave it broken up per-pod.  People can aggregate in their sinks if they want it aggregated, but it's difficult-to-impossible to "unaggregate"

DirectXMan12 (changes requested):
Nix the model change -- it's not necessary,
and you shouldn't be aggregating what amounts to a per-pod metric all together.  People can aggregate in their sinks if they want the sum for all pods for a single PVC, but you can't unaggregate.

deploy/kube-config/influxdb/heapster.yaml
verult: Is the kubelet source still used? If not, I'll add this in another PR.

metrics/sources/summary/summary.go
```diff
@@ -234,6 +241,64 @@ func (this *summaryMetricsSource) decodeContainerStats(podLabels map[string]stri
 	return containerMetrics
 }
 
+// decodePVCStats computes PVC metrics using volume stats of each pod.
+// It selects the volume stats of the pod with the latest scrape time.
+// Metric collection time is the earliest pod start time among all pods.
+// PVC stats are only available when the PVC is in use by at least one pod.
```
kawych: Did you investigate a possibility of retrieving PVCs not used by any pod? Otherwise, what value do these metrics add? Heapster already collects (and exports) volume metrics on pod level.

verult: I think it'd be a lot more difficult to obtain filesystem stats of volumes not mounted on any pod. The main value is to allow users to query PVC stats by PVC name, without having to scan through pods themselves. But maybe adding a `pvc_label` to existing volume metrics is sufficient for this purpose?

kawych: I think Heapster can retrieve PVCs from api server, using core v1 lister: https://github.com/kubernetes/client-go/blob/master/listers/core/v1/persistentvolumeclaim.go. Similar examples are present in processors, for example pod_based_enricher lists all pods and retrieves some information for them: https://github.com/kubernetes/heapster/blob/master/metrics/processors/pod_based_enricher.go

I prefer this over adding pvc label to existing volume metrics, because this label would be applicable only to particular type of the volume. This would leave us without the association between PVCs and pods, but do we need it?

kawych: Actually, I missed a part of it (api server don't have most data about the volumes that we need). So this approach LGTM, this comment is resolved.

verult:
@DirectXMan12 one benefit of pre-aggregating PVC stats and treating it as a separate resource is we can now include PVC-specific information, such as storage classes and volume mode (filesystem or block). These shouldn't be included as pod-level volume metric labels because not all volumes are PVCs.

DirectXMan12:
Heapster is now deprecated, so no new features or sinks will be added.  Please see https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md#initial-deprecation-kubernetes-111 for more information.

PR #1898, opened 2017-12-01
Mark the model API as officially deprecated

DirectXMan12:
It's been unofficially deprecated for a while, and we need to let
users know what to do instead.


DirectXMan12:
cc @kubernetes/heapster-maintainers

PR #1906, opened 2017-12-07
Switch Stackdriver sink to grpc-based Stackdriver API

kawych:
This PR together with #1907 lowers Heapster maximum resource usage by 60% CPU and 25% memory (tested over 2 hours on GCE, 100-node cluster with 3k running user pods).

PR #1907, opened 2017-12-07
Use json-iterator in Kubelet client.

kawych:


piosz:
/lgtm
@wojtek-t PTAL

wojtek-t:
This looks fine.

@kawych @piosz - did you manage to run any benchmarks/load tests with this change?

piosz:
Karol has some results.

kawych:
This is one of changes, the other is #1906. This alone doesn't impact performance significantly - the test with 100 nodes and 3k user pods shows no reduced memory usage and CPU usage reduced only slightly (20m vs 18m). This was on standalone instance of heapster.

PR #1914, opened 2017-12-14
Hawkular, reduce the sink's memory usage in definition caching

burmanm:
* Don't store structs in the memory, instead calculate a hash value which is used for compare
* Always compare the current cached value to the one Heapster provides to see instant change in labels
* Reduce the amount of goroutines for most cache lookups
* Expire the cache items if the Heapster has not seen the metric in two scrapes
* Improve startup time by loading definitions from a new endpoint designed for Openshift (disable with disablePreCaching setup)

With large amount of metrics, this should reduce the amount of memory used by the sink (with 20000 pods, Heapster memory usage was reduced from 5.7GB to 3.9GB)

@DirectXMan12 This replaces the previous PR. These changes have been tested against Heapster 1.3.0, so this is a rebase for newer Heapster.

### DIFF
--- a/metrics/sinks/hawkular/client.go
+++ b/metrics/sinks/hawkular/client.go
@@ -16,8 +16,8 @@ package hawkular
 
 import (
 	"fmt"
+	"hash/fnv"
 	"math"
-	"reflect"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


metrics/sinks/hawkular/client.go
```diff
@@ -28,35 +28,99 @@ import (
...
+		if err != nil {
+			return err
+		}
+		err = h.updateDefinitions(mds)
+		if err != nil {
+			return err
+		}
+	}
+
+	glog.V(4).Infof("PreCaching completed, cached %d definitions\n", len(h.expReg))
```
DirectXMan12: might want this to say "Hawkular pre-caching" instead of just "pre-caching" to make things move obvious in the logs

metrics/sinks/hawkular/client.go
```diff
@@ -28,35 +28,99 @@ import (
...
 			}
-			h.reg[p.ID] = p
 		}
+		h.cache(p)
 	}
 
 	return nil
 }
 
+func hash(md *metrics.MetricDefinition) uint64 {
```
DirectXMan12: please make this name something like `hashDefinition` to be a bit more obvious and avoid naming conflicts in the future.

metrics/sinks/hawkular/client.go
```diff
@@ -28,35 +28,99 @@ import (
...
+	return nil
+}
+
+// cache inserts the item to the cache
+func (h *hawkularSink) cache(md *metrics.MetricDefinition) {
+	h.toCache(md.ID, hash(md))
+}
+
+// toCache inserts the item and updates the TTL in the cache to current time
+func (h *hawkularSink) toCache(key string, hash uint64) {
```
DirectXMan12: `toCache` sounds like a conversion function, name should probably be a bit different (`storeInCache` for example?)

metrics/sinks/hawkular/driver.go
```diff
@@ -198,11 +193,14 @@ func NewHawkularSink(u *url.URL) (core.DataSink, error) {
 }
 
 func (h *hawkularSink) init() error {
-	h.reg = make(map[string]*metrics.MetricDefinition)
+	// h.reg = make(map[string]uint64)
```
DirectXMan12: should just remove this commented-out line

metrics/sinks/hawkular/driver_test.go
```diff
@@ -362,14 +376,15 @@ func TestRegister(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Equal(t, 2, len(hSink.models))
-	assert.Equal(t, 1, len(hSink.reg))
+	assert.Equal(t, 1, len(hSink.expReg))
 
-	assert.True(t, definitionsCalled["gauge"], "Gauge definitions were not fetched")
-	assert.True(t, definitionsCalled["counter"], "Counter definitions were not fetched")
+	// assert.True(t, definitionsCalled["gauge"], "Gauge definitions were not fetched")
```
DirectXMan12: should these be removed?

vendor/github.com/hawkular/hawkular-client-go/metrics/client.go
```diff
@@ -315,6 +315,36 @@ func (c *Client) Create(md MetricDefinition, o ...Modifier) (bool, error) {
 	return true, nil
 }
 
+// AllDefinitions fetches all metric definitions (for every tenant) from the server. Requires admin/service rights
+func (c *Client) AllDefinitions(o ...Modifier) ([]*MetricDefinition, error) {
```
DirectXMan12: This seems to have updated without a corresponding change to the Godeps file.

burmanm:
Sorry for the delay, hopefully all the comments are now solved.

DirectXMan12:
LGTM.  Please squash the commits so that each represents a distinct feature or fix (so squash the last commit, "Address PR review comments"), and then I'll merge

burmanm:
I merged all three as the second one was a fix also.

PR #1921, opened 2018-01-02
Export correct cluster location in Stackdriver sink with new resource model

kawych:


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -365,12 +366,17 @@ func CreateStackdriverSink(uri *url.URL) (core.DataSink, error) {
...
-	// Detect zone
-	zone, err := gce.Zone()
+	// Detect zone for old resource model
+	heapsterZone, err := gce.Zone()
 	if err != nil {
 		return nil, err
 	}
 
+	clusterLocation := ""
+	if len(opts["cluster_location"]) >= 0 {
```
x13n: > 0. Or >= 1. But definitely not >= 0 :-)

Also, please add a warning here (and above, for cluster name) if the key is missing from opts.

PR #1923, opened 2018-01-03
support multi-OS cluster by adding nodeSelector to yaml files

idanshahar:
This adds support to the example deployments for multi-OS Kubernetes clusters,
by introducing a `nodeSelector` to the relevant YAML files.

### DIFF
--- a/deploy/kube-config/google/heapster.yaml
+++ b/deploy/kube-config/google/heapster.yaml
@@ -37,6 +37,8 @@ spec:
       - name: ssl-certs
         hostPath:
           path: /etc/ssl/certs
+      nodeSelector:
+        beta.kubernetes.io/os: linux
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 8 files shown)


idanshahar:
I signed it!

DirectXMan12:
First of all, please elaborate in the commit and PR description about what a hybrid cluster is, and what it is that you're doing (it appears "hybrid" refers to a multi-OS cluster, but you need to be clear in the description for posterity).

Secondly, if we're going to do this, we should change all the non-GCE configs, not just the Influx one.

DirectXMan12:
Might not be a bad idea.  My pet peeve with PR templates is that people tend not to fill them out properly and then leave a bunch of cruft in the commit log...

andyxning:
This seems good to me. Ping @DirectXMan12 @piosz for option about changing to [google/heapster.yaml](deploy/kube-config/google/heapster.yaml)

DirectXMan12:
Heapster is now deprecated, so no new features or sinks will be added.  Please see https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md#initial-deprecation-kubernetes-111 for more information.

PR #1580, opened 2017-03-20
Disable metric export flag

aleksandra-malinowska:
Add flag to disable metrics export.

PR #1927, opened 2018-01-11
Bump Kubernetes version to 1.9.1 in integration tests

x13n:
Previous version was no longer working with gcloud past version 1.6.7. Also, it doesn't make sense to test against old version of Kubernetes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ ifeq ($(INTERACTIVE), 1)
 	TTY=-t
 endif
 
-SUPPORTED_KUBE_VERSIONS=1.7.0-beta.2
+SUPPORTED_KUBE_VERSIONS=1.9.1
 TEST_NAMESPACE=heapster-e2e-tests
 
 HEAPSTER_LDFLAGS=-w -X k8s.io/heapster/version.HeapsterVersion=$(VERSION) -X k8s.io/heapster/version.GitCommit=$(GIT_COMMIT)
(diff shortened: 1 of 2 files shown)


x13n:
@aleksandra-malinowska @piosz Looks like we cannot migrate Heapster integration tests because metrics are missing from /api/v1/metric-export. This is probably caused by 
https://github.com/kubernetes/heapster/pull/1580 Can you advise on how to fix the tests now?

aleksandra-malinowska:
Is there a specific test for this endpoint, or is it needed in tests for another reason? If the former, we should probably enable this endpoint, if it's still supported at all. I'm not up to date on Heapster development, but IIRC this was being deprecated.

x13n:
This is how tests are fetching metric time series, see here: https://github.com/kubernetes/heapster/blob/master/integration/heapster_api_test.go#L249:6

x13n:
Ok, I ran the tests manually. It turns out the problem lies elsewhere, heapster logs are flooded with this:
E0115 13:51:27.283122       1 reflector.go:190] k8s.io/heapster/metrics/util/util.go:30: Failed to list *v1.Node: nodes is forbidden: User "system:serviceaccount:heapster-e2e-tests:default" cannot list nodes at the cluster scope
E0115 13:51:27.283299       1 reflector.go:190] k8s.io/heapster/metrics/heapster.go:328: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:heapster-e2e-tests:default" cannot list pods at the cluster scope
E0115 13:51:27.283492       1 reflector.go:190] k8s.io/heapster/metrics/processors/namespace_based_enricher.go:89: Failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:heapster-e2e-tests:default" cannot list namespaces at the cluster scope

x13n:
I just realized that there is https://github.com/kubernetes/heapster/pull/1849, so closing this in favor of the other PR.

PR #1934, opened 2018-01-22
Add fluent Sink for Heaptser

JooyoungJeong:
Fluentd is an open source data collector.  It would be useful  if heapster could sink into fluentd.


DirectXMan12:
If you're adding this sink, you'll need a sink owner (this may be you).  A sink owner is responsible for responding to issues, fixing bugs, and updating dependencies.  If these tasks are not done in a timely manner, or if we go to update Heapster and cannot resolve incompatibilities and the sink owner cannot be reached, this sink may be removed.

If you are ok with these responsibilities, please also update the `sink-owners.md` file with the relevant information.

JooyoungJeong:
Thanks @DirectXMan12 
I am honored to have ownership!! . I've updated sink-owners.md.

PR #1940, opened 2018-01-26
Fix nil pointer dereference when a sink return (nil, err)

ryarnyah:
* Fix nil pointer dereference when a metric sink return an error (use uri to get sink info vs .Name() on a nil pointer)
* Harmonize metrics and events factory code


DirectXMan12:
LGTM

PR #1942, opened 2018-01-29
Remove node_name from new Stackdriver resource labels

kawych:


x13n:
Why?

x13n:
Thanks for giving the context, makes sense.
/lgtm

PR #1957, opened 2018-02-13
Avoid nil pointer dereference on error

aduston:
When `err` is non-nil, `sink` is likely to be nil.


ryarnyah:
Duplicate https://github.com/kubernetes/heapster/pull/1940

DirectXMan12:
merged the other PR.  Thanks, though!

PR #1959, opened 2018-02-15
sinks/influxdb: allow retention policy name to be set

acobaugh:
Allow influxdb retention policy name to be configured with a new `rpname` URL option. This has the added benefit of allowing events and metrics to live in the same database but with different retention periods if desired.

### DIFF
--- a/common/influxdb/influxdb.go
+++ b/common/influxdb/influxdb.go
@@ -41,6 +41,7 @@ type InfluxdbConfig struct {
 	WithFields            bool
 	InsecureSsl           bool
 	RetentionPolicy       string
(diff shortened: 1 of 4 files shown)


acobaugh:
I signed it!

DirectXMan12:
cc @acobaugh

DirectXMan12:
Heapster is now deprecated, so no new features or sinks will be added.  Please see https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md#initial-deprecation-kubernetes-111 for more information.  I apologize for any in-progress sinks or features that will not be merged.

Issue #1893, opened 2017-11-29
InfluxDB/Heapster/Grafana show incorrect individual memory usage values at pod and host level.


pcm32:
Hi there!

Using the setup available in deploy/kube-config/influxdb, I can observe that there some incorrect plots on the grafana side for memory usage:

- Pod level: the individual memory usage values change for a defined pod and a defined time frame when the length of the period plotted (last 2 days, last 7 days) changes (for the same temporal points of course). See an example in [this gist](https://gist.github.com/pcm32/612a79bee6c2fecfa6e4e9387bde9ce9).

- Host level: we see memory usage values reported that go beyond the actual VMs real logical memory (95 GiB for VMs that have 12 GiB for instance). These machines have no swap, so no memory could be considered in this way.

This is running on a Kubernetes 1.8.1 cluster, deployed with Kubespray v2.3.0, on top of Ubuntu Xenial cloud images on OpenStack.

Containers for Grafana/Heapster/InfluxDB are:

Grafana: gcr.io/google_containers/heapster-grafana-amd64:v4.4.3
Heapster: gcr.io/google_containers/heapster-amd64:v1.4.0
Influxdb: gcr.io/google_containers/heapster-influxdb-amd64:v1.3.3

Any ideas? Thanks!

pcm32:
I think that the incorrect part lies in the Grafana query for influxdb being used. Currently it is doing a sum over the time interval used in the group by, 

```
SELECT sum("value") 
FROM "memory/usage" 
WHERE "type" = 'node' AND "nodename" =~ /$nodename$/ 
              AND $timeFilter 
GROUP BY time($interval), "nodename" fill(null)
```

which is probably aggregating things differently on larger or smaller time intervals. Using the mean or the median as aggregator function in the select, you don't see the compounded effect over different time lapses (value levels stay the same, regardless of the time window chosen).

pcm32:
I can confirm that the same is currently happening with the CPU usage graphs, if you change the time window, values change for the same time points. I know what to change now, and if you agree with the change (I would change the aggregator to mean in all the grafana queries on the json files, in lines like [this](https://github.com/kubernetes/heapster/blob/a2ec4cb45407edab53948236468fec6b702447d8/grafana/dashboards/cluster.json#L80)), I'm happy to create a PR with it.

pcm32:
[...]

PR #1961, opened 2018-02-19
Uses mean instead of sum in Grafana queries to avoid plotting variations at different timescale

pcm32:
This PR changes Grafana queries for pods and nodes to address problem describe in #1893.

### DIFF
--- a/grafana/dashboards/cluster.json
+++ b/grafana/dashboards/cluster.json
@@ -76,7 +76,7 @@
                   }
                 ],
                 "measurement": "cpu/usage_rate",
-                "query": "SELECT sum(\"value\") FROM \"cpu/usage_rate\" WHERE \"type\" = 'node' AND $timeFilter GROUP BY time($interval) fill(null)",
+                "query": "SELECT max(\"value\") FROM \"cpu/usage_rate\" WHERE \"type\" = 'node' AND $timeFilter GROUP BY time($interval) fill(null)",
                 "refId": "A",
                 "resultFormat": "time_series",
                 "select": [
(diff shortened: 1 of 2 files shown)


DirectXMan12:
cc @acobaugh

acobaugh:
I'm not sure I agree with the usage of max() instead of mean() here. If I want higher resolution, then I'm going to reduce my $interval. Using max() would actually give a skewed view of reality. Most of the grafana dashboards I see in the wild use mean(). I could maybe see an argument for separate set of metrics showing max over the selected interval, but that would make the graphs pretty crowded.

pcm32:
well I started it initially because it was to sum, which had horrible results. I first used mean, but then it made more sense for most of my observations to use max. But, as long as it moves away from sum to either mean or max, I think is fine. I do think though that mean has an undesired softening effect, but I agree that max has also a bias.

pcm32:
Maybe if some admin or project owner can let me know what they decide, I'm happy to change the branch to reflect it.

acobaugh:
My personal vote is for `mean()`, but I'm still new to the project. Pretty much every single dashboard I've seen over the last few years of using grafana has used `mean()`.

pcm32:
I have switched all max to mean in the dashboards. Hopefully this can now be merged.

abelsromero:
Sorry to ask, but what's the reason not to merge this fix? The issue got closed by `rotten` but this is still pending

pcm32:
I don't know.... I guess we need to ask another owner to approve it...

pcm32:
Maybe @loburm could help us being an owner? he approved a recent PR. Sorry for the ping @loburm, I hope this is fine.

pcm32:
Thanks @DirectXMan12! The tide robot apparently requires an lgtm label from what I read above, are you able to set that up?

abelsromero:
Thanks a lot !!

PR #1964, opened 2018-02-23
Fix grafana dashboards. Fixes #1958

### LINKED ISSUE #1958: Grafana dashboards are invalid
When building the Grafana image from source, you'll notice that the Cluster and Pods dashboards aren't loaded. The reason being that `cluster.json` and `pods.json` are missing the top `dashboard` object, which was explicitly removed in commit https://github.com/kubernetes/heapster/commit/e960cd47fb14847f642ef275da01726496e7faff

Oddly enough the latest docker image does contain the right json files.

acobaugh:
Grafana dashboard JSON needs to be wrapped in "dashboard": {} when
POSTing to /api/dashboards/db [1]. The bare dashboard is usable when
importing via the UI as it handles that wrapping for us. The wrapped
dashboard also imports just fine in the UI.

[1] http://docs.grafana.org/http_api/dashboard/)

This reverts commit e960cd47fb14847f642ef275da01726496e7faff.

### DIFF
--- a/grafana/dashboards/cluster.json
+++ b/grafana/dashboards/cluster.json
@@ -1,4 +1,5 @@
 {
+  "dashboard": {
     "id": null,
     "title": "Cluster",
     "originalTitle": "Cluster",
@@ -2459,4 +2460,6 @@
     "schemaVersion": 8,
     "version": 10,
     "links": []
+  },
+  "overwrite": false
 }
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Don't know why this didn't auto-merge.  Manually merging.

PR #1966, opened 2018-02-24
cloud builder: add Google Cloud Container builder to create heapster

javier-b-perez:
Create build request file to build heapster images using Google Cloud Container builder.


javier-b-perez:
going with a different approach.

Issue #1969, opened 2018-03-02
Error CrashLoopBackOff in heaspter container with statsd sink configuration


bangau1:
**Description**

Hi,

I got error when trying to deploy the Heapster with statsd sink configuration. The error is segmentation fault.

**Steps to reproduce:**
- kubectl apply -f heapster-rbac.yaml
```
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: heapster
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:heapster
subjects:
- kind: ServiceAccount
  name: heapster
  namespace: kube-system
```
- kubectl apply -f heapster.yaml
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: heapster
  namespace: kube-system
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: heapster
  namespace: kube-system
spec:
  replicas: 1
  template:
    metadata:
      labels:
        task: monitoring
        k8s-app: heapster
    spec:
      serviceAccountName: heapster
      containers:
      - name: heapster
        image: k8s.gcr.io/heapster-amd64:v1.5.1
        imagePullPolicy: IfNotPresent
        command:
        - /heapster
        - --source=kubernetes:https://kubernetes.default
        - --sink="statsd:udp://dd-agent-service.default:8125"
        volumeMounts:
        - mountPath: /etc/ssl/certs
          name: ssl-certs
          readOnly: true
        env:
          - name: DATADOG_AGENT_HOST_IP
            valueFrom:
              fieldRef:
                fieldPath: status.hostIP
      volumes:
      - name: ssl-certs
        hostPath:
          path: /etc/ssl/certs
---
apiVersion: v1
kind: Service
metadata:
  labels:
    task: monitoring
    # For use as a Cluster add-on (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons)
    # If you are NOT using this as an addon, you should comment out this line.
    kubernetes.io/cluster-service: 'true'
    kubernetes.io/name: Heapster
  name: heapster
  namespace: kube-system
spec:
  ports:
  - port: 80
    targetPort: 8082
  selector:
    k8s-app: heapster

```

Note: `dd-agent-service.default` is a kubernetes service hostname to abstract the dd-agent daemonset pod running in all our nodes.

**Describe the results you received:**
[...]

PR #1971, opened 2018-03-03
fix removed quotes from statsd example

mhempstock:
This pull corrects the doc mistake identified in  #1969.  Which removes the quotes from the statsd example.


andyxning:
/lgtm
/approve

Thanks for correct this @stash1001

PR #1975, opened 2018-03-12
Use float64 instead of float32 for metrics of Float type

kawych:
This PR fixes a bug where Heapster sometimes exported incorrect metrics (e.g. node allocatable memory < 0) due to float32 overflow.

### DIFF
--- a/metrics/core/types.go
+++ b/metrics/core/types.go
@@ -86,7 +86,7 @@ func (self *UnitsType) String() string {
 
 type MetricValue struct {
 	IntValue   int64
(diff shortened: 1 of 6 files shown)


x13n:
Is there a github issue for that bug?

kawych:
@x13n No, this was noticed by someone from Stackdriver when he verified exported metric values. I didn't file an issue since I have this fix already, do you want me to file it?

x13n:
Nope, I just wanted to avoid not linking one that exists.
/lgtm

PR #1978, opened 2018-03-16
Bump Heapster to version 1.5.2

kawych:


andyxning:
@kawych @x13n Should the version used in [heapster-summary-controller.yaml](https://github.com/kubernetes/heapster/blob/master/deploy/kube-config/standalone-test/heapster-summary-controller.yaml#L20) also be updated?

kawych:
Sorry, I missed it - issued #1979. Thanks!

PR #1985, opened 2018-03-22
Sink Inodes and InodesFree metrics to Stackdriver

neoseele:
This PR is an attempt to allow heapster to sink inodes and inodesfree metrics scraped from k8s nodes/pods to Stackdriver.


metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -460,6 +460,20 @@ func (sink *StackdriverSink) LegacyTranslateLabeledMetric(timestamp time.Time, l
 			"device_name": metric.Labels[core.LabelResourceID.Key],
 		}
 		return ts
+	case core.MetricFilesystemInodesFree.MetricDescriptor.Name:
```
loburm: I think the same change has to be made to TranslateLabeledMetrics method, @kawych is it correct?

loburm:
Before proceeding you need to sign CLA: https://github.com/kubernetes/community/blob/master/contributors/guide/README.md#sign-the-cla

metrics/sinks/stackdriver/stackdriver.go
kawych: Yes, if we add a new metric, we should have the same metric for new resource model. Can you provide more context behind adding this metric? We should decide what metric do we need and what monitored resource to use.

Also, this metric needs to be defined in Stackdriver first, doesn't it?

neoseele: CLA is signed, not sure if the bot will pick it up automatically or not. :)

The reason for this FR is discussed in b/73268296. Since cadvisor's summary api provides inodes metrics already, I see no reason why we shouldn't sink them to SD.

kawych: Thanks for the context, I agree we should add the metric. I've asked folks from Stackdriver to take a look, when the Stackdriver change is in place we can merge this PR.

DirectXMan12:
Heapster is now deprecated, so we will no longer be adding new features or sinks.  Please see https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md#initial-deprecation-kubernetes-111 for more information.  I appologize for any in progress sinks or features that will now no longer be merged.

PR #1989, opened 2018-03-25
add cpu load metrics

andyxning:
**What this PR does**
This PR adds `cpu/load` metrics to heapster.

### DIFF
--- a/docs/storage-schema.md
+++ b/docs/storage-schema.md
@@ -12,6 +12,7 @@ Heapster exports the following metrics to its backends.
 | cpu/request | CPU request (the guaranteed amount of resources) in millicores. |
 | cpu/usage | Cumulative CPU usage on all cores. |
 | cpu/usage_rate | CPU usage on all cores in millicores. |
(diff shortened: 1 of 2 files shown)


docs/storage-schema.md
```diff
@@ -12,6 +12,7 @@ Heapster exports the following metrics to its backends.
 | cpu/request | CPU request (the guaranteed amount of resources) in millicores. |
 | cpu/usage | Cumulative CPU usage on all cores. |
 | cpu/usage_rate | CPU usage on all cores in millicores. |
+| cpu/load | CPU load. |
```
loburm: It would be great to add a bit more information about units of this metric, because I had to check cadvisor documentation to understand that it is actually multiplied by 1000.

loburm (commented):
Overall lgtm.

docs/storage-schema.md
andyxning: Will do.

andyxning: @loburm Done. PTAL.

PR #1993, opened 2018-03-27
Update influxdb and grafana dependencies

007:
Update the builds for `influxdb` and `grafana` to use recent versions.

Can't get `grafana` to cross-compile properly with 5.x because of warnings like this against `go-sqlite3` even when compiling statically:
```
/tmp.k8s/go-link-464062632/000002.o: In function `unixDlOpen':
/go/src/github.com/grafana/grafana/vendor/github.com/mattn/go-sqlite3/sqlite3-binding.c:36542: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

I don't have a good way to test, but if that's not a showstopper (as suggested [elsewhere](https://software.intel.com/en-us/comment/1892902#comment-1892902)) then It would be nice to get `grafana` bumped to 5.0.3 as well. It might be as easy as changing `FROM busybox` to `FROM busybox:glibc`.

### DIFF
--- a/deploy/kube-config/influxdb/grafana.yaml
+++ b/deploy/kube-config/influxdb/grafana.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: grafana
-        image: k8s.gcr.io/heapster-grafana-amd64:v4.6.3
+        image: k8s.gcr.io/heapster-grafana-amd64:v5.0.4
         ports:
         - containerPort: 3000
           protocol: TCP
--- a/deploy/kube-config/influxdb/grafana.yaml
+++ b/deploy/kube-config/influxdb/grafana.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: grafana
-        image: k8s.gcr.io/heapster-grafana-amd64:v4.4.3
+        image: k8s.gcr.io/heapster-grafana-amd64:v4.6.3
         ports:
         - containerPort: 3000
           protocol: TCP
--- a/deploy/kube-config/influxdb/influxdb.yaml
+++ b/deploy/kube-config/influxdb/influxdb.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: influxdb
-        image: k8s.gcr.io/heapster-influxdb-amd64:v1.3.3
+        image: k8s.gcr.io/heapster-influxdb-amd64:v1.5.1
         volumeMounts:
         - mountPath: /data
           name: influxdb-storage
--- a/deploy/kube-config/influxdb/influxdb.yaml
+++ b/deploy/kube-config/influxdb/influxdb.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: influxdb
-        image: k8s.gcr.io/heapster-influxdb-amd64:v1.5.1
+        image: k8s.gcr.io/heapster-influxdb-amd64:v1.5.2
         volumeMounts:
         - mountPath: /data
           name: influxdb-storage
--- a/grafana/Makefile
+++ b/grafana/Makefile
@@ -19,8 +19,8 @@
 
 all: build
 
-VERSION?=v4.6.3
-DEB_VERSION?=4.6.3
+VERSION?=v5.0.4
+DEB_VERSION?=5.0.4
 
 PREFIX?=staging-k8s.gcr.io
 ARCH?=amd64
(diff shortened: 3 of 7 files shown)


007:
Supporting docs for moving to `busybox:glibc` and bumping to 5.0.3:
Building on [kube-cross](https://github.com/mikedanese/kubernetes/blob/master/build/build-image/cross/Dockerfile) -> `FROM golang:1.9.3`
[golang:1.9.3](https://github.com/docker-library/golang/blob/2f2f3b620d61f533484f24a568c2ca46e4fda91c/1.9/stretch/Dockerfile) -> `FROM buildpack-deps:stretch-scm`
[buildpack-deps:stretch-scm](https://github.com/docker-library/buildpack-deps/blob/1845b3f918f69b4c97912b0d4d68a5658458e84f/stretch/scm/Dockerfile) -> `FROM buildpack-deps:stretch-curl`
[buildpack-deps:stretch-curl](https://github.com/docker-library/buildpack-deps/blob/d662dc69f910feb241f6d0c9d2cd6cc2fb6c5e6c/stretch/curl/Dockerfile) -> `FROM debian:stretch`

[libc](https://packages.debian.org/source/stretch/glibc) on `stretch` is `glibc`.

[busybox](https://hub.docker.com/_/busybox/) shows its default is `uclibc`, but does offer `glibc` and is even based on [debian:stretch-slim](https://github.com/docker-library/busybox/blob/27527570c1d504482c64f0aa8f7eb230a311b48a/glibc/Dockerfile.builder).

007:
Single datapoint: this launches and runs (with `grafana` 5.0.3) on docker 18.03.0-ce on OSX with either `libc` - the default `busybox` version or `busybox:glibc`. I don't know how to exercise it further than that, or how to convince it to load the linked libraries and try to execute code with them.

Nodraak:
Hello :)

What's the status on this PR? Do you need help?

BTW, Grafana v5.0.4 and influxdb v1.5.2 are out

007:
I can update tomorrow, but I'm stuck on how to test the "static" linking in sqlite. It's easy to repro the warnings and resulting image, just switch to 5.0.4 and `docker build`.

Nodraak:
The "dlopen in static linked app" issue did not get many answers in the [go-sqlite3 repo](https://github.com/mattn/go-sqlite3/issues/457) ...

---

I don't know golang but the issue might be related to https://github.com/golang/go/issues/21421#issuecomment-322099345 and https://github.com/confluentinc/confluent-kafka-go/issues/59#issuecomment-379296764 ...

TL;DR: try adding the flag `-static-all` (https://github.com/007/heapster/blob/3cce3e41b8484791a70e0c60b8c9269d81024ba4/grafana/Makefile#L28). I'll try that later if I find the time

007:
@Nodraak Changing `-static` to `-static-all` doesn't work, and using the `-tags static_all` also seems iffy since it's throwing the same warnings. https://github.com/confluentinc/confluent-kafka-go/issues/59#issuecomment-379278816 suggests that it's unstable / unsafe, and suggests the same solution as I indicated previously: it's a Docker image, so pack the appropriate dependencies (i.e. `glibc`) with it.

I pushed the changes to use the `busybox:glibc` version, we'll see how unit + e2e works from there.

Per `k8s-ci-robot` suggestion:
/assign @loburm

Nodraak:
So, I've cloned your repo and checkout-ed your branch. I have ran `make build` in both grafana and influx folders, and I've updated by k8s test cluster with the newly created images `staging-k8s.gcr.io/heapster-grafana-amd64:v5.0.4` and `staging-k8s.gcr.io/heapster-influxdb-amd64:v1.5.2`.

Everything is running fine without errors.

I'm not sure what you meant by `I'm stuck on how to test the "static" linking in sqlite`, it looks good on my side!

007:
Thanks for another good data point!

I say "stuck" in that I don't know EXACTLY where the sqlite code is used, or how I can prove the negative - I'd like to see a +/- where `libc` explodes with the wrong one and works with the right one, but I couldn't get that to happen with a different `busybox` `libc` in my limited use.

007:
@andyxning @loburm any suggestions on how to test this to satisfy `lgtm` and get it merged?

Unit and e2e both seem to be working, but it's unclear if they use the built images for the dependencies or if they're pulling whatever base they already have.

Nodraak:
ping @andyxning

andyxning:
/lgtm

@007 @Nodraak Sorry for the late.

PR #1995, opened 2018-03-28
Exports metrics to track non primary container resource requests

jiayingz:
**What this PR does**
- Exports metrics on container non-primary resource requests.
- Adds "kubernetes.io/container/nvidia_gpu/request" metric generated from such metrics to stackdriver sink.

### DIFF
--- a/metrics/core/metrics.go
+++ b/metrics/core/metrics.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	cadvisor "github.com/google/cadvisor/info/v1"
+	kube_api "k8s.io/api/core/v1"
 )
 
 const (
@@ -147,6 +148,15 @@ var NetworkMetrics = []Metric{
 	MetricNetworkTxErrors,
 	MetricNetworkTxErrorsRate,
 	MetricNetworkTxRate,
+}
+
+// Maps from resource name to the metric that tracks container resource request
+// on that resource. The name of the metric is ResourceName/request where ResourceName
+// is the name of the resource requested in container resource requests.
+var ResourceRequestMetrics = map[kube_api.ResourceName]Metric{
+	kube_api.ResourceCPU:              MetricCpuRequest,
+	kube_api.ResourceMemory:           MetricMemoryRequest,
+	kube_api.ResourceEphemeralStorage: MetricEphemeralStorageRequest,
 }
 
 type MetricFamily string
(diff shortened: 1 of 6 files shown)


metrics/processors/pod_based_enricher.go
```diff
@@ -182,6 +183,31 @@ func updateContainerResourcesAndLimits(metricSet *core.MetricSet, container kube
 		metricSet.MetricValues[core.MetricMemoryRequest.Name] = intValue(0)
 	}
 
+	for key, val := range container.Resources.Requests {
+		if key == kube_api.ResourceCPU || key == kube_api.ResourceMemory {
```
jiayingz: @jingxu97 once you add the metric for local storage, I think you want to add it here as well.

metrics/core/metrics.go
```diff
@@ -141,6 +141,14 @@ var NetworkMetrics = []Metric{
 	MetricNetworkTxRate,
 }
 
+// Maps from non primary resource name (i.e., not cpu or memory) to the metric
+// that tracks container resource request on that resource. The name of the metric
+// is in format of formattedResourceName/request where formattedExtendedResourceName
+// is generated from the resource name with "/" replaced by "_".
+// E.g., a container requesting nvidia.com/gpu resource will generates a
+// nvidia.com_gpu/request metric in this map.
+var NonPrimaryResourceRequestMetrics = map[string]Metric{}
```
kawych: Can you put all resource request metrics here, both primary and extended? You can initialize a map with existing primary resource metric. I don't think the logic should be different for extended resources.

What about limits?

metrics/processors/pod_based_enricher.go
kawych: Note that we have kube_api.ResourceEphemeralStorage.

jiayingz (commented):
Thanks a lot for the comments!

metrics/sinks/stackdriver/stackdriver.go
```diff
@@ -613,6 +613,9 @@ func (sink *StackdriverSink) TranslateMetric(timestamp time.Time, labels map[str
 				"memory_type": "evictable",
 			}
 			return ts
+		case "nvidia.com/gpu/request":
```
kawych: Keep in mind that this exposes the metrics for customers using new resource model. New resource model is going to be opt-in in GKE for Kubernetes 1.10 and legacy resource modes is still going to be a default. If you want to expose this metric also to customers with new resource model, please add a second (legacy metric). The appropriate name would be "container.googleapis.com/container/nvidia_gpu/request" and it should be reported against legacy "gke_container" resource type.

kawych:
LGTM once the metric is added to Stackdriver. Please take a look at presubmits and consider adding a corresponding metrics for legacy resource model. Thanks.

jiayingz:
@kawych I think this PR is ready for review now. Thanks!

jiayingz:
/cc @jingxu97 @andyxning 
ping @kawych

PR #1998, opened 2018-03-29
make use of ObjectMeta ListOptions in meta/v1

hzxuzhonghu:
**What this PR does**
As ListOptions and ObjectMeta was moved to meta/v1, and they are deprecated in core api group a year ago, and will be removed in future release. So here change to use `meta/v1.ListOptions ObjectMeta `


andyxning:
@hzxuzhonghu 

There is a pending to merge PR which bumping kubernetes dependency to 1.9.3 in #1849 . It has done the same job in that PR. :)

I suggest we close this in favor of #1849 .

hzxuzhonghu:
@andyxning Thanks.

PR #1999, opened 2018-03-29
Add local ephemeral storage metric into heapster

jingxu97:
This commit adds pod-level cpu, memory and ephemeralstorage metrics from
   summary API to heaspter.
    
Before, some pod-level metrics are aggregated from containers. After getting those metrics directly from pod summary API, it no longer needs to aggregate from containers.

### DIFF
--- a/metrics/core/metrics.go
+++ b/metrics/core/metrics.go
@@ -47,7 +47,9 @@ var AdditionalMetrics = []Metric{
 	MetricCpuRequest,
 	MetricCpuLimit,
 	MetricMemoryRequest,
-	MetricMemoryLimit}
+	MetricMemoryLimit,
+	MetricEphemeralStorageRequest,
+	MetricEphemeralStorageLimit}
 
 // Computed based on corresponding StandardMetrics.
 var RateMetrics = []Metric{
@@ -90,12 +92,16 @@ var LabeledMetrics = []Metric{
 var NodeAutoscalingMetrics = []Metric{
 	MetricNodeCpuCapacity,
 	MetricNodeMemoryCapacity,
+	MetricNodeEphemeralStorageCapacity,
 	MetricNodeCpuAllocatable,
 	MetricNodeMemoryAllocatable,
+	MetricNodeEphemeralStorageAllocatable,
 	MetricNodeCpuUtilization,
 	MetricNodeMemoryUtilization,
+	MetricNodeEphemeralStorageUtilization,
 	MetricNodeCpuReservation,
 	MetricNodeMemoryReservation,
+	MetricNodeEphemeralStorageReservation,
 }
 
 var CpuMetrics = []Metric{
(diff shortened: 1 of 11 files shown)


jingxu97:
@piosz @kawych Could you please help review this PR? Thanks!

kawych:
Lgtm for the core changes. For stackdriver sink, we need to define Stackdriver metrics first (I'm happy to review, please include there also a reviewer from Stackdriver).

jingxu97:
@kawych Thanks for reviewing the PR. I could not add a reviewer. Who you recommend to review the code for stackdriver change? Thanks!

jingxu97:
cc @dashpole

kawych:
@jingxu97 
"metrics/sinks/riemann/driver_test.go:200: k8s.io/heapster/metrics/core.LabeledMetric composite literal uses unkeyed fields"
Not sure why is it failing now, but adding explicit key names ("Name", "Labels", "MetricValue") to the struct in a file mentioned above should fix it.

andyxning:
@jingxu97 The heapster-unit error has been addressed in master,  rebase should address the pull-heapster-unit error in this PR.

metrics/core/metrics.go
```diff
@@ -244,7 +250,7 @@ var MetricCpuUsage = Metric{
 
 var MetricEphemeralStorageUsage = Metric{
 	MetricDescriptor: MetricDescriptor{
-		Name:        "ephemeralstorage/usage",
+		Name:        "ephemeral_storage/usage",
```
andyxning: New metrics should be added to https://github.com/kubernetes/heapster/blob/master/docs/storage-schema.md#metrics

metrics/core/metrics.go
```diff
@@ -502,6 +508,26 @@ var MetricMemoryLimit = Metric{
...
+		Type:        MetricGauge,
+		ValueType:   ValueInt64,
+		Units:       UnitsBytes,
+	},
+}
+
+var MetricEphemeralStorageLimit = Metric{
+	MetricDescriptor: MetricDescriptor{
+		Name:        "ephemeral_storage/limit",
+		Description: "ephemeral sstorage hard limit in bytes.",
```
andyxning: `s/sstorage/storage/`

metrics/core/metrics.go
```diff
@@ -593,6 +619,16 @@ var MetricNodeMemoryCapacity = Metric{
 	},
 }
 
+var MetricNodeEphemeralStorageCapacity = Metric{
+	MetricDescriptor: MetricDescriptor{
+		Name:        "ephemeral_storage/node_capacity",
+		Description: "Ephemeral storage capacity of a node",
+		Type:        MetricGauge,
+		ValueType:   ValueFloat,
```
andyxning: Shouldn't the `ValueType` and `Units` be the same as `ephemeral_storage/limit`.

andyxning: Ignore this since other capacity metrics are also defined like this. :)

metrics/core/metrics.go
jingxu97: fixed. Thanks!

metrics/core/metrics.go
jingxu97: added. Thanks!

PR #2002, opened 2018-04-03
Fix negative value on "disk/io_{read,write}_bytes_rate"

jcharlytown:
**What this PR does**
Fixes #1935 

Makes rate calculator honor resource ids for disk io metrics. This is basically a copy of https://github.com/kubernetes/heapster/pull/1947. This original PR seems not to make any progress anymore.

PR #2008, opened 2018-04-07
Fix negative value on "disk/io_{read,write}_bytes_rate"

andyxning:
closes #2002

**What this PR does**
Makes rate calculator honor resource ids for disk io metrics. This is basically a copy of #2002.


andyxning:
/cc @kawych 

Ping @jcharlytown @qhyou11

PR #2021, opened 2018-04-23
Fixed cat syntax.

bruksnys:
Cat with single less-than character does not work.


kawych:
/ok-to-test

Please sign the CLA as described above.

kawych:
/lgtm

Please sign the cla

PR #2028, opened 2018-04-27
[release-1.5] Fix stackdriver metrics for node memory using wrong metric type

serathius:
Cherry pick of https://github.com/kubernetes/heapster/pull/2027
/cc @kawych


serathius:
@kawych I added separate commit to fix static analysis failure

PR #2031, opened 2018-04-30
Deploy with the source kubernetes.summary_api

### LINKED ISSUE #1648: Heapster cannot dial API server
install k8s use kudeadm with default settings, and installed calico pod networks.

follow the instruction of `influxdb` section of  Readme

found that heapter pod cannot dial api server, report  `  2017-05-19T02:19:48.713825123Z E0519 02:19:48.713791       1 reflector.go:190] k8s.io/heapster/metrics/processors/namespace_based_enricher.go:84: Failed to list *v1.Namespace: Get https://kubernetes.default/api/v1/namespaces?resourceVersion=0: dial tcp 10.96.0.1:443: i/o timeout `

heapster run as service account ` heapster`, there seemed that the serviceaccount have not permission to access api, and after applied   `heapster-rbac.yaml`, still not worked.

I guest  `heapster-rbac.yaml` just defined clusterRoleBinding, but there have not defined the  clusterRole.

zioproto:
In Issue #1648 multiple people are reporting the Kubernetes source to fail with
the error message

    Failed to load nodes: Get https://kubernetes.default:443/api/v1/nodes: dial tcp: i/o timeout

The newer kubernetes.summary_api seems to work

This PR uses by default in the deployment the newer source kubernetes.summary_api  

closes #1648

### DIFF
--- a/deploy/kube-config/influxdb/heapster.yaml
+++ b/deploy/kube-config/influxdb/heapster.yaml
@@ -24,7 +24,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command:
         - /heapster
-        - --source=kubernetes:https://kubernetes.default
+        - --source=kubernetes.summary_api:https://kubernetes.default
         - --sink=influxdb:http://monitoring-influxdb.kube-system.svc:8086
 ---
 apiVersion: v1


zioproto:
It looks like that upgrading to `v1.5.2` here https://github.com/kubernetes/heapster/blob/master/deploy/kube-config/influxdb/heapster.yaml#L23

fixes the problem of issue #1648 

I will cose this PR then.

PR #1891, opened 2017-11-29
Accelerator metrics integration

rohitagarwal003:
Fixes #1872.

Feature repo: kubernetes/features#369.

PR #2046, opened 2018-06-01
Push accelerator metrics to new Stackdriver sink.

rohitagarwal003:
https://github.com/kubernetes/heapster/pull/1891 added pushing to old resource model in Stackdriver. This adds pushing to new resource model.

/cc @piosz @kawych @vishh @jiayingz @x13n

### DIFF
--- a/metrics/sinks/stackdriver/metadata.go
+++ b/metrics/sinks/stackdriver/metadata.go
@@ -81,6 +81,24 @@ var (
 		Name:       "kubernetes.io/container/ephemeral_storage/limit_bytes",
 	}
 
+	acceleratorMemoryTotalMD = &metricMetadata{
+		MetricKind: google_api5.MetricDescriptor_GAUGE,
+		ValueType:  google_api5.MetricDescriptor_INT64,
+		Name:       "kubernetes.io/container/accelerator/memory_total",
+	}
+
+	acceleratorMemoryUsedMD = &metricMetadata{
+		MetricKind: google_api5.MetricDescriptor_GAUGE,
+		ValueType:  google_api5.MetricDescriptor_INT64,
+		Name:       "kubernetes.io/container/accelerator/memory_used",
+	}
+
+	acceleratorDutyCycleMD = &metricMetadata{
+		MetricKind: google_api5.MetricDescriptor_GAUGE,
+		ValueType:  google_api5.MetricDescriptor_INT64,
+		Name:       "kubernetes.io/container/accelerator/duty_cycle",
+	}
+
 	acceleratorRequestedMD = &metricMetadata{
 		MetricKind: google_api5.MetricDescriptor_GAUGE,
 		ValueType:  google_api5.MetricDescriptor_INT64,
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Hold on, why did this merge?.  Heapster is deprecated.  No new features are supposed to go in.  This is a new feature.  Ergo, it shouldn't have gone in.

kawych:
My understanding of this PR was that it's not adding a new feature, but a missing piece of #1891. @mindprince can add more context.

DirectXMan12:
This seems like it's adding something new to the sink.  The idea is that if something that exists doesn't work, we'll fix it, but otherwise, no new code.

mindprince:
Couple of points:
1. In my mind, this was a missing piece of #1891 where these metrics were added in the old resource model but don't work in the new resource model. But I see your point as well.
2. The [heapster deprecation timeline](https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md) says Heapster is deprecated from 1.11 which is not released yet. The heapster deprecation was announced after the 1.11 cycle had already begun and there were already new features (#1989, #1999) in the master branch.

So, I think we need to make one final release of Heapster from the master branch that will go in 1.11. And we shouldn't merge any new PRs to heapster that are not clear-cut bug fixes after 1.11 is released.

DirectXMan12:
> In my mind, this was a missing piece of #1891 where these metrics were added in the old resource model but don't work in the new resource model. But I see your point as well.

I see your point, but that line of logic could be used to justify a lot of new code.  I suggest that you fork Heapster if you want to add anything else along this line.

> The heapster deprecation timeline says Heapster is deprecated from 1.11 which is not released yet

Heapster was deprecated officially as soon as we merged the deprecation PR.  The 1.11 on the timeline was intended to indicate that it was deprecated "as part of the 1.11 development cycle".  Apologies if that wasn't clear.

> So, I think we need to make one final release of Heapster from the master branch that will go in 1.11.

Our existing policy so far has been that no new features will merge into the 1.11 cycle releases of Heapster that weren't already in place by the time the deprecation PR merged.  There were a few features merged before the deprecation PR was in place, so those'll be in the next release.

mindprince:
> Heapster was deprecated officially as soon as we merged the deprecation PR. The 1.11 on the timeline was intended to indicate that it was deprecated "as part of the 1.11 development cycle". Apologies if that wasn't clear.

> Our existing policy so far has been that no new features will merge into the 1.11 cycle releases of Heapster that weren't already in place by the time the deprecation PR merged. There were a few features merged before the deprecation PR was in place, so those'll be in the next release.

Yes, it wasn't clear. Specially because the expectation is that deprecations are aligned with the Kubernetes release schedule. Doing this at an arbitrary point in the cycle doesn't make a lot of sense to me. I also didn't see any announcement in any release notes or in kubenetes-dev@.

In any case, I don't think there are any plans to add more code like this in the 1.11 cycle. So, we should be good.

DirectXMan12:
> Specially because the expectation is that deprecations are aligned with the Kubernetes release schedule

I'll try and clarify in the doc, but Heapster releases haven't traditionally been officially tied to Kubernetes releases.

PR #2069, opened 2018-07-25
Bump Kubernetes version used in integration tests to fix presubmits

kawych:


integration/.jenkins.sh
```diff
@@ -7,7 +7,7 @@ export GOBIN="$GOPATH/bin"
 export PATH="$GOBIN:$PATH"
 
 # Kubernetes version(s) to run the integration tests against.
-kube_version="1.7.0-beta.2"
+kube_version="1.9.3"
```
loburm: why not to use 1.10.5, or even 1.11.0 (that is the last version for which we are going to support heapster)?

kawych: I chose this version because the same version is used on master. Bumped to 1.10.5 - I agree that it's a good idea. Prefer not to update to 1.11 yet since the issue with Stackdriver and instance_id hasn't been fully resolved yet.

loburm: Makes, sense. Thanks.

kawych:
Looks like the OWNERS file is missing on the release 1.5 branch, manually adding "approved" label.

PR #2070, opened 2018-07-26
Add an option to read Host ID from node annotation if ExternalID is not available

kawych:
**What this PR does**

Cherry-picks PR #2057 to release 1.5


kawych:
cc @DirectXMan12, as you were involved in discussion about this change.

PR #2085, opened 2018-08-28
fix retention policy in influxdb sink

### LINKED ISSUE #2084: Retention policy not work in influxdb sink
**Description**
set retention policy in influxdb sink `--sink=influxdb:http://host:8086?retention=4h` and the retention duration will not work.

```
name    duration shardGroupDuration replicaN default
----    -------- ------------------ -------- -------
default 0s       168h0m0s           1       true 
``` 

**Steps to reproduce:**
Start heapster using the sink options: `--sink=influxdb:http://host:8086?retention=4h` with a new influxdb.

**Describe the results you received:**
```
name    duration shardGroupDuration replicaN default
----    -------- ------------------ -------- -------
default 0s       168h0m0s           1       true 
``` 

**Describe the results you expected:**
```
name    duration shardGroupDuration replicaN default
----    -------- ------------------ -------- -------
default 4h0m0s   1h0m0s             1        true
``` 

**Output of `heapster --version`:**
v1.5.1~

ringtail:
Here is the influxdb retention policy creation in master branch. 
```
q := influxdb.Query{
	Command: fmt.Sprintf(`CREATE DATABASE %s WITH NAME "default"`, sink.c.DbName),
}

if resp, err := sink.client.Query(q); err != nil {
  if !(resp != nil && resp.Err != nil && strings.Contains(resp.Err.Error(), "already exists")) {
		err := sink.createRetentionPolicy()
		if err != nil {
			return err
		}
	}
}
```
If the database is not exist,heapster will create a retention policy named `default` without duration. The retention policy with duration will only be called when the query `CREATE DATABASE %s WITH NAME "default"` failed.  

but this query will never fail,even the database has been existed. It will return the msg below.
```
HTTP/1.1 200 OK
Content-Type: application/json
Request-Id: 36d40f46-aabf-11e8-802d-000000000000
X-Influxdb-Build: OSS
X-Influxdb-Version: unknown
X-Request-Id: 36d40f46-aabf-11e8-802d-000000000000
Date: Tue, 28 Aug 2018 12:38:07 GMT
Transfer-Encoding: chunked

{"results":[{"statement_id":0}]}
``` 
So I change the condition of checking database existing with query `SHOW DATABASES`,And create retention policy with duration when the database is not exist before and created by heapster.

### DIFF
--- a/metrics/sinks/influxdb/influxdb.go
+++ b/metrics/sinks/influxdb/influxdb.go
@@ -29,7 +29,7 @@ import (
 )
 
 type influxdbSink struct {
-	client influxdb_common.InfluxdbClient
+	client   influxdb_common.InfluxdbClient
 	sync.RWMutex
 	c        influxdb_common.InfluxdbConfig
 	dbExists bool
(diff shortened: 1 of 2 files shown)


ringtail:
@andyxning Could you help to show reasons of e2e test failed.

ringtail:
I'll check the unit tests and recreate a pr.

PR #2022, opened 2018-04-24
Mark Heapster as Deprecated

DirectXMan12:
This marks Heapster as deprecated, adding a deprecation timeline that
culminates in the retirement of the Heapster project for the Kubernetes
1.13 release.

As per the discussion in SIG Instrumentation.

PR #2089, opened 2018-09-05
eventer support dingtalk sink

ringtail:
DingTalk is a free and all-in-one enterprise communication and collaboration platform. A lot of developers using Alicloud Container Service Kubernetes has their own DingTalk group so that our engineer can solve the problem as soom as possible. And this sink can export kubernetes critical events to DingTalk group through a DingBot.so that our developers can do chatops with eventer and DingTalk. ChatOps maybe the best way to emit event because of it's timeliness.

The usage is very simple：
```
	dingtalk sink usage
	--sink:dingtalk:https://oapi.dingtalk.com/robot/send?access_token=[access_token]&level=Warning&label=[label]

	level: Normal or Warning. The event level greater than global level will emit.
	label: some thing unique when you want to distinguish different k8s clusters.
```

The show case is below:
image


ringtail:
some of user developers has use this eventer sink,and I am also developing the slack sink to support some overseas  developers.

x13n:
Heapster is deprecated and doesn't accept new features. For reference, see https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md

ringtail:
@x13n Is eventer deprecated as well as Heapster？

ringtail:
@x13n  Metric Server can replace Heapster. But there is not any optional choice to replace eventer.

x13n:
Whoops, you're right, I don't think we deprecated eventer as well.
/reopen
/ok-to-test

ringtail:
@x13n Maybe we can move eventer to metrics server as well or simply create a new repo including some other monitoring tools in kubernetes.

kawych:
I believe the deprecation plan still applies to eventer - as documentations says, it is still a part of heapster: https://github.com/kubernetes/heapster/blob/master/docs/overview.md it's also distributed as one image.

I don't know what's the replacement. The purpose was to shift maintenance of integrations from Kubernetes community to logging/monitoring system maintainers, i.e. we rely on third-party tooling to export metrics and events to each logging/monitoring system. Note that Metrics Server has much narrower scope than Heapster and does not cover Eventer functionality.

ringtail:
@kawych That sounds bad.I think eventer will live for a long time although the heapster is deprecated.Our developers use promtheus,heapster or something else as their monitoring tool. But they all use eventer as an alarm tool.so in some ways,eventer is much more useful.We have developed DingTalkSink and AlicloudSink to support the scenarios.

kawych:
Please see our discussion from the PR that deprecated heapster: #2022. The solution suggested by @brancz and @DirectXMan12 in similar case was to create a special-purpose Heapster fork. Maybe you can do similar thing with Eventer? Also, there is a short comment about potential alternatives: #2022#issuecomment-413022143

DirectXMan12:
There's also something like Heptio's [eventrouter](https://github.com/heptiolabs/eventrouter).  Don't know too much about it, but it seems like it might do what you want.

DirectXMan12:
Closing this PR

PR #2091, opened 2018-09-12
Allow setting of export interval for Librato measurements.

ohookins:
Similar to kubernetes/heapster#1755, we want to be able to set the interval between exports from Heapster to Librato.

There is a slight mismatch in the interface here since the last export time is based on when the request was made, not on when the measurements were submitted, but from a practical standpoint this is the most pragmatic.

I'm not so familiar with this codebase so it's possible there's more work to be done. Let me know :)

### DIFF
--- a/common/librato/librato.go
+++ b/common/librato/librato.go
@@ -25,6 +25,8 @@ import (
 	"time"
 )
 
+const defaultMinExportInterval = 60 * time.Second
+
 type Measurement struct {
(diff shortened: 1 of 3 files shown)


DirectXMan12:
Unfortunately, Heapster is now deprecated, and is thus closed to new features.  See [docs/deprecation.md](https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md) for more information.

PR #2105, opened 2018-11-12
Set influxdb version number

jesseshieh:
**What this PR does**
Sets the version number.

See https://github.com/influxdata/influxdb/issues/6777

Copied from https://github.com/influxdata/influxdb/blob/master/build.py#L553


jesseshieh:
I signed it!

PR #2110, opened 2018-11-26
Mark Heapster as retired

DirectXMan12:
We are approaching the 1.13 release, and as per the [deprecation timeline](https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md), that means that Heapster will be officially retired, and moved to the kubernetes-retired organization.

If you're a maintainer and you have some last-minute concerns, please file them here.

Otherwise, on behalf of the Heapster maintainers team, I'd like to thank all the contributors for their contributions :-).

This PR updates the README with a list of potential alternatives to Heapster, and marks Heapster as retired in the README.


DirectXMan12:
cc @piosz @brancz @kubernetes/sig-instrumentation-misc

nikopen:
@DirectXMan12 what are the implications for k/dashboard? It's widely used by end users.

We're very near the release of 1.13 (marked for next Monday), and there's an outstanding change to make the dashboard work without heapster: https://github.com/kubernetes/dashboard/issues/2986

Is there an owner for the above integration? Shouldn't it be first available before retiring Heapster?

pinging release leads
cc @spiffxp @AishSundar

AishSundar:
@DirectXMan12 what does "retiring" mean? If I read the deprecation.md correctly, we only stop any maintenance / bug fixes to Heapster but it will continue to work for end users until they migration to other solutions called out in the README. is that right?

DirectXMan12:
@AishSundar "retiring" means moving to kubernetes-retired, no longer fixing bugs, adding features, or building new releases (Heapster doesn't quite follow the release cadence of Kubernetes).  Existing images will continue to work until they don't, but we provide no guarantees when that is.

As for migration: all new installations of Kubernetes using the recommended setup options in the Kubernetes repo are using metrics-server instead.  We've been mentioning this in community meetings and Kubernetes release notes, and have had the deprecation notice up for several releases now, so hopefully, this hasn't caught anyone off guard.

As for the dashboard, it's unclear what the story is beyond the immediate implications.  The dashboard could continue to consume Heapster until it no longer works, but once Heapster is retired, there's no guarantees about compatibility.  I know the dashboard team has been aware of this change for a while now, but I haven't really heard anything on their path forward.

AishSundar:
@DirectXMan12 Thanks for clarifying the expectations here (FWIW its inline with what thought wll be the state of the world post 1.13) I will followup on kubernetes/dashboard#2986 regarding dashboard support for metricsAPI

kawych:
FYI @bmoyles0117

DirectXMan12:
@loburm @piosz @brancz  can one of you LGTM?  Also, I think the e2e failure is from running different versions of the Go between tests when running gofmt.  I'm disabling that test for now, since this should be literally the last PR we'll make to this repository.

